PR #1, opened 2020-01-16
Timed experiments

blaxill:
Some experimental proofs using lists as signals over time

### DIFF
--- a/cava/timed_experiments/Timed.v
+++ b/cava/timed_experiments/Timed.v
@@ -1,0 +1,158 @@
+(****************************************************************************)
+(* Copyright 2020 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
+(* you may not use this file except in compliance with the License.         *)
+(* You may obtain a copy of the License at                                  *)
+(*                                                                          *)
+(*     http://www.apache.org/licenses/LICENSE-2.0                           *)
+(*                                                                          *)
+(* Unless required by applicable law or agreed to in writing, software      *)
+(* distributed under the License is distributed on an "AS IS" BASIS,        *)
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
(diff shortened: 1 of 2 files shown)


cava/timed_experiments/Timed.v
```diff
@@ -0,0 +1,158 @@
...
+From Coq Require Import ZArith.
+From Coq Require Import btauto.Btauto.
+Require Import Program.Basics.
+Import ListNotations.
+Local Open Scope program_scope.
+Local Open Scope list_scope.
+
+Definition Trace (t: Type) := list t.
+
+Fixpoint zipWith {A : Type} {B : Type} {C: Type} f (l : list A) (l' : list B) : list C :=
```
satnam6502: Note that the Coq Streams library has a zipWith.

cava/timed_experiments/Timed.v
```diff
@@ -0,0 +1,158 @@
...
+    | Delay e => false :: (runExpr e)
+    end.
+
+Definition nand a b := Not (And a b).
+Definition xor a b := 
+    nand (nand a (nand a b)) (nand b (nand a b)).
+
+(* for combinational expressions, set inputs to singleton lists, 
+and take the head of the result trace*)
+Definition combinational2 e a b: _ :=
```
satnam6502: Or consider using hd with a default value with false or zero?

satnam6502:
Very interesting! Do you think you could write a README.md which explains in more detail which is happening in each definition?

cava/timed_experiments/Timed.v
blaxill: I used simple lists in this example but in the future I think either length indexed lists or streams is the way to go. Also the `Trace` definition didn't buy me anything here, but I'm planning on experimenting with more exotic propositions/lists for time samples, which Kami seems to do, that will hopefully allow abstraction/flexibility in model equivalence proofs.

cava/timed_experiments/Timed.v
blaxill: Using length indexed lists or streams would also remove the need for handling this impossible case, I think I'll have a go at converting this file into indexed lists as it shouldnt be too hard

satnam6502:
Also, you might want to consider https://coq.inria.fr/library/Coq.Vectors.VectorDef.html in future.

PR #2, opened 2020-01-23
Mostly show the half/full adders do arithmetic, plus a bit of cleanup

benlaurie:


cava-examples/FullAdder.v
```diff
@@ -50,17 +50,14 @@ Definition halfAdderTop {m t} `{CavaTop m t} :=
...
 Lemma halfAdder_behaviour : forall (a : bool) (b : bool),
                             combinational (halfAdder a b) = (xorb a b, a && b).
 
 Proof.
   intros.
-  unfold combinational.
-  unfold fst.
-  simpl.
   case a, b.
-  all : reflexivity.
```
satnam6502: How about just "intuition" after cases? Does that work?

cava-examples/FullAdder.v
```diff
@@ -60,7 +60,29 @@ Proof.
...
-   
+(* From the original nand *)
+
+Definition p_bit_to_nat(a : bool) : nat := if a then 1 else 0.
+
+Fixpoint p_bits_to_nat(b : list bool) : nat :=
+  match b with
+    | nil => 0
+    | b0::t => (p_bit_to_nat b0) + 2 * p_bits_to_nat t
+  end.
```
satnam6502: Here you could try to express in the type the result is less than 4?
Or at least add it as a comment?

cava-examples/FullAdder.v
```diff
@@ -60,7 +60,29 @@ Proof.
...
 
-   
+(* From the original nand *)
+
+Definition p_bit_to_nat(a : bool) : nat := if a then 1 else 0.
+
+Fixpoint p_bits_to_nat(b : list bool) : nat :=
+  match b with
+    | nil => 0
+    | b0::t => (p_bit_to_nat b0) + 2 * p_bits_to_nat t
```
satnam6502: Do you need the brackets?

satnam6502: Pattern match using names like `| b0:bs` rather than using `t`?

cava-examples/FullAdder.v
```diff
@@ -60,7 +60,29 @@ Proof.
...
+  end.
+
+(* end from *)
+
+(* A proof that the half adder does arithmetic *)
+Lemma halfAdder_nat :
+  forall (a b : bool),
+    let (l, h) := combinational (halfAdder a b) in p_bits_to_nat (l::h::nil) = p_bit_to_nat a + p_bit_to_nat b.
```
satnam6502: With the right imports you can write `[l; h]` instead of `(l::h::nil)`. Otherwise nice!

cava-examples/FullAdder.v
```diff
@@ -97,6 +119,15 @@ Proof.
   all : reflexivity.
 Qed.
 
+(* Prove the full adder does arithmetic *)
+Lemma fullAdder_nat :
+  forall (a b c : bool),
+    let (l, h) := combinational (fullAdder a b c) in p_bits_to_nat (l::h::nil) = p_bit_to_nat a + p_bit_to_nat b + p_bit_to_nat c.
```
satnam6502: Likewise here you can also write `[l; h]`.

cava-examples/FullAdder.v
```diff
@@ -133,3 +164,13 @@ Proof.
...
+
+(* Prove the full adder does arithmetic *)
+Lemma fullAdderFC_nat :
+  forall (a b c : bool),
+    let (l, h) := combinational (fullAdderFC a b c) in p_bits_to_nat (l::h::nil) = p_bit_to_nat a + p_bit_to_nat b + p_bit_to_nat c.
+Proof.
+  intros.
+  case a, b, c.
+  all: reflexivity.
+Qed.
```
satnam6502: Nice!

cava-examples/FullAdder.v
benlaurie: It does, but so does just "reflexivity", not sure why we'd prefer one over the other?

cava-examples/FullAdder.v
benlaurie: OK, but what does the "s" stand for? "t" is "tail", btw...

cava-examples/FullAdder.v
benlaurie: The result is not less than 4.

cava-examples/FullAdder.v
benlaurie: Done.

cava-examples/FullAdder.v
benlaurie: Done.

cava-examples/FullAdder.v
satnam6502: Ah, right, this is a general bit to nat. See `bits.v` for various way to do this. This is fine for now.

cava-examples/FullAdder.v
satnam6502: Adding this lemma is going to cause confusion with similar lemmas I am working on but which have not been added yet. I don't think it is a good idea for both of us to bang on the `cava-examples` directory.

cava-examples/FullAdder.v
```diff
@@ -141,6 +172,17 @@ Proof.
   all : reflexivity.
 Qed.
 
+
+(* Prove the full adder does arithmetic *)
+Lemma fullAdderFC_nat :
+  forall (a b c : bool),
+    let (l, h) := combinational (fullAdderFC a b c) in p_bits_to_nat (l::h::nil) = p_bit_to_nat a + p_bit_to_nat b + p_bit_to_nat c.
```
satnam6502: Again, this is going to clash with what I am working on. I will approve for now, but likely delete it later. 
Again, you can change `(l::h::nil)` to `[l; h]`.

benlaurie: Fair enough. I thought I had make that change, probably lost in the rebasing confusion :-)

satnam6502:
Also note that I am trying to keep the nat level proofs in a separate libraries e.g. `FullAddernat.v` and I am trying to move over bit-vector operations to `BitVector.v`.

satnam6502:
Also, I think your lemmas that try to prove these circuits do "arithmetic" should be written in a way that has only arithmetic (nat) expressions on one side, say `a + b + cin` e.g.
```coq
Lemma fullAdderNat :
  forall (a : nat) (b : nat) (cin : nat), a < 2 -> b < 2 -> cin < 2 ->
  let (sum, carry_out) := combinational (fullAdder (nat2bool a) (nat2bool b) (nat2bool cin)) in
  bits_to_nat [sum; carry_out] = a + b + cin.
```
I think this is an easy change for you to make, and this would be a more useful lemma about proving the behaviour of a full adder.

satnam6502:
Do you want to resolve the conflicts or do you have new code that supersedes this work?

satnam6502:
Let's close this for now. Feel free to reopen later if you would like this to land.

PR #3, opened 2020-01-24
Ignore more things.

benlaurie:


.gitignore
```diff
@@ -4,6 +4,7 @@
 *.glob
 *.pb
 *.log
+*.sv
```
satnam6502: This would exclude human written SystemVerilog files. Instead, in each directory, name the SystemVerilog files that need to be ignored.

satnam6502:
I think we should close the PR because I don't think we should have a directory wide rule to suppress all SystemVerilog files because some of them are written by humans and checked in.

PR #8, opened 2020-03-10
Simplify netlist types

blaxill:
- Separates CavaState into CavaState and Module, and adds alias Netlist for list Instance 
- Merges instance into primitive

I've attempted to separate some of the concerns of the netlist types, I've checked the generated systemverilog is the same.

### DIFF
--- a/cava/Cava2SystemVerilog.hs
+++ b/cava/Cava2SystemVerilog.hs
@@ -1,11 +1,11 @@
 {- Copyright 2020 The Project Oak Authors
-  
+
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
    You may obtain a copy of the License at
-  
+
        http://www.apache.org/licenses/LICENSE-2.0
-  
+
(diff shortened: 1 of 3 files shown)


cava/Cava2SystemVerilog.hs
```diff
@@ -106,17 +106,32 @@ instanceArgs inst
       Netlist.Xorcy li ci o -> [o, ci, li]
       Netlist.Muxcy s di ci o -> [o, ci, di, s]
 
-generateInstance :: Netlist.Instance -> String
-generateInstance (Netlist.Coq_mkInstance number (Netlist.DelayBit i o))
+instanceNumber :: Netlist.Primitive -> Integer
+instanceNumber inst
```
satnam6502: So this is the instance number for gates with just one output. Something else will need to be done for gates with more than one bit of output.

blaxill: Yes this Haskell would need to change, although the Primitive definition doesn't preclude multiple outputs.

PR #10, opened 2020-03-11
Add a few comments for the NAND and XOR arrow examples

satnam6502:
I think it would be good to add just a few brief comments about what is going on here.


cava/Cava/ArrowExamples.v
```diff
@@ -12,10 +12,15 @@ Require Import Cava.Arrow.
 Require Import Cava.Netlist.
 
 Section Example1.
+
+
+  (* Implement a NAND circuit by composing an AND gate and INV gate. *)
   Definition nand
     {Cava: Cava}
```
satnam6502: Oh, and for consistency, perhaps this should be `{_ : Cava}`.

PR #13, opened 2020-03-12
Add SystemVerilog compilation for arrow version of looped NAND

satnam6502:
Also fix bug in netlist generation for arrow delays.


satnam6502:
I'll self merge so I can unblock the PR after this.

PR #14, opened 2020-03-12
Add verilator lint checks for automatically generated SystemVerilog files

satnam6502:
Add Verilator lint checks for auto-generated SystemVerilog files.


satnam6502:
Ah, the Dockerfile now needs to provide `verilator`.

PR #16, opened 2020-03-12
Clean .*.aux files in Cava

satnam6502:
Works just for the Cava directory for now. Perhaps there is a way to use in information in the `.gitignore` of the parent directory.

### DIFF
--- a/cava/Makefile
+++ b/cava/Makefile
@@ -70,4 +70,5 @@ clean:
 		rm -rf Makefile.coq .Makefile.coq.conf \
                        *.hi *.o ExamplesSV FixupAscii dist work-obj93.cf
 		xargs rm -rf < .gitignore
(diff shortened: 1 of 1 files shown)


satnam6502:
Alas that was not enough to fix the CI build break :-(
```
Step #1 - "run_cava_make": Linking ExamplesSV ...
Step #1 - "run_cava_make": ./ExamplesSV
Step #1 - "run_cava_make": verilator --lint-only verilator.vlt nand2.sv
Step #1 - "run_cava_make": make: verilator: Command not found
Step #1 - "run_cava_make": make: *** [Makefile:40: nand2.sv] Error 127
```

blaxill:
I've temporarily fixed the CI, but now there is a failure due to deprecation warnings from Verilator.

PR #20, opened 2020-03-16
Simple mechanized proofs

blaxill:
Attempt at making some simple generalised tactics


satnam6502:
I'm not very familiar with `Ltac` programming but at some informal level this looks great to me, thanks!

PR #21, opened 2020-03-23
Cleanup Arrow

blaxill:
This moves around arrow parts into a less adhoc hierarchy


satnam6502:
Looks like a rational re-organization. At some point we can think more carefully about how to invoke Coq's extraction mechanism esp. regarding recursive extraction of modules.

PR #22, opened 2020-03-23
Unsigned 4-bit adder (synthesizable)

satnam6502:
A request for a checkpoint commit that adds just enough functionality to express a 4-bit adder which can be synthesized from SystemVerilog, and a simple test bench (not assertion checking yet).


blaxill:
Looks good to me, perhaps in the future we can wrap the list bool along with its bitlength

satnam6502:
The use of `list bool` to represent a bit-vector is only temporary here. Soon I want us to replace this with something that represents the vector length in the type e.g. `Bitvector` from the Coq library.

PR #29, opened 2020-03-26
Make adder used Vector type

satnam6502:
This reworks `unsignedAdder` to use the Vector type which allows us to explicitly encode bit-vector sizes in the type.


satnam6502:
This version still encodes what vectors are to be added using the indexes calculated on the Coq side, s although the typeclass has:
```
unsignedAdd : forall {a b}, Vector.t bit a -> Vector.t bit b -> m (Vector.t bit (max a b + 1))
```
the netlist primitive is still:
```
UnsignedAdd : Z -> Z -> Z -> Primitive
```
In a future PR I may move the mapping to SystemVerilog vectors to the Haskell side, and this would allow the netlist primitive to adopt a dependent type using the Coq `Vector` type.

PR #37, opened 2020-03-31
Arrow: Finish primitives and simplify

blaxill:


satnam6502:
I'll wait until this is not in draft mode before reviewing. Please let me know if you want a pre-review.

cava/Cava/Arrow/ArrowExamples.v
```diff
@@ -90,7 +90,8 @@ Section SimpleExamples.
 End SimpleExamples.
 
 Section StreamProofs.
-  (* 1. explicit proof - manual method *)
+  (* TODO: revive this proof? didn't survive refactor *)
+  (*(* 1. explicit proof - manual method *)
```
blaxill: Note that proofs 1. 2. 3. all cover the same thing, proof 3. which uses a custom (but very naive) method survived the refactor! :)

cava/Cava/Arrow/Arrow.v
```diff
@@ -87,21 +88,33 @@ Print Arrow.
...
 
-  (* representable : object -> Type;
-  constant {x} : representable x -> (unit ~> x); *)
-
   bit : object;
-  fromBool : bool -> (unit ~> bit);
+  bitvec : nat -> object;
+
+  constant : bool -> (unit ~> bit);
+  (* constant_vec (n:nat) : Bvector n -> (unit ~> bitvec n); *)
```
satnam6502: Ah, is there an issue with `constant_vec` ? Or have you just not around to making an instance for it?
So `constant` is for relating a `bool` value to `bit` and you can't make a `constant` that is more general?

cava/Cava/Arrow/Arrow.v
```diff
@@ -87,21 +88,33 @@ Print Arrow.
...
+  xor_gate:  bit ** bit ~> bit;
+  xnor_gate: bit ** bit ~> bit;
+  buf_gate:  bit        ~> bit;
+
+  xorcy:     bit ** bit ~> bit;
+  muxcy:     bit ** bit ** bit ~> bit;
 
-  not_gate : bit ~> bit;
-  and_gate : (bit ** bit) ~> bit;
+  unsigned_add a b: bitvec a ** bitvec b ~> bitvec (max a b + 1);
```
satnam6502: Nice.

cava/Cava/Arrow/Instances/Coq.v
```diff
@@ -39,11 +42,28 @@ Section CoqEval.
...
+    muxcy '(x,y,z) := if x then y else z;
+
+    unsigned_add m n '(av, bv) :=
+      let a := bitvec_to_nat av in
+      let b := bitvec_to_nat bv in
+      let c := a + b in
+      nat_to_bitvec (max m n + 1) c;
   }.
 
   Eval cbv in not_gate true.
```
satnam6502: How about turning these into "tests"?
```
  Example not_true : not_gate true = false.
  Proof. reflexivity. Qed.

  Example not_false : not_gate false = true.
  Proof. reflexivity. Qed.
```

cava/Cava/Arrow/Arrow.v
blaxill: Done! I had initially held off as I thought it might be a pain for arrow netlist but it turned out to be very straight forward.
 
`constant` can be made more general, although because of the generality of `object` I think it requires a little machinery e.g.:

~~~coq
Inductive Representable :=
| RBool: bool -> Representable
| RVec (n: nat): Bvector n -> Representable.

Class Cava := {
   denoteRepresentable (r: Representable): object :=
   match r with
   | RBool _ => bit
   | RVec n _ => bitvec n
   end;

   constant (rep: Representable): (unit ~> denoteRepresentable rep);

   ...
~~~

I'm holding off on this for now.

cava/Cava/Arrow/Instances/Coq.v
blaxill: Done

Issue #41, opened 2020-04-01
Improve directory structure for Cava


satnam6502:
The current directory structure for Cava has grown organically but now I think we should move to something more organized. Currently Cava files are split across three directories: cava (the core system but this also has some examples that serve as light regression tests), cava-examples (which has examples) and a nandgame directory (a Cava version of the nand game done by Ben Laurie in the nand directory). Here's a proposal for a better directory structure. Let's discuss and iterate until we come to something that we can agree upon. Here's the current state of the proposal (which we can edit as we 
discuss).

Key points:

- Move to having everything in a single cava directory, with examples nested underneath and a top level README.md that has a good overall explanation of the Cava project and which is well linked to examples, documentation etc.

- To build the system should just require installation of pre-requisites and then typing `make clean all` or just `make`. Regression tests. could be run by `make test` (and perhaps automatically run as part of `make all`).

- The CI system should make sure to track enough of the code under Cava.

- Right now it is just a few us doing experimental evaluation but at some point we should have a way to incorporating contributed circuits and proofs by others e.g. into a `contrib` directory.

- We need to think of where to put the code for the OpenTitan components and a process for using our generated circuits being incorporated into the OpenTitan flow.

A first cut at directory structure inside the `cava` directory:

- Another directory called `cava` with contains a subdirectory Cava: the directory that contains the core Cava system and "standard" library circuits and proofs. Similar to a cleaned up version of the current cava directory, but with examples etc. removed. Also, this is where the current arrow code will end up (hopefully replacing the monadic version).
- examples: The current `cava-examples` plus any examples we want to pick from the current `cava` directory. Include a process for generating SystemVerilog for simulation and synthesis.
- tests: Regression etc. tests that include a process for generating SystemVerilog and testbenches which are run and checked using Verilator.  Include a process for generating SystemVerilog for simulation (but synthesis is probably not required for this directory).
- opentitan: A directory that contains the components we are developing for OpenTitan.
- contrib: A directory where we can accept contributions. We should seed this with one or two examples and a clear process for how to add circuits and write tests etc.
- A `Makefile` at the top level which builds everything with rules for `all` and `clean`. Each sub-directory should also have a `Makefile` with `all` and `clean` rules for its components.

So from the top-level cava directory:
```bash
$ ls
cava contrib examples Makefile opentitan tests
```

blaxill:
> The CI system should make sure to track enough of the code under Cava.

[...]

PR #45, opened 2020-04-02
Move cava into a subdirectory and make cava-examples a peer

satnam6502:
The PR is a stepping stone towards a better directory structure as described in #41 
It also removes old versions of Cava `v1` and `v2` and it deletes the `nandgame` example which I don't think is worth maintaining anymore.


blaxill:
Lgtm!

Issue #53, opened 2020-04-06
Prove properties about nat to bit-vector conversion


satnam6502:
Prove:
```coq
Lemma bits_of_nat_sized: forall n bv, nat_to_bitvec_sized n (bitvec_to_nat bv) = bv.
Lemma nat_of_bits: forall v, bitvec_to_nat (nat_to_bitvec v) = v.
Lemma nat_of_bits_sized: forall (v : nat),
      bitvec_to_nat (nat_to_bitvec_sized (N.size_nat (N.of_nat v)) v) = v.
```

satnam6502:
Closed via #54.

PR #54, opened 2020-04-06
Proofs about nat to bitvector conversion

satnam6502:
Addresses #53 
Proofs for:
```coq
Lemma bits_of_nat_sized: forall n bv, nat_to_bitvec_sized n (bitvec_to_nat bv) = bv.
Lemma nat_of_bits: forall v, bitvec_to_nat (nat_to_bitvec v) = v.
Lemma nat_of_bits_sized: forall (v : nat),
      bitvec_to_nat (nat_to_bitvec_sized (N.size_nat (N.of_nat v)) v) = v.
```

Still to think about:
```coq
Lemma nat_of_bits_sized_n: forall n (v : nat),
      bitvec_to_nat (nat_to_bitvec_sized n v) = v.
Admitted.
```

### DIFF
--- a/cava/cava/Cava/Arrow/Instances/Coq.v
+++ b/cava/cava/Cava/Arrow/Instances/Coq.v
@@ -63,7 +63,7 @@ Section CoqEval.
       let a := bitvec_to_nat av in
       let b := bitvec_to_nat bv in
       let c := a + b in
-      nat_to_bitvec (max m n + 1) c;
+      nat_to_bitvec_sized (max m n + 1) c;
   }.
 
   Example not_true: not_gate true = false.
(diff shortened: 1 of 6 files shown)


satnam6502:
Attempting a proof of:
```coq
Lemma nat_of_bits_sized_n: forall n (v : nat),
      bitvec_to_nat (nat_to_bitvec_sized n v) = v.
```
with an appropriate context had made me think we should deal with bit-growth differently. For example:

* Specify to result of addition in terms of `N.size_sat` or `N.size`
* Always require the size of the addition result to be explicitly specified by the instantiation of the adder circuit, and define the semantics of the adder in terms of addition and modulus.

satnam6502:
Well attempting that proof made me think that this is not the right way to deal with bit-growth. The theorem as stated might not be proveable depending on how `nat_to_bitvec_sized n v ` is defined because we have to worry about values of `n` which are not large enough to represent `v`.

PR #55, opened 2020-04-07
Reintroduce shape from v2 and separate primitives from binding sites

blaxill:
This reintroduces 'shape' from Cava v2, allowing named ports of arbitrary shape and for the shape to be reused in higher level representations e.g. arrows. 

It also separates primitives from their bindings, which will allow the primitives to be reused in Netlist DSLs (and other DSLs), that might not want to directly specify `N` bindings. 

Also temporarily remove the existing Netlist DSL, a newer iteration to be reintroduced in follow up PRs.

### DIFF
--- a/cava/cava/Cava2SystemVerilog.hs
+++ b/cava/cava/Cava2SystemVerilog.hs
@@ -13,6 +13,8 @@
    limitations under the License.
 -}
 
+{-# LANGUAGE ViewPatterns #-}
+
 module Cava2SystemVerilog
 where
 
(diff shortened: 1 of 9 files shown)


cava/cava/Cava2SystemVerilog.hs
```diff
@@ -114,14 +110,17 @@ computeVectors instances
...
   = do NetlistGenerationState v vDecs <- get
        let vA = VectorDeclaration AssignTo   v       HighToLow a
            vB = VectorDeclaration AssignTo   (v + 1) HighToLow b
            vC = VectorDeclaration AssignFrom (v + 2) HighToLow s
        put (NetlistGenerationState (v + 3) (vA:vB:vC:vDecs))
        return ("  assign v" ++ show (v + 2) ++ " = v" ++ show v ++
-               " + v" ++ show (v + 1) ++ ";")                                       
+               " + v" ++ show (v + 1) ++ ";")
+  where
+   (a, b) = Netlist.unsafeCoerce any_ab
```
blaxill, line 122: There are a number of manual coercions in this file im not happy with, I think I should write these functions in Coq so that any coercions needed are performed by Coq. Do you have any thoughts on this?

blaxill:
I've rebased and moved the Haskell functions into Coq that were needed to not use the Coq extracted `Any`s that resulted unsafeCoerce.

Issue #56, opened 2020-04-07
Rationalize unsigned adders to explicitly specify the size of the sum result vector


satnam6502:
I think it is better to explicitly specify the size of the sum result vector and define the semantics of unsigned addition in terms of addition and  modulus.

PR #57, opened 2020-04-07
Change unsignedAdder to take explicit sum length

satnam6502:
Also some re-organization to put unsigned adders into a file under `Cava.Monad`.
Addresses #56


blaxill:
Merging so I can rebase my PR on top

PR #59, opened 2020-04-10
SystemVerilog test bench for adder_tree4_8

satnam6502:


satnam6502:
I'll self merge this due to the long holiday weekend in the UK.

PR #63, opened 2020-04-12
Move arrow versions of xor and loopedNAND to correct directory and ac…

satnam6502:
…tivate


satnam6502:
Abandoning in preference to #64 .

PR #66, opened 2020-04-15
Add Arrow notation

blaxill:
This adds arrow notation. It is still missing a key part of the compilation to generalized arrows which I am continuing to work on.

Some of the syntax will simplify further when the compilation pass is done. Items here built as `kappa` types are left as `kappa` types and require separate notation, but they directly reduce to an arrow in future and not require separate notation.

### DIFF
--- a/cava/cava/Cava/Arrow/Syntax.v
+++ b/cava/cava/Cava/Arrow/Syntax.v
@@ -3,112 +3,119 @@ Require Import Coq.Bool.Bool.
 Require Import Coq.Lists.List.
 Require Import Coq.Strings.String.
 
-From Coq Require Import btauto.Btauto.
+Require Import Cava.Arrow.Arrow.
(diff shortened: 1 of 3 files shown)


cava/cava/Cava/Arrow/Syntax.v
```diff
@@ -3,112 +3,122 @@ Require Import Coq.Bool.Bool.
...
+
+Notation "x" := (Var x) (in custom expr, x ident).
+
+
+Notation "( x )" := x (in custom expr, x at level 4).
+
+(* misc items for testing *)
+
+Definition not'
+  {_: Cava}
```
satnam6502, line 57: You could also just write:
```
`{Cava}
```
to be consistent with code elsewhere.

blaxill: Done

PR #67, opened 2020-04-15
Enable Arrows with the updated Netlist

blaxill:
The Netlist Arrow instance is currently restricted to one dimensional bitvectors, and use some intermediate types that will be removed in the future.

### DIFF
--- a/cava/cava/Cava/Arrow/Instances/Netlist.v
+++ b/cava/cava/Cava/Arrow/Instances/Netlist.v
@@ -1,6 +1,6 @@
 From Coq Require Import Program.Tactics.
 From Coq Require Import Bool.Bool.
-From Coq Require Import Bool.Bvector.
+(* From Coq Require Import Bool.Bvector. *)
 From Coq Require Import Vector.
 From Coq Require Import Lists.List.
 From Coq Require Import Strings.String.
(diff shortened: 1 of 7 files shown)


cava/cava/Cava/Arrow/Instances/Netlist.v
```diff
@@ -19,27 +19,51 @@ From Cava Require Import Netlist.
...
+  Definition numberItem (a: wireTy) (i:N): (wireToNetlistTy a * N) :=
+  match a with
+  | Bit1 => (i, (i+1)%N)
+  | BitVec1 n => (map N.of_nat (seq (N.to_nat i) n), (i + N.of_nat n)%N)
+  end.
 
   (* fill an shape with a incremental numbering corresponding to new ports *)
-  Fixpoint numberShape (s: shape) (i:N) : (signalTy N s * N) :=
-  match s in shape return (signalTy N s * N) with
+  Fixpoint numberShape (s: shape) (i:N) : (shapeToNumbered s * N) :=
```
satnam6502, line 52: So another approach might be to define a function `bitsInShape` which take a shape and returns the number of bits in it (i.e. how much `i` to use up). This could avoid returning a tuple of shape values and a new index offset.

PR #68, opened 2020-04-15
Switch from nat to N for values convertible to bit-vectors

satnam6502:
As we process lager numbers it is going to be much more efficient to represent them in a binary way using `N` rather than `nat`. I can't immediately work out how to re-do one of the proofs with `N` instead of `nat` for the rule `Nat.pow_succ_r` so admitting for now.


satnam6502:
I think this is a non-contraverstial change so I shall self-merge it because I have quite a few follow up PRs that depend on it.

PR #69, opened 2020-04-15
Make adder semantics assume bit-growth implicitly

satnam6502:
This simplifies semantic and proofs, and I think we should explicitly separate out the process of `mod`-ing values to shave off bits.


satnam6502:
Again, self-merging this change to unblock other PRs.

PR #70, opened 2020-04-17
Add circuit interface mechansim

satnam6502:
This PR is a first attempt at adding a mechanism for describing circuit interfaces which will allow us to do things like automate the generation of test benches. Several things are not right with the current approach but it is a step forward from the current ad hoc port wiring process.

The `singalTy` function does not seem to work for tuples of tuples, as described in the `Cava.v` files. Currently the adder tree example uses a workaround for this.

This PR no longer associates port net numbers with the port declarations . Instead, the association between port names and internal nets is made using four wiring components. This means the arrow examples don't generate working SystemVerilog now because I was not able to adjust the arrow code to use to new components. If this is too much of a hassle then we can try to add back the port numbers to the port declarations (but right now I feel that information does not belong there: a port specification is independent of how the net numbers are generated for the ports).

The syntax for specifying circuit interface types is far from ideal but after doing a few examples I suspect we can come up with some way of making it more rational.

The use of the `Flatten` type class means that circuits that are to be mapped to a netlist can only have one input and the input type is a combination of bits, bit-vectors or tuples of these types but no other types.

### DIFF
--- a/cava/cava/Cava/Monad/Cava.v
+++ b/cava/cava/Cava/Monad/Cava.v
@@ -70,13 +70,6 @@ Class Cava m bit `{Monad m} := {
 }.
 
 (******************************************************************************)
-(* Execute a monadic circuit description and return the generated netlist.    *)
-(******************************************************************************)
-
-Definition makeNetlist {t} (circuit : state CavaState t) : CavaState
-  := execState circuit initState.
-
-(******************************************************************************)
 (* Netlist implementations for the Cava class.                                *)
 (******************************************************************************)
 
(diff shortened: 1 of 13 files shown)


cava/cava/Cava/Netlist.v
```diff
@@ -22,28 +22,59 @@
...
+(* shape describes the types of wires going into or out of a Cava circuit,    *)
+(* but not all shapes can be bound to a SystemVerilog port name. Those that   *)
+(* can are identified by the One constructor.                                 *)
+(******************************************************************************)
 
 Inductive shape {A: Type} : Type :=
   | Empty : shape
   | One : A -> shape
-  | Tuple2 : shape -> shape -> shape (* A pair of bundles of wires *).
+  | Tuple2 : shape -> shape -> shape. (* A pair of bundles of wires *)
```
blaxill, line 43: I've realised we should probably name this as tree, and perhaps abstract it out from netlist as we are gathering a number of generic functions that apply to trees. I'll take a shot at it in a follow up once i've finished my current task

cava/cava/Cava/Monad/Cava.v
```diff
@@ -228,43 +221,152 @@ Definition setModuleName (name : string) : state CavaState unit :=
...
+     = ([0%N; 1%N; 2%N; 3%N; 4%N; 5%N; 6%N; 7%N], [8%N; 9%N; 10%N; 11%N; 12%N; 13%N; 14%N; 15%N],
+       ([16%N; 17%N; 18%N; 19%N; 20%N; 21%N; 22%N; 23%N], [24%N; 25%N; 26%N; 27%N; 28%N; 29%N; 30%N; 31%N]))
+     : signalTy N (mapShape snd tupleOfTuples)
+
+Which is wrong because it has the structure (a, b, (c, d)) instead of ((a,b), (c, d)).
+TODO(satnam): Fix.
+
+*)
+
+Definition instantiateInputPort (pd : PortDeclaration) : state CavaState unit :=
```
blaxill, line 316: Is it possible to write these as `state "NetlistState" unit` and then lift to `state CavaState unit`? I'm not sure how featureful the coq monad library is. Doesn't need to be addressed here I'm just thinking ahead

cava/cava/Cava/Monad/Cava.v
```diff
@@ -228,43 +221,152 @@ Definition setModuleName (name : string) : state CavaState unit :=
...
+        let insts' := BindPrimitive (WireOutputBitVec name size) v tt :: insts in
+        put (mkCavaState o isSeq (mkModule n insts' inputs (newPort :: outputs))) ;;
         ret v
   end.
 
+(******************************************************************************)
+(* Execute a monadic circuit description and return the generated netlist.    *)
+(******************************************************************************)
+
+Fixpoint bitsInPort (s : @shape portType) : nat :=
```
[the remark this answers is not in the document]
satnam6502: Thank you: moved to `Netlist.v`

cava/cava/Cava/Monad/Cava.v
```diff
@@ -228,43 +221,152 @@ Definition setModuleName (name : string) : state CavaState unit :=
...
+
+Compute numberPort 0 (mapShape snd tupleOfTuples).
+
+(* This gives:
+
+     = ([0%N; 1%N; 2%N; 3%N; 4%N; 5%N; 6%N; 7%N], [8%N; 9%N; 10%N; 11%N; 12%N; 13%N; 14%N; 15%N],
+       ([16%N; 17%N; 18%N; 19%N; 20%N; 21%N; 22%N; 23%N], [24%N; 25%N; 26%N; 27%N; 28%N; 29%N; 30%N; 31%N]))
+     : signalTy N (mapShape snd tupleOfTuples)
+
+Which is wrong because it has the structure (a, b, (c, d)) instead of ((a,b), (c, d)).
```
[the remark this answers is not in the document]
satnam6502: Will produce a stripped down example, repo, and ask question on Coq mailing list.

cava/cava/Cava/Netlist.v
satnam6502: OK, let's discuss this F2F.

satnam6502:
Thank you! Updating and merging.

PR #79, opened 2020-04-21
Scalability tests using large adder trees (synthesizable and direct Xilinx fast-carry chain versions)

satnam6502:
This PR adds two sets of scalability tests based on adder trees: one type uses a synthesized adder, and the other type is based on a fast-carry chain Xilinx adder which is explicitly instantiated. Simulation of Xilinx-specific netlists with Verilator is now supported. Some re-roganization and refactoring has also been done.


satnam6502:
A picture of the FPGA layout of the 64 input adder of 128-bit values.
Screen Shot 2020-04-21 at 1 20 42 AM

PR #80, opened 2020-04-21
Update Arrow/Instances/Netlist.v for the new netlist types

blaxill:
This updates arrows for the new netlist types and adds a few functions port functions.

I also moved your existing `numberBitVec` `numberPort` functions into Netlist.v so they can be used for arrows too. I added a `flattenBitVec` function as I was having trouble getting the classes to resolve when the value I had was not concrete/was dependent on a sub call, but that might have been a limitation in my Coq understanding. 

~~There is one issue left, the inputs/outputs aren't connected (see below), I couldn't see how to connect them up with mkModule now- @satnam6502 do I need to use the new CircuitInterface for this? thanks~~ I think I know what to do, working on this now

~~~
module xorArrow(
  input logic input1,
  input logic input2,
  output logic output1
  );

  timeunit 1ns; timeprecision 1ns;

  logic[11:0] net;

  // Constant nets
  assign net[0] = 1'b0;
  assign net[1] = 1'b1;
  // Populate vectors.

  not inst_1 (net[11],net[10]);
  and inst_2 (net[10],net[9],net[7]);
  not inst_3 (net[9],net[8]);
  and inst_4 (net[8],net[5],net[3]);
  not inst_5 (net[7],net[6]);
  and inst_6 (net[6],net[5],net[2]);
  not inst_7 (net[5],net[4]);
  and inst_8 (net[4],net[2],net[3]);

endmodule
~~~

### DIFF
--- a/cava/arrow-examples/ArrowExamples.v
+++ b/cava/arrow-examples/ArrowExamples.v
@@ -16,32 +16,21 @@ Require Import Cava.Arrow.Instances.Coq.
 Require Import Cava.Arrow.Instances.Netlist.
 Require Import Cava.Arrow.Instances.Stream.
 
-(* Implement a NAND circuit by composing an AND gate and INV gate. *)
-Definition nand
-  {Cava: Cava}
-  := and_gate >>> not_gate.
+Require Import Nand.
+Require Import Xor.
+Require Import FeedbackNand.
 
-(* An implementation of an XOR gate made out of the NAND circuit
-   defined above. *)
-Definition xor
-  {_: Cava}
-  : (bit**bit) ~> bit :=
-  copy
-  >>> first (nand >>> copy)                    (* ((nand,nand),(x,y)) *)
-  >>> assoc                                    (* (nand,(nand,(x,y))) *)
-  >>> second (unassoc >>> first nand >>> swap) (* (nand,(y, x_nand)) *)
-  >>> unassoc >>> first nand                   (* (y_nand,x_nand) *)
-  >>> nand.
+Local Open Scope string_scope.
 
 Definition twoBits
-  {Cava: Cava}
+  `{Cava}
   : unit ~> (bit**bit) :=
   copy
   >>> first (constant true)
   >>> second (constant false).
 
 Definition twoBools
-  {Cava: Cava}
+  `{Cava}
   (x y: bool): unit ~> (bit**bit) :=
   copy
   >>> first (constant x)
@@ -49,13 +38,6 @@ Definition twoBools
 
 Definition nandb : bool -> bool -> bool := fun a b => negb (a && b).
 Definition uncurry {a b c} (f: a -> b -> c) : (a*b) -> c := fun xy => f (fst xy) (snd xy).
-
-(*nand previous output and current input, output delayed 1 cycle*)
-Definition loopedNand
-  {Cava: CavaDelay}
-  {ArrowLoop: @ArrowLoop (@cava_delay_arr Cava)}
-  : bit ~> bit :=
-  loopl (nand >>> delay_gate >>> copy).
 
 Existing Instance CoqArr.
 Existing Instance CoqCava.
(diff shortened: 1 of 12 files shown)


cava/arrow-examples/ArrowExamples.v
```diff
@@ -16,22 +16,9 @@ Require Import Cava.Arrow.Instances.Coq.
...
-  >>> assoc                                    (* (nand,(nand,(x,y))) *)
-  >>> second (unassoc >>> first nand >>> swap) (* (nand,(y, x_nand)) *)
-  >>> unassoc >>> first nand                   (* (y_nand,x_nand) *)
-  >>> nand.
+Require Import Nand.
+Require Import Xor.
+Require Import FeedbackNand.
 
 Definition twoBits
   {Cava: Cava}
```
satnam6502, line 24:
```
  `{Cava}
```
would work too.

cava/arrow-examples/ArrowExamples.v
```diff
@@ -16,22 +16,9 @@ Require Import Cava.Arrow.Instances.Coq.
...
-  >>> first (nand >>> copy)                    (* ((nand,nand),(x,y)) *)
-  >>> assoc                                    (* (nand,(nand,(x,y))) *)
-  >>> second (unassoc >>> first nand >>> swap) (* (nand,(y, x_nand)) *)
-  >>> unassoc >>> first nand                   (* (y_nand,x_nand) *)
-  >>> nand.
+Require Import Nand.
+Require Import Xor.
+Require Import FeedbackNand.
 
 Definition twoBits
```
satnam6502, line 23: I asked Coq to report the type for `twoBits` and got:
```
twoBits = 
fun H : Cava => copy >>> first (constant true) >>> second (constant false)
     : forall H : Cava, unit ~[ cat ]~> bit ** bit
```
Can you explain what this type means, especially the `[cat]`?

cava/arrow-examples/ArrowExamples.v
```diff
@@ -151,28 +131,27 @@ Section StreamProofs.
...
+  Definition xorArrow := arrowToHDLModule
+
     "xorArrow"
     (@xor NetlistCava)
-    (fun '(l,r) =>
-      [ mkPort "input1" Bit
-      ; mkPort "input2" Bit
-      ])
-    (fun o => [mkPort "output1" Bit]).
+    ("input1","input2")%string
```
satnam6502, line 146: Adding a:
```
```
earlier would help to reduce some syntax noise by allowing this to be written as:
```
("input1","input2")
```

cava/arrow-examples/Xor.v
```diff
@@ -0,0 +1,15 @@
...
+   defined above. *)
+Definition xor
+  {_: Cava}
+  : (bit**bit) ~> bit :=
+  copy
+  >>> first (nand >>> copy)                    (* ((nand,nand),(x,y)) *)
+  >>> assoc                                    (* (nand,(nand,(x,y))) *)
+  >>> second (unassoc >>> first nand >>> swap) (* (nand,(y, x_nand)) *)
+  >>> unassoc >>> first nand                   (* (y_nand,x_nand) *)
+  >>> nand.
```
satnam6502, line 15: That is a pretty mind-bending description of an XOR gate made out of NAND gates :-)

cava/cava/Cava/Arrow/Instances/Netlist.v
```diff
@@ -223,26 +197,93 @@ Section NetlistEval.
...
-    (mkInputs: shapeToNumbered X -> list PortDeclaration)
-    (mkOutputs: shapeToNumbered Y -> list PortDeclaration)
-    : (Netlist.Module * N) :=
-      let (i, n) := @numberShape X 2 in
-      let '(o, (nl, count)) := runState (f i) ([], n) in
-      (mkModule name nl (mkInputs i) (mkOutputs o), count).
+    (input_names: nameTy X)
+    (output_names: nameTy Y)
+    : Netlist.CavaState :=
+      let i := numberPort 2 X in
```
satnam6502, line 274: I guess we should make a constant for this offset at some point.

cava/arrow-examples/ArrowExamples.v
blaxill: `unit ~[ cat ]~> bit ** bit` is an arrow from `unit` to `bit**bit`. The arrows are dependent on some category which comes implicitly from whichever Cava instance you instantiate. 

In this case the only category in scope is from the Cava class, but since the Cava class could be any instance the category could also be any instance and so it displays a compiler-produced name. 

Which category instance you are using is not important unless you are defining a Cava instance, but if you ask Coq the arrow type it will display the `[cat]` in the arrow to be explicit.

PR #82, opened 2020-04-22
Add support for wiring up debug ports for logic analyzer.

satnam6502:
Also other changes to make synthesized test-benches work with logic analyzer.
The test-benches can now be synthesized and run on Xilinx FPGA and their inputs and outputs observed with the ILA logic anlayzer core.

image


satnam6502:
Self-merging to unblock other PRs that will follow on for this, and it's already pretty late in the UK.

PR #83, opened 2020-04-28
Syntax and compilation to arrows

blaxill:


satnam6502:
Some of the lemmas that can be proved by using just `lia` you might like to consider just performing "inline" in your proof rather than breaking out a separate lemma for them.

PR #84, opened 2020-05-01
Proofs relating the col combinator to the below combinator

satnam6502:
These proofs tell us general facts about the properties of `col` and how it relates to `below` which we can then use for proofs of col-based circuits e.g. a ripple-carry adder.


satnam6502:
Self-merging to make progress during the weekend.

PR #85, opened 2020-05-02
Add a _CoqProject to use for vscoq for Visual Studio Code

satnam6502:
@blaxill as part of this change I put the examples (including the arrow examples) into their own libraries with a name other than the default `Lib`.
Now we can start Visual Studio code from the top level caca directory and the `_CoqProject` there makes all the cava source and the library source available in Visual Studio code.


satnam6502:
Self-merging so I can make further dependent changes this weekend.

PR #86, opened 2020-05-02
Convert compute in Xilinx adder tree example to Example as regression test

satnam6502:
Minor cleanup for Xilinx examples.


satnam6502:
I forgot, we can't run the Xilinx UNISIM based SystemVerilog test files in CI yet because we have not cloned the simulation sources for the UNISIM. Removing the `small` target.

satnam6502:
Self-merging clean-up.

PR #87, opened 2020-05-05
Switch arrow tests to produce Verilator testbenches

satnam6502:
This PR makes the arrow examples automatically generate Verilator testbenches (and runs them to ensure they pass). We no longer have a dependency on `iverilog` which can be removed in a subsequent PR.


blaxill:
Looking good, thanks Satnam

PR #92, opened 2020-05-07
Emit kappa variable lookup arrow directly

blaxill:
Emit environment to variable morphism directly and also mechanize the proofs


satnam6502:
This is well beyond my current ability to review in any detailed way, so I've had a superficial look (which looks fine) and defer to your judgement that this is all good. Thanks!

PR #93, opened 2020-05-07
Improve kappa syntax 

blaxill:


cava/arrow-examples/ArrowExamples.v
```diff
@@ -39,8 +39,25 @@ Definition twoBools
...
+  <[ \ cin ab =>
+     let a = fst' ab in
+     let b = snd' ab in
+     let part_sum = !xor_gate a b in
+     let sum      = !xorcy part_sum cin in
+     let cout     = !muxcy part_sum cin a in
+     (sum, cout)
+  ]>.
+
+Definition xilinxFullAdder' `{Cava} := Closure_conversion xilinxFullAdder.
```
satnam6502, line 55: So we have to write `Closure_conversion` for each definition....

satnam6502:
Again, I only have superficial comprehension, but let's get this in for now and I shall catch up via doing some examples myself.

cava/arrow-examples/ArrowExamples.v
blaxill: Closure conversion is changing from kappa to arrow form. This could be done in the notation but it might be useful to have access to the kappa form. There was also a performance impact doing it directly in notation which could likely be avoided and is likely related to implicit parameters.

cava/cava/Cava/Arrow/Instances/Combinational.v
```diff
@@ -0,0 +1,143 @@
...
+    unassoc _ _ _ '(x,(y,z)) := ((x,y),z);
+  }.
+  Proof.
+    simple_solve.
+    simple_solve.
+    simple_solve.
+    simple_solve.
+    simple_solve.
+    simple_solve.
+    simple_solve.
```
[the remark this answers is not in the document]
blaxill: This unfortunately doesn't work as they are distinct goals, there might be another way but I'm not sure of one at the moment.

blaxill:
Fixes #44

cava/arrow-examples/ArrowExamples.v
blaxill: I'll address notation issues as we progress with OpenTitan components

PR #97, opened 2020-05-07
Implement and test LUT3

satnam6502:
Add Cava support for LUT3 and a test to make sure the Verilog UNISIM simulation matches the behaviour predicted by the Cava axioms.


satnam6502:
Self-merging LUT3 support since this is not likely to be controversial.

PR #113, opened 2020-05-15
Implement muxing of an array of buses

satnam6502:

This implements the muxing of an array of buses. Later, we should generalize this to allow for muxing of elements that are not just a flat bus. Also removed the `WIDTH` waiver from Verilator config files (this prevented me from spotting a bug) and added a few consequent fixes.


satnam6502:
Addresses #113

PR #120, opened 2020-05-20
SystemVerilog bench for arrow mux2_1 example

satnam6502:

This adds a test bench for the arrow 2 to 1 mux example. It's a shame that the `tt` has to be added when computing the expected results but hopefully this will not be required in a future version.


blaxill:
> It's a shame that the `tt` has to be added when computing the expected results but hopefully this will not be required in a future version.

This has given me a clearer idea of how to address this. Previously I was thinking of adding object equivalences to the arrow class, but actually addressing this in each instance makes sense and should be straightforward...

cava/arrow-examples/Mux2_1.v
```diff
@@ -35,8 +39,36 @@ Definition mux2_1'
 
 Local Open Scope string_scope.
 
+(* TODO: Two different interfaces for now, need to rationalize this down
```
blaxill, line 42: Yeah, I'll address this soon and make the arrow netlist use `mkCircuitInterface`

PR #124, opened 2020-05-26
Refactor circuit-representable types out of Netlist.v

blaxill:
- Part of the motivation here is to start unifying some of the types that i've duplicated in various arrow places.
- I've rolledback the well-founded recursion fixpoint i demonstrated
- I've added experimental  "<< x , y, .., z >>" tuple notation

### DIFF
--- a/cava/arrow-examples/ArrowExamples.v
+++ b/cava/arrow-examples/ArrowExamples.v
@@ -10,6 +10,7 @@ From Coq Require Import ZArith.
 From Coq Require Import btauto.Btauto.
 
 Require Import Cava.Netlist.
+Require Import Cava.Types.
 
 Require Import Cava.Arrow.Arrow.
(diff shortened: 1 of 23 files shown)


blaxill:
More notes (some of these changes are opinionated and I welcome discussion):

- I changed `portTy` to `Kind`, which I believe makes sense. 
- `@shape Kind` is the most common usage of shape and so Ive added `Notation bundle := (@shape Kind).`. The name bundle is not very descriptive- if this is a problem perhaps `KindBundle` or similar would be better? 

Not being addressed now but at some point in the future I would like to
- Remove the ad-hoc laws from the Arrow class, and replace with the correct ones, and move any extras to the Cava class
- Remvoe the Cava parts of the constructive arrow instance and make the type parametric on a Cava instance. (This would allow the type being used for non-cava arrows, for which I can think of a few use cases in the far future...)

PR #128, opened 2020-05-29
Convert the wire representation from N to Signal

satnam6502:
This PR changes the wire representation from `N` to the structured `Signal` type which helps us generate better looking SystemVerilog and sets us up better for vector slicing expressions.


satnam6502:
Well, I think there is going to have to be more Haskell if we want to produce higher quality SystemVerilog that can pass muster with human readers and lint tools. I'd like to consider the formal verification of the netlist back-end not to be in scope for the first phase of Silver Oak because there is so much challenging and high value work to be done at the higher level representation side.

PR #132, opened 2020-05-30
Simplify Arrow.Netlist

blaxill:
- Remove 'N' from state, as it is no longer needed using the new Signal type.
- Move `state CavaState` routines from monadic Cava into root Netlist to
and reuse them in Arrow.Netlist.
- Remove redundant Arrow.Netlist interface code and use CircuitInterface.


cava/cava/Cava/Netlist.v
```diff
@@ -201,42 +201,104 @@ Definition newWires (width : nat) : state CavaState (list Signal) :=
...
   | mkCavaState o isSeq (mkModule name insts inputs outputs)
     => put (mkCavaState o isSeq (mkModule name (newInst::insts) inputs outputs))
   end.
 
+Fixpoint addInstances (insts: list Instance) : state CavaState unit :=
+  match insts with
+  | [] => ret tt
+  | x :: xs => addInstance x >>= fun _ => addInstances xs
+  end.
+
```
satnam6502, line 237:
```
Definition addInstances := map addInstance.
```

satnam6502: Additional note, if you want to write it out like this then you can use `;;`:
```
Fixpoint addInstances (insts: list Instance) : state CavaState unit :=
  match insts with
  | [] => ret tt
  | x :: xs => addInstance x ;;
               addInstances xs
  end.
```

cava/cava/Cava/Types.v
```diff
@@ -97,6 +97,15 @@ Fixpoint flattenShape {A} (s : @shape A) : list A :=
   | Tuple2 t1 t2 =>  flattenShape t1 ++ flattenShape t2
   end.
 
+Fixpoint removeTerminal {A} (s: @shape A): shape :=
+  match s with
+  | Empty => Empty
+  | One thing => One thing
+  | Tuple2 t1 Empty => t1
+  | Tuple2 t1 t2 =>  Tuple2 t1 (removeTerminal t2)
+  end.
```
satnam6502, line 106: Does this remove the last `unit`-typed value from the input type? It might be wroth writing that in a comment. Can you think of another name? `dropUnit` or some such thing? Fine to leave it as `removeTerminal` if nothing else comes to mind.

cava/cava/Cava/Types.v
```diff
@@ -211,3 +220,28 @@ Fixpoint numberPort (i : N) (inputs: bundle) : signalTy Signal inputs :=
...
+  - exact s.  
+  - rewrite reduceSignal in s.
+    refine ((fst s, _)).
+    apply snd in s.
+    apply IHA2.
+    apply s.
+Defined.
+
+Fixpoint contraRemoveTerminal {A B} (f: signalTy Signal A -> B): signalTy Signal (removeTerminal A) -> B :=
+  fun a => f (addTerminal a).
```
satnam6502, line 247: Please can you explain in a comment what this does? Does it do the opposite of `removeTerminal` in some way?

cava/cava/Cava/Netlist.v
blaxill: Thanks, I think my thought process had timed out- changed to use `;;` now. To map directly I'd need something like `sequence_ (map ...` or `traverse`, but various items seemed to be missing from coq-ext-lib.

cava/cava/Cava/Types.v
blaxill: Changed to `withoutRightmostUnit` and added comment

cava/cava/Cava/Types.v
blaxill: Sorry this was confusing, I was trying to think of a way to have different names for removing units from types vs terms vs mapping those functions over an input/output, but I just made it more confusing. 

I've changed the name to `removeRightmostUnit` and added a comment.

cava/arrow-examples/ArrowAdder.v
```diff
@@ -75,13 +71,14 @@ Definition fullAdder_tb_inputs :=
    (true, (true, true))
 ].
 
-Definition fullAdder `{Cava}: (bit ** (bit ** bit) ** unit) ~> bit ** bit
-   := toCava _ (Closure_conversion fullAdder').
+Definition fullAdder_netlist :=
+  makeNetlist fullAdderInterface
+    (removeRightmostUnit (fullAdder NetlistCava)).
```
satnam6502, line 76: Perhaps time for a `makeArrowNetlist` that does the `removeRightmostUnit` under the hood?

cava/cava/Cava/Arrow/Instances/Netlist.v
```diff
@@ -202,140 +201,30 @@ Section NetlistEval.
...
-
-      (* 2 *)
-      '(y,n') <- f (x,n) ;;
-
-      (* 3 *)
-      let links := linkShapes AssignBit N n n' in
-      '(nl, i) <- get ;;
-      put (links ++ nl, i) ;;
-
+    loopr _ _ Z f x :=
```
satnam6502, line 205: I found `Z` confusing here because at first I thought you meant the integer type `Z`.

blaxill: Ah yes, it would have been clearer if I had named the other arguments, as it should be `X Y Z`. I think I might try to collect some style notes and apply them in the not too distant future.

PR #135, opened 2020-06-01
Arrow version of below and col combinators

satnam6502:
Work of @blaxill


satnam6502:
I am going to abandon this for now.

PR #136, opened 2020-06-02
Add LUTs and multidimensional indexing to Arrows

blaxill:
Currently this generates a dummy Netlist instance `IndexAlt`. After switching Netlist `list`s back to vectors an indexing function will be created and this netlist constructor can be removed.

### DIFF
--- a/cava/cava/Cava/Arrow/Arrow.v
+++ b/cava/cava/Cava/Arrow/Arrow.v
@@ -170,7 +170,7 @@ Class Cava  := {
   buf_gate:  bit        ** unit ~> bit;
 
   xorcy:     bit ** bit ** unit ~> bit;
-  muxcy:     bit ** bit ** bit ** unit ~> bit;
+  muxcy:     bit ** (bit ** bit) ** unit ~> bit;
   
   unsigned_add a b c: bitvec [a] ** bitvec [b] ** unit ~> bitvec [c];
 
--- a/cava/cava/Cava/Arrow/Arrow.v
+++ b/cava/cava/Cava/Arrow/Arrow.v
@@ -138,6 +138,11 @@ Class ArrowLoop (A: Arrow) := {
   loopl {x y z} : (z**x ~> z**y) -> (x ~> y);
 }.
 
+(*
+replicate_object returns an object in the argument format expect by Kappa Calculus.
+That is, the object is repeated n times, as a right nested tuple, with the rightmost
+tuple element set to the unit type.
+*)
 Fixpoint replicate_object `{Arrow} (n: nat) (o: object) :=
 match n with 
 | O => unit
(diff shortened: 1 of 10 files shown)


cava/cava/Cava/Arrow/Arrow.v
```diff
@@ -139,12 +138,24 @@ Class ArrowLoop (A: Arrow) := {
   loopl {x y z} : (z**x ~> z**y) -> (x ~> y);
 }.
 
+Fixpoint replicate_object `{Arrow} (n: nat) (o: object) :=
```
satnam6502, line 141: Does it ever make sense to call `replicate_object` with `0`? Perhaps a comment?

cava/cava/Cava/Arrow/Arrow.v
```diff
@@ -160,12 +171,36 @@ Class Cava  := {
 
   xorcy:     bit ** bit ** unit ~> bit;
   muxcy:     bit ** bit ** bit ** unit ~> bit;
+
```
satnam6502, line 174: I know it's not part of this review, and a very minor point, but using tuples does help to express what is the control input to the mux and what are the inputs.

cava/cava/Cava/Arrow/Arrow.v
```diff
@@ -160,12 +171,36 @@ Class Cava  := {
...
   xorcy:     bit ** bit ** unit ~> bit;
   muxcy:     bit ** bit ** bit ** unit ~> bit;
+  
+  unsigned_add a b c: bitvec [a] ** bitvec [b] ** unit ~> bitvec [c];
+
+  lut n: (bool^^n --> bool) -> (replicate_object n bit) ~> bit;
+
+  degenerate_bitvec xs := match xs with | [] => bit | _ => bitvec xs end;
+
+  index_array n xs
```
satnam6502, line 181: So you want to have support for array indexing (and slicing) at the netlist layer?

cava/cava/Cava/Netlist.v
```diff
@@ -135,6 +135,11 @@ Inductive Instance : Type :=
   (* Multiplexors *)
   | IndexBitArray: list Signal -> list Signal -> Signal -> Instance
   | IndexArray: list (list Signal) -> list Signal -> list Signal -> Instance
+  | IndexAlt: forall sz szs,
```
satnam6502, line 138: Does this then work for SystemVerilog netlist generation?

cava/cava/Cava/Arrow/Arrow.v
blaxill: I'm using this function to replicate the object in a way that fits the argument format expected by kappa calculus (right imbalanced tupling), I've added a comment.

cava/cava/Cava/Arrow/Arrow.v
blaxill: Done thanks

cava/cava/Cava/Arrow/Arrow.v
blaxill: No, I'll add a function to create the indexing manually, but right now I;ve added a Netlist.Instance constructor as a placeholder.

PR #138, opened 2020-06-03
Support negative edges for clock and reset

satnam6502:

The circuit interface mechanism is tweaked to add information and the names and edge-behaviour of clock and reset inputs.


satnam6502:
Self-merging to unblock other PRs during London night.

PR #140, opened 2020-06-04
Arrows cleanup

blaxill:
The aim of this in progress PR is to:

- [x] - Pay closer adherence to generalized arrow class hierarchy
- [x] - Separate theoretically sound arrow laws and adhoc circuit equivalences
- [x] - Merge cava loops and delay into single arrow class- Having separate classes for Cava requires producers (e.g. kappa constructions) to make a distinction between what type of circuit they are creating which adds complexity for no advantage (things requiring combinational circuits at the arrow class level can recover the distinction through propositions)  
- [x] - Rewrite combinational evaluator to allow non-combinational circuits and bad indexing, and a safe evaluation method that takes a well formed prop to prevent these components.
- [x] - Revert rightmost unit arrows in cava class
- [x] - Provide cava arrow class methods equivalents directly as constructors in kappa class. Combination of kappa fragments should become the main kappa circuit composition strategy, although lifting arrows methods should still be supported
- [x] - Simplify syntax


satnam6502:
How about squashing them into one commit so it is easier to review?

blaxill:
> How about squashing them into one commit so it is easier to review?

Done

cava/arrow-examples/ArrowAdder.v
```diff
@@ -17,42 +17,44 @@
...
+Section definition.
+  Import KappaNotation.
+
+  Definition halfAdder
+  : Kappa_sugared << Bit, Bit, Unit >> <<Bit, Bit>> :=
+  <[ \ a b =>
+    let part_sum = xor a b in
+    let carry = and a b in
+    (part_sum, carry)
+  ]>.
```
satnam6502, line 36: Nice! No more `!`s here!

cava/cava/Cava/Arrow/Instances/Combinational.v
```diff
@@ -11,170 +13,203 @@ Require Import Cava.Types.
...
+Definition wf_combinational {x y} (circuit: x ~> y) := forall i, {o | circuit i = Some o}.
+Definition evaluate {x y} (circuit: x ~> y) (wf: wf_combinational circuit) (i: x): y.
+Proof.
+  cbv [morphism CoqCat wf_combinational] in *.
+  specialize (wf i).
+  inversion wf.
+  apply x0.
+Defined.
+
+Lemma not_gate_wf: wf_combinational not_gate.
```
satnam6502, line 188: I assume likewise for the other combinational gates.

cava/cava/Cava/Arrow/Kappa/Syntax.v
```diff
@@ -1,137 +1,210 @@
...
+  Notation "x y" := (App x y) (in custom expr at level 3, left associativity) : kappa_scope.
 
-Notation "! x" := ( (Arr x)) (in custom expr at level 2, x global) : expression_scope.
+  Notation "x" := (Var x) (in custom expr, x ident) : kappa_scope.
+  Notation "( x )" := x (in custom expr, x at level 4) : kappa_scope.
+  Notation "'let' x = e1 'in' e2" := (Let e1 (fun x => e2)) (in custom expr at level 1, x constr at level 4, e2 at level 5, e1 at level 1) : kappa_scope.
 
-Notation "~!( x )" := (Arr (Closure_conversion x)) (in custom expr, x constr) : expression_scope.
+  (* todo: turn into a recursive pattern *)
+  Notation "'let' '( x , y ) = e1 'in' e2"
```
satnam6502, line 35: A lot of notation here... I wonder how confusing it is to define a notation for `let`?

cava/cava/Cava/Arrow/Kappa/Syntax.v
```diff
@@ -1,137 +1,210 @@
...
+    (n: nat)
+    (f:  Kappa_sugared << A, A, Unit >> A)
+    {struct n}
+    : Kappa_sugared << copy_object_pow2 A n, Unit >> A :=
+  match n with
+  | O => <[ \x => x ]>
+  | S n' =>
+    <[\ x =>
+        let a = !(tree A n' f) (fst x) in
+        let b = !(tree A n' f) (snd x) in
```
satnam6502, line 145: As mentioned before, an escape is still needed here, I assume because `tree` is from an "earlier phase" that needs to be evaporated first.

cava/cava/Cava/Arrow/Instances/Combinational.v
blaxill: Yes, currently this is just an example. `wf_combinational` is also very experimental at this stage, I see splitting this into simpler propositions e.g. `loopless` `wf_indexing` such that

`delayless -> loopless -> wf_indexing -> wf_combinational` or even more lax:

`delayless -> all loops converge -> wf_indexing -> wf_combinational`

cava/cava/Cava/Arrow/Kappa/Syntax.v
blaxill: I'd argue that the angled brackets delimiting kappa code should be enough to indicate that you're leaving regular Coq world, but it also should be easier to revisit all the notation with the underlying classes/types now in a better place. 

Currently the notation defines an explicit custom expression scope that separates regular Coq expressions from the kappa code. This might not be the best way.

cava/cava/Cava/Arrow/Kappa/Syntax.v
blaxill: The escape will let you access anything in the large Coq scope. As you saw there is notation (and also kappa_sugared constructors) for Cava/common functionality, but you could also escape them directly (although the expressions would be ugly).

PR #149, opened 2020-06-18
Add AES canright and canright_masked_noreuse sboxes

blaxill:
- Currently compiles but is untested

### DIFF
--- a/cava/arrow-examples/Aes/Util.v
+++ b/cava/arrow-examples/Aes/Util.v
@@ -2,13 +2,15 @@
 Require Import Cava.Arrow.Kappa.Syntax.
 Require Import Cava.Arrow.Arrow.
 
+Require Import Cava.Arrow.Instances.Prop.
+
 From Coq Require Import NArith.
 
 Import KappaNotation.
 
 Section var.
 
-Context {var: Kind -> Kind -> Type}.
+Variable var: Kind -> Kind -> Type.
 
 Fixpoint reduce {n T}
   (f : kappa_sugared var <<T, T, Unit>> T)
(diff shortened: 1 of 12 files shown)


cava/arrow-examples/Aes/Util.v
```diff
@@ -24,18 +26,58 @@ match n with
...
\ No newline at end of file
+Definition map2_xor {n}
+  : kappa_sugared var << Vector (S n) Bit, Vector (S n) Bit, Unit >> <<Vector (S n) Bit>> :=
+  map2 Xor.
+
+Definition map2_xnor {n}
+  : kappa_sugared var << Vector (S n) Bit, Vector (S n) Bit, Unit >> <<Vector (S n) Bit>> :=
+  map2 Xnor.
+
+Definition bitvec_eq {n}
```
satnam6502, line 62: Worthy a comment like "We check to see if two bit-vectors for equal by XNORs corresponding bits to check for equality and then folding AND to form the conjunction of each bit-level test."

cava/arrow-examples/Aes/Util.v
```diff
@@ -24,18 +26,58 @@ match n with
...
+  : kappa_sugared var << Vector (S n) Bit, Vector (S n) Bit, Unit >> <<Bit>> :=
+  <[\ x y => !(reduce And) (!map2_xnor x y) ]>.
+
+Definition enable_vec {n}
```
satnam6502, line 66: Perhaps a comment like "if the enable input is 1 then the vector is return as is, otherwise a vector of corresponding size is returned with all elements masked out to 0".

cava/arrow-examples/Aes/canright.v
```diff
@@ -0,0 +1,93 @@
...
+      let a = (gamma[7:4]) ^ (gamma[3:0]) in
+      let b = !aes_mul_gf2p4 (gamma[7:4]) (gamma[3:0]) in
+      let c = !aes_square_scale_gf2p4_gf2p2 a in
+      let d = !aes_inverse_gf2p4 (c ^ b) in
+
+      concat 
+        (!aes_mul_gf2p4 d (gamma[7:4]))
+        (!aes_mul_gf2p4 d (gamma[3:0]))
+  ]>.
+
```
satnam6502, line 63: Nice work with the notation.

cava/cava/Cava/Arrow/Arrow.v
```diff
@@ -63,18 +63,20 @@ Add Parametric Relation (object: Type) (C: Category object) (x y: object): (morp
...
 Add Parametric Morphism (object: Type) (C: Category object) (x y z: object) : (@compose object C x y z)
   with signature (morphism_equivalence _ _ ==> morphism_equivalence _ _ ==> morphism_equivalence _ _)
   as parametric_morphism_comp.
-Proof. auto. Defined.
+Proof. 
+  Hint Extern 1 => apply compose_respects_equivalence: core.
+  Hint Extern 1 => apply id_left: core.
+  Hint Extern 1 => apply id_right: core.
+  auto.
+Defined.
```
satnam6502, line 74: Ah, so these hints are defined just locally? And the `auto` makes repeated use of them until the proof is done I assume.

cava/cava/Cava/Arrow/Arrow.v
```diff
@@ -193,6 +197,13 @@ match v with
 | x::xs => (x, vec_to_nprod A _ xs)
 end%vector.
 
+Definition log2_up_min_1 (n: nat): nat :=
```
satnam6502, line 200: A comment about why this function is needed would be nice (for vector indexing I assume).

cava/arrow-examples/Aes/Util.v
blaxill: Done thanks.

cava/arrow-examples/Aes/Util.v
blaxill: Done

cava/cava/Cava/Arrow/Arrow.v
blaxill: The intention was for them to only be used locally, but actually they survive past the definition, so I've reverted this change.

cava/cava/Cava/Arrow/Arrow.v
blaxill: Yeah, currently at the Kappa level I've somewhat banished 0-sized vectors (it may not be the best choice), so this function prevents 0-widths for indices to 1-length vectors. This might actually increase resource although I'd hope synthesis tools would trim it.

PR #150, opened 2020-06-22
Remove inductive type definition for closure conversion environment and use lists

blaxill:


blaxill:
@satnam6502 line change is misleading- the changes to files in Aes directory are mostly changes to make add consistent indentation when using sections.

cava/arrow-examples/Aes/Pkg.v
```diff
@@ -8,43 +8,37 @@ Import KappaNotation.
...
-Arguments CIPH_INV {var n}. 
\ No newline at end of file
+  Context {var: Kind -> Kind -> Type}.
+
+  (* typedef enum logic {
+    CIPH_FWD = 1'b0,
+    CIPH_INV = 1'b1
+  } ciph_op_e; *)
+  Definition CIPH_FWD {n} : kappa_sugared var Unit (Vector n Bit) := <[ #0 ]>.
+  Definition CIPH_INV {n} : kappa_sugared var Unit (Vector n Bit) := <[ #1 ]>.
```
satnam6502, line 18: So this is mimicking a struct? What does the `n` parameter get used for?

cava/arrow-examples/Aes/canright.v
```diff
@@ -10,84 +10,86 @@ Require Import pkg.
...
-  : kappa_sugared var <<Vector 8 Bit, Unit>> (Vector 8 Bit) :=
-  <[\ gamma =>
-      let a = (gamma[7:4]) ^ (gamma[3:0]) in
-      let b = !aes_mul_gf2p4 (gamma[7:4]) (gamma[3:0]) in
-      let c = !aes_square_scale_gf2p4_gf2p2 a in
-      let d = !aes_inverse_gf2p4 (c ^ b) in
+  Program Definition aes_inverse_gf2p4
+    : kappa_sugared var <<Vector 4 Bit, Unit>> (Vector 4 Bit) :=
+    <[\ gamma =>
+        let a = (gamma[3:2]) ^ (gamma[1:0]) in
```
satnam6502, line 29: So the indexing/slicing now lines up with the SystemVeriloig version?

cava/arrow-examples/Aes/Pkg.v
blaxill: This is just emulating the enum as constants, `n` is used to allow the constants to be generic over any bitwidth, which is likely not necessary. I have some ideas for introducing more types inc. sum types in the future.

cava/arrow-examples/Aes/canright.v
blaxill: I believe so, although I think switching the indexing order might be preferable. https://github.com/project-oak/oak-hardware/issues/40#issuecomment-644231004

PR #151, opened 2020-06-23
Use the usual Fst Snd naming convention in Kappa

blaxill:
Change the type of `append` from `n + 1` to `S n`. `S n` is
definitionally correct and so requires less work in proving type
equality.


cava/cava/Cava/Arrow/Arrow.v
```diff
@@ -193,10 +193,10 @@ Notation "<< x >>" := (x) : kind_scope.
...
 
 Fixpoint vec_to_nprod (A: Type) n (v: Vector.t A n): A^n :=
-match v with
-| [] => tt
-| x::xs => (x, vec_to_nprod A _ xs)
-end%vector.
+  match v with
+  | [] => tt
+  | x::xs => (x, vec_to_nprod A _ xs)
+  end%vector.
```
satnam6502, line 200: I find it hard to see what the change is here.... just indentation I think!

blaxill: You're correct, I'm making small consistency changes as I go

PR #153, opened 2020-06-24
Make wf_combinational non-informative

blaxill:
This weakens the requirement for `wf_combinational` in arrows, and allows opaque proofs of circuits being combinational.

### DIFF
--- a/cava/cava/Cava/Arrow/Instances/Combinational.v
+++ b/cava/cava/Cava/Arrow/Instances/Combinational.v
@@ -11,6 +11,8 @@ Require Import Cava.Arrow.Instances.Prop.
 (* Evaluation as function evaluation, no delay elements or loops              *)
 (******************************************************************************)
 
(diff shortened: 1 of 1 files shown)


blaxill:
An opaque proof is one that ends with Qed. 

Using dependent pairs e.g. `forall i, {o | circuit i = Some o}` carries not only the proof but the object  that satisfies the proof and so is informative, whereas `forall i, exists o, circuit i = Some o` is only the proof that some object satisfies the proof and so is non-informative. That is, in the second case there is no direct way to access the object but might be easier to prove. Also, the first definition must be created with a transparent proof to access the object.

blaxill:
I have a few places where I'm depending on transparent proofs and this removes one of them. The aim being to completely remove the requirement for transparent proofs where possible as the proofs themselves may rely on opaque proofs and then will not be computable.

PR #155, opened 2020-06-26
Add dynamic indexing and static slicing to the netlist layer.

satnam6502:
In order to support indexing and slicing in a way that will lead to good SystemVerilog extraction the following changes have been made:

* Indexing and slicing are now part of the `Signal` type and are viewed as parts of expressions rather than an instantiation of a "multiplexor circuit".
* The `Signal` type now carries a type argument which specified the type of information flowing over a `Signal` bus.
* The computed netlists is analyzed and re-written to name vector literals in contexts where they would be illegal in SystemVerilog e.g. in indexing positions and as vectors to be indexed.
* The list-based combinators have been re-written with `Vector.t` based versions instead. There are still some significant issues to be resolved but the current formulation works for the all the monad kernel examples.
* The arrow netlist layer is disconnected and I shall ask @blaxill to fix it up and re-integrate it. Thank you kindly.
* All the tests have been rewritten to use the new scheme and the new vector-based combinators.
* An experimental first cut at an alternative vector representation is also included.


cava/cava/Cava/Arrow/Instances/Netlist.v
```diff
@@ -100,7 +100,8 @@ Section NetlistEval.
   Local Open Scope monad_scope.
```
blaxill, line 100: I think the changes in this file are accidental? They seem to be reversions

satnam6502: I've reverted this file to what's in HEAD at the moment.

PR #164, opened 2020-07-06
A two sorter example.

satnam6502:
The leaf component I want to use for a recursive Batcher's bitonic merger/sorter.
The two sorter works in Verilog simulation.

However, I expected the proof of the two sorter's correctness to be easy, but it unfolds in a complicated way.
The next step is to define riffile, but the dependent `Vector.t` types need to be ironed out.


satnam6502:
Yes, I shall leave the proof for later (I'll probably need your help). Thank you!

PR #165, opened 2020-07-07
Add missing Compare_dec dependency

satnam6502:


blaxill:
I think the (unrelated) errors should be fixed now with #166

PR #167, opened 2020-07-07
Enable indexing and slicing on vectors containing arbitrary arrow kinds

blaxill:
Previously arrow netlist indexing and slicing would only work for vectors containing bits, this fixes the other cases allowing indexing and slicing on vectors containing any kind.


blaxill:
I'm an aggressive proponent of aggressive `_` usage :D  The usage here though is mostly so I can delay the expression definitions to the following refine block and add them as Ltac. I'll likely revert to gallina definitions in a future cleanup, but for quick experimentation I often find Ltac quicker.

PR #168, opened 2020-07-08
Recover smashed arrays in Haskell back-end

satnam6502:
This improves the quality of the generated SystemVerilog for many uses of dynamic and static indexing.
There is still an explosion in the unsmshing and de-literalization process that makes the two sorter example not work out well (but still generated valid SystemVerilog that is synthesis equivalent to a good implementation). Slicing also still needs to be supported. These two points will be addressed in a subsequent PR.


satnam6502:
Also, for the moment, I produce a debug comment in the generated SystemVerilog code to help with debugging. Eventually this will be removed, but for now helps to pinpoint at which stage of the compilation process we experience blow-ups.

PR #172, opened 2020-07-09
Close Arrow expressions so that variables are out of scope for splices

blaxill:
- Also add method for delimiting modules, currently unused.

### DIFF
--- a/cava/arrow-examples/Mux2_1.v
+++ b/cava/arrow-examples/Mux2_1.v
@@ -15,7 +15,7 @@
 (****************************************************************************)
 
 From Arrow Require Import Category ClosureConversion.
-From Cava Require Import Arrow.ArrowExport.
+From Cava Require Import Arrow.ArrowExport Arrow.CavaArrow.
 
 Require Import Coq.Strings.String.
 From Coq Require Import Lists.List.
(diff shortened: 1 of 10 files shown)


cava/cava/Cava/Arrow/CombinationalArrow.v
```diff
@@ -128,46 +130,147 @@ Instance CombinationalSTKC : ArrowSTKC CoqKindMaybeArrow := { }.
       | right Hneq => (ltac:(exfalso;lia))
       end;
 
-  slice n x y o H v := _;
```
satnam6502, line 131: So this is different from the slice in `VectorUtils`?

cava/arrow-examples/Mux2_1.v
```diff
@@ -25,12 +25,13 @@ Local Open Scope string_scope.
...
+  Local Open Scope kind_scope.
 
-  Section var.
-    Variable var: Kind -> Kind -> Type.
+  (* Section var.
+    Variable var: Kind -> Kind -> Type. *)
 
     Definition mux2_1
-    : CavaExpr var << Bit, << Bit, Bit >>, Unit >> Bit :=
+    : forall cava: Cava, << Bit, << Bit, Bit >>, Unit >> ~> Bit :=
```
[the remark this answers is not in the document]
blaxill: I've made all the expression reduce to the arrows (instead of doing it later explicitly with a `to_arrow` function). So this arrow notation making a come back (`~>`). Unfortunately you have a good point that its hard to visually parse with the angled brackets being used for generalized tupling. `~>` could be written explicitly as `morphism`.. I'll have a think about how to make the notation clearer here.

cava/cava/Cava/Arrow/CombinationalArrow.v
blaxill: It has different input hypothesis, the definition has been in arrows for a while, but effectively it is code duplication. I plan to rethink how proofs for indexing are done (in arrows & arrow expressions) and Ill address this at that time (I'll change this code to rely on vectorutils version at some point and probably change to the same proof requirement)

PR #173, opened 2020-07-13
Add evaluator for Arrows

blaxill:
This evaluator is modelled as a transition function from 'input -> state -> (output, state)'.
It should be capable of handling all loop/delay/etc elements. The
semantics are currently that looping elements are evaluated once- i.e.
prevention of combinational loops is intended to be handled by external
lemmas. In the future this evaluator could be extended to handle combinational
loops that can be proved to converge after 'n' loops within the same cycle, if
this would be a useful property.

The intention is to remove the 'CombinationalArrow' in a future PR, with 'stateless_evaluation' from `EvaluationArrow` added here to replace evaluation from `CombinationalArrow`.

### DIFF
--- a/cava/cava/Cava/Arrow/EvaluationArrow.v
+++ b/cava/cava/Cava/Arrow/EvaluationArrow.v
@@ -1,0 +1,288 @@
+(****************************************************************************)
+(* Copyright 2020 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
+(* you may not use this file except in compliance with the License.         *)
+(* You may obtain a copy of the License at                                  *)
+(*                                                                          *)
+(*     http://www.apache.org/licenses/LICENSE-2.0                           *)
+(*                                                                          *)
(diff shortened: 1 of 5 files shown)


cava/cava/Cava/Arrow/EvaluationArrow.v
```diff
@@ -0,0 +1,288 @@
...
+  | Unit => unit
+  end.
+
+Fixpoint kind_default (ty: Kind): denote ty :=
+  match ty return denote ty with 
+  | Tuple l r => (kind_default l, kind_default r)
+  | Bit => false
+  | Vector ty n => const (kind_default ty) n
+  | Unit => tt
+  end.
```
satnam6502, line 40: In the monad version I've experimented with using a `Default` typeclass. However, this looks good too.

satnam6502:
I think we should consider circuits with combinational loops as out of scope for Cava (i.e. they are illegal circuits). Although we have experimented with combinational and sequential interpretations, I suspect ultimately we just want a sequential interpretation.

PR #174, opened 2020-07-14
Refactor reusable Arrow circuits and combinators into Library.v

blaxill:


blaxill:
> I worry that `Library` is too generic a name. Break down further, into things like `ArrowCombinators` etc?

I've renamed the file in a follow up PR

PR #175, opened 2020-07-15
Add expression-arrow automated proof of stateless circuits

blaxill:
This adds an automated process for proving a circuit built from expressions containing no loops and only stateless subexpressions is stateless. 

This also renames the `Library` to `Combinators`- I didn't use `ArrowCombinators` as the file is under `arrow-examples/` rather than the main `cava` directory.

### DIFF
--- a/cava/arrow-examples/ArrowExtraction.v
+++ b/cava/arrow-examples/ArrowExtraction.v
@@ -22,14 +22,14 @@ From Coq Require Import extraction.ExtrHaskellNatInteger.
 
 Extraction Language Haskell.
 
-Require Import Library.
+Require Import Combinators.
 Require Import SyntaxExamples.
 Require Import Mux2_1.
 Require Import UnsignedAdder.
 
 Require Import ArrowAdderTutorial.
 
-Extraction Library Library.
+Extraction Library Combinators.
 Extraction Library SyntaxExamples.
 Extraction Library Mux2_1.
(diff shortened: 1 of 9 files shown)


blaxill:
> Is is this work necessary for generating the SystemVerilog for the AES circuits and doing the proofs?
> If it is necessary then I'd like to understand more why we need to prove a circuit is a well formed sequential circuit? Surely there are many other easier options:
> 
> * Make it impossible to construct a circuit that contains loops without delays (as is the case with the monad version). You can do this by exposing a loop combinator that has a delay built into the "input" or "output".
> * Before netlist generation reject any circuits with loops that contain no delay.
> 
> I ask because I am keen to focus our energy on circuit implementation and proof for the AES circuits and I am not sure if automated proofs of stateless circuits are adding much?
> 
> Of course, perhaps I am missing something and this is really needed!
> 
> Thank you kindly.

First let me say you're correct that I should maintain focus on AES, I'll do that.

Regarding what this does- it provides a proof that certain *classes* of expressions are combinational. I was mostly motivated not by the combinational proof aspect but exploring the proof over a class of expressions. e.g. With both the monad and arrow versions its possible to prove a property about a particular expression, but I think this approach should make connecting proofs on different instances (netlist/evaluation) more tractable, and I think this is key to automated proofs on circuit in general (or at least the proof by reflection approach behind it is).

PR #176, opened 2020-07-15
Update existing AES code for recent arrow changes

blaxill:
- Update existing AES code for recent arrow changes
- Add some small evaluation examples (for regression/value correctness testing) in `sbox_canright_pkg.v ` 
- Notation: Change slicing from `v [ x : y ]"` to `v [: x : y ]"` to make syntax grammar easier
- Notation: Remove `vector {a, b, c}` in favor of just `a :: b :: c :: []`  (at least for now)


cava/arrow-examples/Aes/sbox_canright.v
```diff
@@ -0,0 +1,106 @@
...
+      let c = !aes_square_scale_gf2p4_gf2p2 a in
+      let d = !aes_inverse_gf2p4 (c ^ b) in
+
+      concat 
+        (!aes_mul_gf2p4 d (gamma[:7:4]))
+        (!aes_mul_gf2p4 d (gamma[:3:0]))
+  ]>.
+Next Obligation. lia. Qed.
+Next Obligation. lia. Qed.
+Next Obligation. lia. Qed.
```
satnam6502, line 77: Fine for now, but I hope at a later stage we can come up with descriptions that are easier to comprehend for humans (i.e. without the `Next Obligation` lines.

blaxill: Yeah there are some ways this could be addressed. The most straight forward way would be to have a clearer fallibly lowering frontend pass. That would allow all the indexing constraints to be automatically generated. However, this is currently not very high on my priorities.

PR #177, opened 2020-07-16
AES logic: Finish sub_bytes and add shift_rows

blaxill:
- Also simplify arrow split definition


blaxill:
Merging to work on top

PR #179, opened 2020-07-17
Add remaining missing AES primitives.

blaxill:
- Also adds a `unsigned_sub` cava arrow, which is currently left unimplemented for netlist interpretation.
- Key scheduling and a fully unrolled cipher still to come- key scheduling relies on primitives already implemented.
- Most of the AES code remains unchecked for numerical mistakes. Once the last part (key scheduling) is done, I'll go over everything for such simple mistakes (and then the next step would be proofs!).


cava/arrow-examples/Combinators.v
```diff
@@ -269,13 +280,44 @@ Definition enable_vec {n}
...
   : forall cava: Cava, << Bit, Vector Bit n, Vector Bit n, Unit >> ~> <<Vector Bit n>> :=
   <[\ switch xs ys =>
       let not_switch = not switch
       in !(map2 <[\x y => or x y ]>) (!enable_vec switch xs) (!enable_vec not_switch ys)
   ]>.
 
+Definition mux_item {T}
+  : forall cava: Cava, << Bit, T, T, Unit >> ~> <<T>> :=
+  <[\ switch xs ys =>
+      let not_switch = not switch
```
satnam6502, line 317: Is it worth defining `not_switch` if it is only used once below.

cava/cava/Cava/Arrow/CombinationalArrow.v
```diff
@@ -97,6 +97,12 @@ Instance CombinationalSTKC : ArrowSTKC CoqKindMaybeArrow := { }.
     let c := (a + b)%N in
     Some (Ndigits.N2Bv_sized s c);
 
+  unsigned_sub s '(av, bv) :=
+    let a := Ndigits.Bv2N av in
+    let b := Ndigits.Bv2N bv in
+    let c := (a - b)%N in
```
satnam6502, line 103: I wonder if the semantics of unsigned subtract for `N` in Coq match unsigned subtraction semantics for SystemVerilog?

cava/arrow-examples/Combinators.v
blaxill: Removed

cava/cava/Cava/Arrow/CombinationalArrow.v
blaxill: You're right this is almost certainly incorrect on underflow and maybe other conditions, I've added a `todo` inline.

PR #183, opened 2020-07-31
Update makefiles and add third_party submodules

blaxill:


blaxill:
I think it should be possible to unlock more parallelism during the build steps, but there are some unexpected (to me) failures when removing more 'waitFor' constraints in the `cloudbuild.yaml` file.

PR #184, opened 2020-07-31
Add non-recursive let and refactor Arrow code

blaxill:
In preparation of removing redundant CavaExpr expression in favour of arrow-lib Kappa, which CavaExpr currently immediately desugars to.

### DIFF
--- a/cava/arrow-examples/Aes/sbox.v
+++ b/cava/arrow-examples/Aes/sbox.v
@@ -91,4 +91,5 @@ Section regression_testing.
   Goal aes_sbox_lut Combinational (false, #127) = aes_sbox_canright Combinational (false, #127).
     vm_compute; auto.
   Qed.
+
 End regression_testing.
--- a/cava/arrow-examples/Combinators.v
+++ b/cava/arrow-examples/Combinators.v
@@ -439,7 +439,7 @@ Section regression_tests.
     (carry, result)
     ]>.
  
-  Lemma interleave_is_stateless : forall n, has_no_state (interleaveVectors n EvalCava).
+  (* Lemma interleave_is_stateless : forall n, has_no_state (interleaveVectors n EvalCava).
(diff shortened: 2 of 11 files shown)


cava/arrow-lib/Debruijn.v
```diff
@@ -0,0 +1,192 @@
...
+  match argts with
+  | nil => unit
+  | cons T argts' => T * hlist argts'
+  end.
+
+Import ListNotations.
+Import EqNotations.
+
+Set Implicit Arguments.
+Set Asymmetric Patterns. *)
```
satnam6502, line 21: Commented out?

cava/arrow-lib/Debruijn.v
```diff
@@ -0,0 +1,192 @@
...
+    | DComp _ _ _ _ e1 e2 => fun vars => Comp (debruijn_to_phoas e1 vars) (debruijn_to_phoas e2 vars)
+    | DMorph _ _ _ m => fun vars => Morph _ m
+    | DLetRec _ _ _ _ v f =>
+      fun vars => LetRec 
+        (fun x => debruijn_to_phoas v (x, vars))
+        (fun x => debruijn_to_phoas f (x, vars))
+    end.
+
+(* 
+  (* todo: I would like to have a lemma such as the following rather than use Kappa_equivalence axiom in Equiv.v *)
```
satnam6502, line 182: Perhaps we can try to formalize the TODOs so we can use them with search or tooling? For example:
```
TODO(benblaxill): ....
```

cava/arrow-lib/Debruijn.v
blaxill: I've removed this file for now, it should allow a path to removing the kappa_equiv axiom in the Kappa.v file, but some nasty dependently typed proofs are required that were taking me too long.

PR #185, opened 2020-08-03
Allow heterogenous collection of signals in a component instantiation

satnam6502:


satnam6502:
Self-merging to allow dependent changes overnight.

PR #190, opened 2020-08-12
Performance and Coq representations

blaxill:


blaxill:
Currently this is "too slow" in both Coq evaluation and evaluating the extracted Haskell. A possible immediate issue in the extracted haskell, is that `ArrowKind.Kind`s are explicit everywhere, and often include the `Vector` constructor which takes a `nat`, e.g.:

~~~Haskell
cipher_round :: Pkg.SboxImpl -> CavaArrow.Cava -> Category.Coq_morphism
                ArrowKind.Kind
cipher_round sbox_impl cava =
  ClosureConversion.coq_Closure_conversion ArrowKind.Unit
    (CavaArrow.cava_cat cava) (\x x0 -> ArrowKind.Tuple x x0)
    (CavaArrow.cava_arrow cava) (CavaArrow.cava_arrow_stkc cava)
    (CavaArrow.cava_arrow_loop cava) ArrowKind.kind_decidable_equality_inst
    (Arrow.product (CavaArrow.cava_cat cava) ArrowKind.Unit (\x x0 ->
      ArrowKind.Tuple x x0) (CavaArrow.cava_arrow cava) ArrowKind.Bit
      (Arrow.product (CavaArrow.cava_cat cava) ArrowKind.Unit (\x x0 ->
        ArrowKind.Tuple x x0) (CavaArrow.cava_arrow cava) (ArrowKind.Vector
        (ArrowKind.Vector (ArrowKind.Vector ArrowKind.Bit (Prelude.succ
        (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ
... many more lines
~~~

This is exacerbated in places using "flat" `Vector`s with a large dimension, such as taking a `Vector Bit 256`.

The immediate reason for this representation is that the expressions are defined directly as Arrow.kappa expressions (and functions processing it) and have both the input and output type of an expression as type indexes (the `object` in the following`):

~~~Coq
    Inductive kappa : object -> object -> Type :=
~~~

Now, I'm not sure this is the primary reason for the Haskell evaluation being slow, but it seems a good place to start. (Note: extraction of `sbox_lut` is disabled as the expression is so large due to the reason above that it freezes `ghci` when trying to load the file.)

@jadephilipoom any suggestions or advice on what path to take would be greatly appreciated, but please don't feel any pressure to do so.

Currently my thoughts on possible ways forward: 

- Make a frontend untyped language, and change the notation to use it. Pros: expressions won't contain the excessive type information, so at least the extracted Haskell should be a reasonable size (?) Cons: I think this would lose the automatic type inference that the current notation has.
-  Try to reduce/partially evaluate expressions in Coq before extraction forced via Ltac evaluation. This requires the evaluation within Coq to be performant, which it is currently not for larger expressions. 
-  Play more with extraction dials: Perhaps usage of `Extraction Implicit` can reduce the amount of `Kind` terms.

Perhaps there is a better way of approaching this? The goal being to be able evaluate expressions (whether in Coq or Haskell) in a reasonable time to allow value-based debugging of the unrolled cipher.

blaxill:
After following some suggestions from @satnam6502 to debug from the Haskell side, compiling with `ghc -O2` is quite a bit faster than I perhaps thought:

~~~
time ./Main                                   
./Main  9.81s user 1.03s system 95% cpu 11.334 total
~~~

10 seconds for evaluation of full AES-256 14 rounds, not fast but fast enough for me to continue finishing the cipher. I'm still curious if other/better Coq representations would help.

jadephilipoom:
> @jadephilipoom any suggestions or advice on what path to take would be greatly appreciated, but please don't feel any pressure to do so.

My first thoughts when something is unexpectedly very slow to evaluate, in order, would be:
1. There are proofs somewhere in the term you're evaluating. Coq makes it easy to put proofs into types and mix proofs and code, but these proof terms are often huge and evaluating/extracting them is very slow.
2. The term itself, when evaluated, is exponentially sized. Have you evaluated smaller, analogous terms? Could the size of the result scale exponentially?
3. There's a huge `nat` somewhere, or another ~unary datatype that wasn't designed for evaluation at scale. 256 should be okay for `nat`, though.

jadephilipoom:
I checked out the code and looked into it a bit, and it seems like problem 1) is at least contributing. I found `Kappa_wf`in the term under a couple of layers of definition, which seems to be a major theorem.

I tried to find some smaller examples that are slow, and it turns out just `Compute Combinational` also hangs. It looks like `slice` is the culprit there (figured this out by trying e.g. `Compute (@CavaArrow.snoc Combinational)`, `Compute (@CavaArrow.splice Combinational` until I found what was slow), but `snoc`, `uncons`, etc also all contain proof terms that *just happen* to be small enough to compute; a change to the proof or to the underlying tactics like `lia` could easily make them large too.

Mixing proofs into types and definitions is fine and convenient for theoretical-math type Coq proofs, but for program verification it's usually not a good idea. The best way is to have an expression with no proofs, and just prove things about it separately (for instance, instead of having a datatype that tracks a property in the type, have a simpler datatype that doesn't track that property and then separately prove that your functions operating on that datatype guarantee/preserve the property). If there's a very strong justification for putting a proof in the datatype, it should be stripped out before you evaluate. For instance, you could use a sigmatype (`sig`), which pairs a term and a proof about that term, and then make sure that you have extracted the non-proof part with `proj1_sig` before evaluating.

blaxill:
Closing in favour of #193

PR #193, opened 2020-08-14
WIP Proof separation

blaxill:


blaxill:
@jadephilipoom thanks for the pointers. I have (perhaps naive) follow up questions, please address at your leisure.

- I've been using decidable equality routines, and have noticed they seem to have non trivial code expansion. When I first started, I was doing the rewrites with lemmas directly e.g. `rew my_eq_proof in _`, of course as I was defining opaque lemmas these were not computable. The matches on decidable equality at least are computable, but perhaps there is a better way to do the type rewriting? (Or perhaps I should structure things to not require this at all?)

- In `CombinationalArrow.v` I've removed the inline proof and also argument proof and replaced it with default values, e.g.

   https://github.com/project-oak/oak-hardware/blob/37f9be9f91057778b7150d0790ad82ec8f78cd4c/cava/cava/Cava/Arrow/CombinationalArrow.v#L121-L131

   Since these decidable equalities are trivially true, I could just match on implies-false proofs. Is this a bad idea for computation? e.g.

~~~Coq
Lemma neq_n1_sn_false: forall n, n+1<>S n -> False. 

Definition ... := 
    match Nat.eq_dec (n + 1) (S n)  with 
      | left Heq => ...
      | right Hneq => match neq_n1_sn_false Hneq with end (* is this ok, or default value is better *)
~~~

- `Kappa_wf`, this one I'm least clear on how to address. `closure_conversion` [(relevant location link)](https://github.com/project-oak/oak-hardware/blob/37f9be9f91057778b7150d0790ad82ec8f78cd4c/cava/arrow-lib/ClosureConversion.v#L329) currently requires a proof term and `Closure_conversion` provides it automatically via `Kappa_wf`.
   * Should I rewrite `closure_conversion` to not require a proof and use default values similarly to the `Combinational` changes?
   * Instead of using `Closure_conversion`/`Kappa_wf` we could have `wf_phoas_context` proofs next to definitions, and pass them to `closure_conversion`- I'm unsure if this would make a difference to the representations in computing/extraction..? :)
   * Does the opaqueness of `Kappa_wf` help at all here?

- Unrelated to the proof separation topic but; I've noticed that e.g. `Compute aes_mul_gf2p2` produces more vector/Fin related code than perhaps i'd expect. I've read at least some community posts of recommending against the usage of standard library vectors, do you have a different vector representation you would recommend? Or perhaps you think we should just use lists everywhere and extra propositions as necessary?

jadephilipoom:
>  * I've been using decidable equality routines, and have noticed they seem to have non trivial code expansion. When I first started, I was doing the rewrites with lemmas directly e.g. `rew my_eq_proof in _`, of course as I was defining opaque lemmas these were not computable. The matches on decidable equality at least are computable, but perhaps there is a better way to do the type rewriting? (Or perhaps I should structure things to not require this at all?)

For this purpose, you probably want to be using the boolean versions of these deciders (`Nat.eqb` instead of `Nat.eq_dec`), like this:
```coq
if (n-y =? (x - y + 1) + (n - x - 1))
then fst (Vector.splitat (x-y+1) (rew [fun x => Vector.t (denote_kind o) x] Heq in v)) 
else kind_default _
```
There are standard library proofs (`Nat.eqb_eq` and `Nat.eqb_neq`) that state that the boolean decider will correspond to logical equality, so you can use these when you get to proofs. There are similar boolean definitions for less-than (`Nat.ltb`), less-than-or-equal (`Nat.leb`), etc. For custom types, you should define your own fully computable boolean equality decider, and then prove the correspondence to logical equality separately. It feels like more work to do the two-step process of computing the boolean version and then calling in the proofs, but it's worth it to prevent issues with computation! See:
```coq
Definition foo :=
  fun x => match Nat.eq_dec 3 x with
                | left Heq => 0
                | right _ => 1
           end.
Compute foo. (* huge 109-line term with lots of proof matches *)
Definition bar :=
  fun x => if 3 =? x then 0 else 1.
Compute bar.
(* = fun x : nat =>
     if match x with
        | 3 => true
        | _ => false
        end then 0 else 1
   : nat -> nat *)
```

> Since these decidable equalities are trivially true, I could just match on implies-false proofs. Is this a bad idea for computation?

Default values are generally better than matching on proofs for computation. You can just use preconditions in your proof that preclude the default-value cases.

>    * Should I rewrite `closure_conversion` to not require a proof and use default values similarly to the `Combinational` changes?

That's what I would do! All your proofs can have the precondition that the expression is well-formed, and then the default-value cases won't show up.

>    * Instead of using `Closure_conversion`/`Kappa_wf` we could have `wf_phoas_context` proofs next to definitions, and pass them to `closure_conversion`- I'm unsure if this would make a difference to the representations in computing/extraction..? :)

As a general rule of thumb, what makes a difference is whether the proof terms are still there when you compute. If the change is something that lets you strip out the proof before computing, it will help. Otherwise, as long as the proof term is somewhere in there, Coq will try to compute it.

>    * Does the opaqueness of `Kappa_wf` help at all here?

Not all computations respect opaqueness; it depends on what exactly you're calling.
```coq
Definition x := 5.
Opaque x.
Eval cbv in x. (* x *)
Eval compute in x. (* x *)
Eval vm_compute in x. (* 5 *)
Eval native_compute in x. (* 5 *)
Compute x. (* 5 *)
```
Versions that don't respect opaqueness will _still_ try to compute your proof (but these are also the fastest computing tactics, so they're probably the ones you want to call). And versions that don't respect opaqueness won't be able to simplify your proof when they have the information. For instance, using the `foo` example above that contains `Nat.eq_dec`, suppose I make `Nat.eq_dec` opaque (it's not by default).
```coq
Eval cbv in (foo 4) (* 1 *)
Opaque Nat.eq_dec.
Eval cbv in (foo 4).  (* if Nat.eq_dec 3 4 then 0 else 1 *)
```
Even though the arguments are fully computed, `cbv` can't tell that `Nat.eq_dec` will return the not-equal branch because it's opaque.

>  * Unrelated to the proof separation topic but; I've noticed that e.g. `Compute aes_mul_gf2p2` produces more vector/Fin related code than perhaps i'd expect. I've read at least some community posts of recommending against the usage of standard library vectors, do you have a different vector representation you would recommend? Or perhaps you think we should just use lists everywhere and extra propositions as necessary?

There are a lot of opinions circulating on the correct representation of bitvectors in Coq, and no consensus on the correct answer. My personal opinion is that I do have a preference against the standard library vectors. My research group at MIT had a couple of different [homegrown](https://github.com/mit-plv/coqutil/blob/c8006ceca816076b117c31d7feaefb5bbb850754/src/coqutil/Word/Interface.v#L7) [approaches](https://github.com/mit-plv/bbv) that I've used. There's also, of course, always the `list bool` option, or the standard library datatypes that are bit vectors under the hood (e.g. `positive`, `Z`). Since we'll want various different sizes of bitvector, I'd lean towards either `list bool` with length preconditions in proofs and default values when the list terminates prematurely, or `positive` with `Pos.size` preconditions in proofs, handled similarly. An advantage of using `positive` would be that you could, for instance, prove addition equivalent to `Pos.add`. It's probably worth running some tests with extraction to see whether `positive` and `list bool` extract nicely to Haskell, though, and which of them comes out saner.

jadephilipoom:
> Since we'll want various different sizes of bitvector, I'd lean towards either `list bool` with length preconditions in proofs and default values when the list terminates prematurely, or `positive` with `Pos.size` preconditions in proofs, handled similarly. An advantage of using `positive` would be that you could, for instance, prove addition equivalent to `Pos.add`.

Thinking about this a little more, I think there's some major points in favor of `list bool` over `positive`. First, `list bool` will give us two constructors (either there's a next bit or there's not) and `positive` will give us 3 (either we've reached the highest 1 bit, and there's no more, or there's a 0 and more bits beyond, or there's a 1 and more bits beyond). I think the two-constructor version is more natural for circuits because it lets us pass along the bit without inspecting its value; with `positive` we'd probably end up needing to repeat code twice all over the place. And second, `positive` locks us into a single interpretation of endianness (confusingly, big-endian in the notation and little-endian in construction), where we might want to make a different choice.

cava/arrow-lib/ClosureConversion.v
```diff
@@ -138,19 +141,17 @@ Definition rewrite_object {x y: object} (H: x = y)
...
+    if x =? (length ctxt') then
       match eq_dec ty' ty with
       | left Heq2 => rew Heq2 in (second drop >>> cancelr) 
-      | right Hneq => match lookup_top_contra ctxt' Heq H Hneq with end
+      | right Hneq => drop >>> default_object _
       end
-    | right Hneq => first drop >>> cancell >>> extract_nth ctxt' x (push_lookup _ _ Hneq H)
-    end
+    else first drop >>> cancell >>> extract_nth ctxt' _ x
   end.
```
blaxill, line 155: Ive removed the proof passing and some proof matches, but I'm not sure if I can remove these last decidable proof matches- I need the proof for the rewrite. This seems to be the trade off for using dependent types?

blaxill:
This doesn't do everything mentioned in this thread but is enough to be useful. Does it look ready to merge?

jadephilipoom (approved):
Yes, LGTM!

PR #194, opened 2020-08-14
Naive key expansion

blaxill:
Key expansion not based on the open titan routines

### DIFF
--- a/cava/cava/Cava/Arrow/CavaNotation.v
+++ b/cava/cava/Cava/Arrow/CavaNotation.v
@@ -111,8 +111,8 @@ Module KappaNotation.
   Notation "x ++ y" := (App (App (Morph (Primitive (Concat _ _))) x) y) (in custom expr at level 4) : kappa_scope.
   Notation "x :: y" := (App (App (Morph (Primitive (Cons _ _))) x) y) (in custom expr at level 7, right associativity) : kappa_scope.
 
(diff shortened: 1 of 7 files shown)


cava/cava/Cava/Arrow/CavaNotation.v
```diff
@@ -111,8 +111,8 @@ Module KappaNotation.
   Notation "x ++ y" := (App (App (Morph (Primitive (Concat _ _))) x) y) (in custom expr at level 4) : kappa_scope.
   Notation "x :: y" := (App (App (Morph (Primitive (Cons _ _))) x) y) (in custom expr at level 7, right associativity) : kappa_scope.
 
-  Notation "'true'" := (Morph (Primitive (Constant true))) (in custom expr at level 2) : kappa_scope.
-  Notation "'false'" := (Morph (Primitive (Constant false))) (in custom expr at level 2) : kappa_scope.
+  Notation "'true''" := (Morph (Primitive (Constant true))) (in custom expr at level 2) : kappa_scope.
```
satnam6502, line 114: What's the reason for the tick?

blaxill: In a section that imports the notation, the Coq parser will no longer parse `true`/`false` in scopes that are not in kappa_scope (and also for following sections). Often this doesn't matter (even here the part that required it was refactored out), and there might be a simple fix that allows both, but I've bumped into this a few times so I've changed it for now.

PR #196, opened 2020-08-17
Add OpenTitan at commit 783edaf444eb0d9eaf9df71c785089bffcda574e

blaxill:
Adds OpenTitan to `third_party/opentitan` as a submodule, the opentitan directory is currently not used in any build scripts.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "third_party/coq-ext-lib"]
 	path = third_party/coq-ext-lib
 	url = https://github.com/coq-community/coq-ext-lib
+[submodule "third_party/opentitan"]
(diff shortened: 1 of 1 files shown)


satnam6502:
Where is the checkout set to 783edaf444eb0d9eaf9df71c785089bffcda574e ?
I think it might be better to just check out OpenTitan at HEAD and then have build scripts that check out the required version of OpenTitan?

blaxill:
> Where is the checkout set to 783edaf444eb0d9eaf9df71c785089bffcda574e ?

The commit version is stored in the hidden `.git` folder, the Github interface shows this by showing the diff for `third_party/opentitan` as 

> Submodule opentitan added at 783eda

PR #203, opened 2020-09-02
Coq style guide

jadephilipoom:
First pass at a Coq style guide! I just wrote down the invisible code-style rules I use for myself (that I could remember off the top of my head). Happy to agree on different standards if anything seems tedious or counter-intuitive.


cava/CoqStyleGuide.md
```diff
@@ -0,0 +1,242 @@
...
+     1 2)
+  (Z.pow 1 2 3
+         4 5 6)
+  ```
+- `Inductive` cases should not be indented. Example:
+  ```coq
+  Inductive Foo : Type :=
+  | FooA : Foo
+  | FooB : Foo
+  .
```
satnam6502, line 80: Should the `.` always be on a line by itself at the end? Likewise for a type with just one constructor?

cava/CoqStyleGuide.md
```diff
@@ -0,0 +1,242 @@
...
+           | |- context [x * (?a * ?b)] =>
+             replace (x * (a * b)) with (a * b * x) by lia
+           | _ => reflexivity
+           end.
+  Qed.
+End Z.
+(* Now we can access the lemmas above as Z.pow_3_r and Z.pow_4_r, as if they
+   were in the ZArith library! *)
+
+Definition bar (x y : Z) := x ^ (y + 1).
```
satnam6502, line 204: Any style guidelines on when to give the result type explicitly? I see it is implicit here.

satnam6502 (approved):
Looks great! Thanks! Keen to get comments from @blaxill too.

cava/CoqStyleGuide.md
jadephilipoom: I always put it on a line by itself at the end, because that makes it easy to add/remove constructors. I was debating putting it in, but it seemed maybe too picky for a style guide. What do you think?

cava/CoqStyleGuide.md
jadephilipoom: I'm personally all over the place with when I give the result type explicitly, except that I generally do it less often for simple types and more often for complex types. But maybe we should come up with a norm for this. I wouldn't be opposed to declaring that we always write the type explicitly.

satnam6502: I agree that it is best to just use judgement here and omit the type in simple cases, but provide it in other cases where the return type is not totally obvious.

jadephilipoom: Sounds good. I'll add a bullet point somewhere for that.

cava/CoqStyleGuide.md
```diff
@@ -0,0 +1,254 @@
...
+
+  repeat match goal with
+         | _ => progress subst
+         | _ => reflexivity
+         end.
+
+  do 2 lazymatch goal with
+       | |- context [eq] => idtac
+       end.
+  ```
```
blaxill, line 106: is there an advantage to keeping them aligned with the match/lazymatch vs unindented? e.g.

```
 repeat match goal with
 | _ => progress subst
 | _ => reflexivity
 end.
```

I don't feel strongly either way, but having it aligned to the outer scope might require less formatting?

cava/CoqStyleGuide.md
```diff
@@ -0,0 +1,254 @@
...
+## Proofs and tactics
+
+- Use the `Proof` command (lined up vertically with `Lemma` or `Theorem` it
+  corresponds to) to open a proof, and indent the first line after it 2 spaces.
+- Very small proofs (where `Proof. <tactics> Qed.` is <= 80 characters) can go
+  all in one line.
+- When ending a proof, align the ending statement (`Qed`, `Admitted`, etc.)
+  with `Proof`.
+- Avoid using autogenerated names (e.g. `H0`, `n0`). They make the proof
+  break when new hypotheses are added or variable names change.
```
blaxill, line 135: Does this mean all instances of e.g `intros` should include new identifiers- is it ok to omit when the proof is immediately dispatched e.g. `intros; my_solver`?

cava/CoqStyleGuide.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Naming
+
+- Helper proofs about standard library datatypes should go in a module that is
+  named to match the standard library module (see example).
+  * This makes the helper proofs look like standard-library ones, which is
+    helpful for categorizing them if they're genuinely at the standard-library
+    level of abstraction.
+- Names of modules should start with capital letters.
+- Names of inductives and their constructors should start with capital letters.
```
blaxill, line 175: Going by the example perhaps add:
`- Names of other definitions/lemmas should be snake case`

blaxill (approved):
- It would be useful at some point to have a shell script that just does something like `awk` and checks for style violations 
- To proceed with the style guide I'd suggest we opportunistically make edits and then at some point in the future make a sweep for remaining violations

cava/CoqStyleGuide.md
jadephilipoom: I usually surround colons with spaces, but I don't feel strongly about it either, and am happy to leave it up to individual preference. Maybe for these things we want to leave unspecified, to minimize ambiguity, we should add a point to the guide saying explicitly that either way is okay?

cava/CoqStyleGuide.md
jadephilipoom: Ah, I should clarify this a bit. To me, it doesn't matter whether Coq generates automatic names, it matters whether the proofs explicitly *refer* to them, because then the proofs will break if the automatic names change. So `intros; my_solver` is fine; `intros; apply H; my_solver` is not. Will reword this.

cava/CoqStyleGuide.md
jadephilipoom: The reason I added it is that the emacs + ProofGeneral setup that I use automatically aligns them with the `match`, so I have to constantly manually re-indent if the match is aligned differently. I thought coqide indented it the same way -- out of curiosity, what setup are you using?

If I were to back-solve for a more philosophical explanation, there's kind of an invisible parenthesis around the `match...end`, so it makes some sense to format a `repeat match` similarly to a function application.

blaxill: > emacs + ProofGeneral setup that I use automatically aligns

Thats good enough reason for me! 

> what setup are you using?

I'm using vscode, with as many changes/plugins to make it as close to vim as possible :) (Not using vim directly as I had issues with existing Coq vim plugins)

cava/CoqStyleGuide.md
jadephilipoom: I've changed the wording, mind taking another look at this bullet?

jadephilipoom:
> * It would be useful at some point to have a shell script that just does something like `awk` and checks for style violations
> * To proceed with the style guide I'd suggest we opportunistically make edits and then at some point in the future make a sweep for remaining violations

I agree with the way to proceed for now, and I think a shell script is a great idea. Right after we make the shell script is probably a great time to do a sweep, too.

PR #205, opened 2020-09-11
Naive AES: Fix decryption mode

blaxill:


jadephilipoom (commented):
It looks like this PR changes decryption from the "equivalent inverse cipher" to the "inverse cipher". OpenTitan docs say they implement equivalent inverse, if I understand them correctly. Why the change?

blaxill:
I wasn't trying to follow the OpenTitan version with this implementation, but you're right it makes sense to follow their decryption form, I'll update this PR.

cava/arrow-examples/Aes/naive_unrolled_cipher.v
```diff
@@ -194,9 +193,18 @@ Section tests.
...
+
+  Goal combinational_evaluation (unrolled_cipher_naive SboxCanright) unrolled_cipher_naive_comb
+      (false, (test_data, test_key))
+    = test_encrypted.
+  Proof. time now vm_compute. Qed.
+  Goal combinational_evaluation (unrolled_cipher_naive SboxCanright) unrolled_cipher_naive_comb
+      (true, (test_encrypted, test_key))
+    = test_data.
+  Proof. time now vm_compute. Qed.
 End tests.
```
jadephilipoom, line 210: Style nitpick: file should end with a newline.

jadephilipoom (approved):
LGTM!

PR #207, opened 2020-09-14
Add continuation expression to arrow laws

jadephilipoom:
This is a change that helps arrow laws work for manipulating expressions. Because the compose operator `>>>` is right-associative, Coq interprets `(A >>> B >>> C >>> D)` as `(A >>> (B >>> (C >>> D))`. Therefore, despite initial appearances, `B >>> C` is not a subexpression. In order to change the example to `A >>> E >>> D`, you need a lemma with a continuation argument, e.g. `forall K, (B >>> C >>> K)`, on the left-hand side.

We'll need to prove an alternate version of the lemma for the case when we want to rewrite the very innermost `compose` (`C >>> D` in the example), because that one doesn't have a continuation. But the non-continuation version should be implied by the continuation-included version as long as there's an identity morphism we can insert. Do we?

### DIFF
--- a/cava/arrow-lib/Arrow.v
+++ b/cava/arrow-lib/Arrow.v
@@ -47,17 +47,23 @@ Class ArrowLaws
   (unit: object) (product: object -> object -> object) (A: Arrow object category unit product)
   := {
   (* (un)cance{l,r} and (un)assoc are natural isomorphisms *)
-  first_cancelr   {x y} (f: x ~> y): first f >>> cancelr =M= cancelr >>> f;
-  uncancelr_first {x y} (f: x ~> y): uncancelr >>> first f =M= f >>> uncancelr;
+  first_cancelr   {x y z} (f: x ~> y) (k: _ ~> z):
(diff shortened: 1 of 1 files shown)


blaxill (commented):
> as long as there's an identity morphism we can insert. Do we?

Yeah there is `id` from the Category definition.

I think the continuation version is derivable from the Category laws too (at least it should be!), that is you should be able to use 

https://github.com/project-oak/oak-hardware/blob/0bcebd3bc4e83241e415ba949c94369c1d3c9ac1/cava/arrow-lib/Category.v#L48

Then `first f >>> (cancelr >>> g)`
- by `associativity` is  `(first f >>> cancelr) >>> g` then 
- by `first_cancelr` is `(cancelr >>> f) >>> g`

Do you find the continuation version more straightforward to work with?

jadephilipoom:
I just need the continuation-style lemmas to exist somewhere. If they're derivable from the original versions, there's no need for this change, because the non-continuation ones are simpler to use in the typeclass. I'll close this and try making separate proofs in continuation-style instead.

PR #208, opened 2020-09-14
Derive arrow-law lemmas with continuations

jadephilipoom:
Revised version of #207. As per discussion in that PR, I left the arrow laws unchanged and derived the continuation-style versions as separate lemmas.


blaxill (approved):
Lgtm!

PR #209, opened 2020-09-14
Kappa specialisation/separation

blaxill:
This specialises kappa and moves the closure_conversion routine to the cava directory. It also standardises the arrow files names.


cava/arrow-examples/Aes/naive_unrolled_cipher.v
```diff
@@ -215,6 +218,6 @@ Section tests.
   Goal combinational_evaluation (unrolled_cipher_naive SboxCanright) unrolled_cipher_naive_comb
```
blaxill, line 218: Something about the new representation is causing this to evaluate much slower. The following runs past 3mins before I terminated it without completing, whereas the master branch version is <60s.

~~~Coq
  Goal combinational_evaluation' (closure_conversion (unrolled_cipher_naive SboxCanright))
      (false, (test_data, (test_key,tt)))
    = test_encrypted.
  Proof. time now vm_compute. Qed.
~~~

The main change is that subexpressions are no longer wrapped with an extra constructor Morph- I don't immediately understand why this would negatively impact performance. @jadephilipoom any ideas or hints?

jadephilipoom: Is it possible that this change would cause the subexpressions to be duplicated in lots of places?

blaxill: I don't think so, although I'll follow up by comparing the output with implicit parameters. An example of the change on the expression AST, top is this branch and is x10 slower, bottom is master. Note with this change `Morph sbox_canright.aes_sbox_canright` becomes `sbox_canright.aes_sbox_canright var`.

~~~
is_combinational
  (closure_conversion
     (fun var : Kind -> Kind -> Type =>
      Abs
        (fun op_i : var Unit Bit =>
         Abs
           (fun data_i : var Unit (Vector Bit 8) =>
            Let
              (App (App (sbox_canright.aes_sbox_canright var) (Var op_i))
                 (Var data_i))
              (fun data_o : var Unit (Vector Bit 8) => Var data_o)))))

Tactic call ran for 1.409 secs (1.331u,0.074s) (success)
Tactic call ran for 1.791 secs (1.746u,0.043s) (success)
Tactic call ran for 1.188 secs (1.157u,0.029s) (success)
Tactic call ran for 1.093 secs (1.036u,0.056s) (success)
Tactic call ran for 1.118 secs (1.116u,0.001s) (success)


is_combinational
  (ClosureConversion.closure_conversion
     (fun var : Kind -> Kind -> Type =>
      Abs
        (fun op_i : var u Bit =>
         Abs
           (fun data_i : var u (Vector Bit 8) =>
            Let
              (App (App (Morph sbox_canright.aes_sbox_canright) (Var op_i))
                 (Var data_i))
              (fun data_o : var u (Vector Bit 8) => Var data_o)))))

Tactic call ran for 0.175 secs (0.168u,0.006s) (success)
Tactic call ran for 0.042 secs (0.042u,0.s) (success)
Tactic call ran for 0.024 secs (0.023u,0.s) (success)
Tactic call ran for 0.023 secs (0.023u,0.s) (success)
Tactic call ran for 0.042 secs (0.042u,0.s) (success
~~~

blaxill: Both the unfolded representation and its cbv trace (Set Debug Cbv) are shorter/smaller on this branch, so no clarity there.

blaxill: There are more issues to discover but argh, _as you've already told me_, the matches for vector rewriting have terrible performance, and seems to be compounded by vectors that have lengths that are from previous operations.

https://github.com/project-oak/oak-hardware/blob/d69db5468dc1e32c121b7b11fef4c1f35acf2410/cava/cava/Cava/Arrow/CombinationalArrow.v#L48-L50

I think this means its a good time to move away from Coq.Vector

jadephilipoom: Ah, yeah, that might well be the problem. I'm guessing it's a pretty significant refactor? If so, maybe we should use the next project chat to decide what bit-vector representation we want to go for.

cava/cava/Cava/Arrow/CircuitProp.v
```diff
@@ -57,20 +57,20 @@ Ltac simply_combinational :=
...
-  | [ |- forall p, (?H1 -> ?H2 -> p) -> p ] => 
+  | [ |- forall p, (?H1 -> ?H2 -> p) -> p ] =>
     let x := fresh in (let y := fresh in (
       intros x y; apply y; clear x y
     ))
   | [ |- _ ] => fail "Term wasn't simply combinational"
   end. *)
 
 Lemma is_combinational_first: forall x y z (circuit: x ~> y),
-  is_combinational (first circuit : x**z ~> y**z) =
```
jadephilipoom, line 68: I'm noticing a lot of deleted whitespace -- just curious, is coqtail removing it for you on save or is this manual? It might be worth us making a quick script that gets rid of all the trailing whitespace so these lines don't show up in diffs, even if we're not ready for a full linter yet.

jadephilipoom (approved):
LGTM! The diff stats on this one are very impressive, nice work!

cava/arrow-examples/Aes/naive_unrolled_cipher.v
```diff
@@ -196,16 +193,15 @@ Section tests.
...
-
-  Goal combinational_evaluation (unrolled_cipher_naive SboxCanright) unrolled_cipher_naive_comb
-      (false, (test_data, test_key))
-    = test_encrypted.
-  Proof. time now vm_compute. Qed.
-  Goal combinational_evaluation (unrolled_cipher_naive SboxCanright) unrolled_cipher_naive_comb
-      (true, (test_encrypted, test_key))
-    = test_data.
-  Proof. time now vm_compute. Qed.
+  (* Vector.eqb is extracted under a different namespace, so we redefine
```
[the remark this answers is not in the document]
blaxill: `Vector.eqb` is extracting as `VectorDef.eqb` and almost all other vector definitions extract with `Vector` prefix. I haven't investigated too deeply why but an easy fix for me was to make my own definition.

cava/arrow-examples/Aes/naive_unrolled_cipher.v
```diff
@@ -196,16 +193,15 @@ Section tests.
...
+  * boolean vector equality here *)
+  Definition vector_equality {n} (v1: Vector.t bool n) (v2: Vector.t bool n) :=
+    Vector.fold_left Bool.eqb true (Vector.map2 Bool.eqb v1 v2).
+
+  Definition naive_cipher_test :=
```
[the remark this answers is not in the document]
blaxill: This was previously a `Goal` but changes in this PR caused the computation time to spike to 5mins for `native_compute` and 30mins for `vm_compute`. I am evaluating this in the extracted Haskell (3second evaluation) but only printing the result, not erroring on it. I'm hoping that future changes will significantly speed things up again.

cava/cava/Cava/Arrow/CircuitProp.v
blaxill: This is from me switching back to vim over vscode. My preference is no trailing whitespace, and my vim config currently enforces this. A script would be better if everyone is on board.

cava/arrow-examples/Combinators.v
```diff
@@ -14,33 +14,34 @@
...
+Local Open Scope category_scope.
 Local Open Scope string_scope.
 Local Open Scope kind_scope.
 
 (* *************************** *)
 (* Rewrites *)
 
 Definition rewrite_kind {x y} (H: x = y) 
-  : << x, Unit >> ~> y :=
+  : << x, Unit >> ~[KappaCat]~> y :=
```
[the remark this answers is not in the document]
blaxill: CircuitCat/CircuitArrow can actually use this `Category` notation, but I'm not sure if that is useful... After this PR all definitions will be KappaCat. I think you're right it would be good to clean/clear this up, I'll hold off since we might do other representational changes first

PR #211, opened 2020-09-22
Remove trailing whitespace

jadephilipoom:
As per https://github.com/project-oak/oak-hardware/pull/209#discussion_r492446055

Removes all trailing whitespace from Coq source files! The only new code here is the script that does the magic, `remove-trailing-whitespace.sh`.


blaxill (approved):
Thanks, lgtm! We could make this run on commits/push via githooks, do you think that would be too obnoxious ?

jadephilipoom:
> We could make this run on commits/push via githooks, do you think that would be too obnoxious ?

I think it's a good idea -- it's less obnoxious to catch it early and give someone a commit error message than to do another round trip on their PR because of whitespace.

PR #212, opened 2020-09-22
Add .nia.cache to .gitignore

jadephilipoom:
The autogenerated file `.nia.cache` showed up in my `git status`. This PR prevents it from doing that again!


blaxill (approved):
IIRC `native_compute` also generates files which could be added too

PR #214, opened 2020-09-22
Tree combinator proofs

jadephilipoom:
Here's the tree combinator proofs I walked through in the last meeting. I'll recap here anyway for archival purposes. The main punchlines are the proofs `tree_equiv` and `treeList_equiv`, both in `Combinators.v`, which serve as the proofs of functional correctness for the preexisting tree combinator operations. In order to prove them, I also needed various proofs about vectors and lists which now live in `VectorUtils.v` and the new file `ListUtils.v`. I created a helper tactic for lists that's pretty generic, so I went ahead and made it exported from `ListUtils.v` and added some tests for it to make it more robust.

### New `resize` definition for vectors

Proving the vector tree combinator as-is was tricky because of vector size coercions, so I created a new definition `resize` that, given a vector and a proof that its is equal to a target size, produces a vector with the target size. For example, it can change a vector of size `2+3` to a vector of size `5`. The `resize` definition pretty much just hides a `rew` behind a definition label, but that makes the `rew` much easier to reason about because I could define a bunch of lemmas about `resize` and use them instead of reasoning about `rew` directly.

### The `resize_default` variant

Although I managed to prove the vector tree combinator proof with that method, I later switched to using `resize_default`, a version of `resize` that uses truncation/default values to coerce a vector to the target size no matter whether the size matches or not. There are two advantages of this strategy. The first is that it actually ends up making proof reasoning easier, because mismatches in the proof arguments don't cause problems. The second is performance.

### Performance comparisons

I used the three versions of tree combinators (list-based, vector-based using `resize`, vector-based using `resize_default`) to make some trees of simple `orb` computations, then tested out fully computing them on different sizes of input. Note that the x-axis here is exponential in number of bits.

combinator_performance_comparison

### DIFF
--- a/cava/monad-examples/AdderTree.v
+++ b/cava/monad-examples/AdderTree.v
@@ -39,9 +39,10 @@ Require Import Cava.Monad.UnsignedAdders.
 (******************************************************************************)
 
 Definition adderTree {m bit} `{Cava m bit} {sz: nat}
-                     (n: nat) (v: Vector.t (Vector.t bit sz) (2^(n+1))) :
+                     (default : bit)
+                     (n: nat) (v: Vector.t (Vector.t bit sz) (2^(S n))) :
                      m (Vector.t bit sz) :=
-  tree n addN v.
+  tree (Vector.const default _) n addN v.
 
 (******************************************************************************)
 (* Some tests.                                                                *)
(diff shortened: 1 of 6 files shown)


cava/monad-examples/AdderTree.v
```diff
@@ -39,9 +39,10 @@ Require Import Cava.Monad.UnsignedAdders.
 (******************************************************************************)
 
 Definition adderTree {m bit} `{Cava m bit} {sz: nat}
-                     (n: nat) (v: Vector.t (Vector.t bit sz) (2^(n+1))) :
+                     (default : bit)
+                     (n: nat) (v: Vector.t (Vector.t bit sz) (2^(S n))) :
                      m (Vector.t bit sz) :=
-  tree n addN v.
+  tree (Vector.const default _) n addN v.
```
blaxill, line 45: I think the `default` value should come from the `Cava` monad/be defined by a `Cava` instance, so that plugging a default value is not required for all the top level definitions. (I don't think there is a reason to need different default values per definition)

blaxill (approved):
Lgtm. The plugging of `default` could be improved but its not a blocker

cava/monad-examples/AdderTree.v
blaxill: (I should have highlighted the argument line ` (default : bit)` )

jadephilipoom: This raises the question of whether it's okay to put a plain `bit`, rather than `m bit` as a field in the Cava class. Maybe we could put it in and call it "dummy" or something that makes it very clear it shouldn't appear in specs and we shouldn't rely on any of its properties.

PR #215, opened 2020-09-23
rewrite_or_default items + proof automation

blaxill:
WIP proofs using circuit equivalence.

Also contains duplicated `resize_default` which should be removed (i started using it before the PR was merged)


satnam6502 (approved):
I've looked over some of this but there is a lot going on so right now I am just going to "stamp" this rather than offer a proper code review.

PR #217, opened 2020-09-25
Update to a working `unrolled_opentitan_cipher`

blaxill:
This variant follows the outline of the OpenTitan implementation,
but doesn't have the same stateful core logic. A stateful version is
next which will aim to be a direct translation of the OpenTitan
implementation.


jadephilipoom (approved):
LGTM!

PR #219, opened 2020-09-29
Sequential evaluation for arrows

blaxill:


jadephilipoom (approved):
LGTM. I did look through the code, but it all seems very logical, so I don't actually have any comments!

PR #220, opened 2020-09-29
Revert "Circuit equivalence infrastructure + resize_default (#215)"

blaxill:
This reverts commit 13a52f7376d36db26f668386c66e9dc57dfd843b, so we can fix author and make an additional change.


jadephilipoom:
Just for the GitHub record, the revert is because I handled the co-authorship incorrectly on the original merge, and Ben was listed as the sole author instead of both of us.

jadephilipoom (approved):

PR #222, opened 2020-09-30
Add stateful cipher_control.v

blaxill:
This is a conversion of the OpenTitan `aes_cipher_control.v`. Conversion was done in a somewhat naive manner, and it is likely there is a more concise translation.

I have not checked the evaluation of the outputs yet.


cava/arrow-examples/Aes/cipher_control.v
```diff
@@ -0,0 +1,683 @@
...
+  , Bit
+  (* output logic                    key_expand_clear_o, *)
+  , Bit
+  (* output logic [3:0]              key_expand_round_o, *)
+  , Vector Bit 4
+  (* output aes_pkg::key_words_sel_e key_words_sel_o, *)
+  , key_words_sel_e
+  (* output aes_pkg::round_key_sel_e round_key_sel_o *)
+  , round_key_sel_e
+  >>.
```
jadephilipoom, line 115: We might need to rearrange these single bits into a vector to avoid having a whole bunch of arrow manipulations that change around their order and associativity. I haven't tested this hunch, but I'm guessing more arguments means larger arrow expressions. Probably best to not prematurely optimize, but we should keep this in mind if things are slow later on.

cava/cava/Cava/Arrow/CavaNotation.v
```diff
@@ -42,21 +42,145 @@ Module KappaNotation.
...
-      )
-    ))
+  Notation "'let' x = a 'in' b" := (Let a (fun x => b))
+    (in custom expr at level 1, x constr at level 4, b at level 7, a at level 1) : kappa_scope.
+  Notation "'let' x : ty = a 'in' b" := (Let a (fun x : _ Unit ty => b))
+    (in custom expr at level 1, x constr at level 4, b at level 7, ty constr at level 7, a at level 1) : kappa_scope.
+  Notation "'letrec' x = a 'in' b" := (LetRec (fun x => a) (fun x => b))
+    (in custom expr at level 1, x constr at level 4, b at level 7, a at level 1) : kappa_scope.
+
+  (* TODO(blaxill): can this be turned into a recursive pattern?
```
jadephilipoom, line 52: [This](https://coq.inria.fr/refman/user-extensions/syntax-extensions.html#notations-with-recursive-patterns) is the Coq manual link for recursive patterns in notations, just in case you hadn't found that page yet. But also...

cava/cava/Cava/Arrow/CavaNotation.v
```diff
@@ -42,21 +42,145 @@ Module KappaNotation.
...
+    match ty with
+    | Tuple l r => Fst l r
+    | _ => Fst Unit Unit
+    end.
+
+  Notation "'let' '( x , y ) = a 'in' b" := (
+    Let a (fun a_binder =>
+    Let (App (Primitive (Fst _ _)) (Var a_binder)) (fun x =>
+      Let (App (Primitive (Snd _ _)) (Var a_binder)) (fun y => b))))
+    ( in custom expr at level 1, x constr, y constr, b at level 7) : kappa_scope.
```
jadephilipoom, line 68: ...would something like this work? It'll duplicate `a`, but that's only an issue if `a` is a complicated expression, and I think it's basically always a `Var`, no? (You could try to enforce that in the notation as well, I think.)
```suggestion
  Notation "'let' '( x , y ) = a 'in' b" := (
    Let (App (Primitive (Fst _ _)) a) (fun x =>
      Let (App (Primitive (Snd _ _)) a) (fun y => b))
    ( in custom expr at level 1, x constr, y constr, b at level 7) : kappa_scope.
```

cava/cava/Cava/Arrow/ExprLowering.v
```diff
@@ -341,4 +341,23 @@ Definition closure_conversion {i o} (expr: Kappa i o) : i ~> o
 Hint Resolve closure_conversion' : core.
 Hint Resolve closure_conversion : core.
 
-(* End Arrow. *)
+(* Provides some idea of term sized caused by context *)
```
jadephilipoom, line 344: Seems useful for debugging. Is this a debugging tool, or does it have some other purpose?

jadephilipoom (approved):
LGTM!

I'm a little concerned about performance of some very lengthy Kappa expressions with lots of arguments; we should probably do some experiments to see what exactly results in large Arrow expressions, but I think "large state" and "long expression" are likely contenders. I've currently been working with `unrolled_cipher_naive'`, which is very short compared to e.g. `aes_cipher_control_transition` but has 1230 Arrow constructors when computed.

cava/cava/Cava/Arrow/ExprLowering.v
blaxill: Yeah intended for debugging

cava/cava/Cava/Arrow/CavaNotation.v
blaxill: Unfortunately there is still irregularity (`a_binder` `a_tl_binder`) to larger tuples (see below), which seems to mess with the recursive notation (I run into `Cannot find where the recursive pattern starts`). Perhaps if I abstract the `App ...` to a function and have it indexed by a number it will work...
  
```  Notation "'let' '( x , y , z ) = a 'in' b" := (
    Let a (fun a_binder =>
    Let (App (Primitive (Fst _ _ )) (Var a_binder)) (fun x =>
      Let (App (Primitive (Snd _ _ )) (Var a_binder)) (fun a_tl_binder =>
        Let (App (Primitive (Fst _ _ )) (Var a_tl_binder)) (fun y =>
          Let (App (Primitive (Snd _ _ )) (Var a_tl_binder)) (fun z =>
```

blaxill:
> I'm a little concerned about performance of some very lengthy Kappa expressions with lots of arguments; we should probably do some experiments to see what exactly results in large Arrow expressions, but I think "large state" and "long expression" are likely contenders. I've currently been working with `unrolled_cipher_naive'`, which is very short compared to e.g. `aes_cipher_control_transition` but has 1230 Arrow constructors when computed.

A very valid concern. Just to be clear `unrolled_cipher_naive'` = 1230 constructors? That is more than I would expect, is that after unfolding say `rewrite_or_default`? I wonder if a large number of that comes from `rewrite_or_default` or a similar expression that we could treat independently. 

I think to address it here, I can refactor the state branches to independent functions, which should help drastically in this case.

jadephilipoom:
1230 is without unfolding `rewrite_or_default`. I'm counting every `Circuit` constructor, so that includes `Composition`, `First`, `Second`, `Unassoc`, `Resize`, etc. I didn't unfold any other definitions, including `rewrite_or_default` and subcircuits.

PR #223, opened 2020-09-30
Proofs about circuit_equiv

jadephilipoom:
I've been working on getting an AES top-level proof (top-level circuit proven assuming its subroutines are correct) put together, and in the process I found it useful to:
- change the definition of `circuit_equiv` so it obeys functional extensionality (i.e. if a circuit `c` matches the behavior of a function `s`, and the spec `s'` produces the same result as `s` for all inputs, then `c` matches the behavior of `s'` as well)
- prove that `circuit_equiv` now obeys functional extensionality
- make the `circuit_spec` tactic use `cbn` instead of `unfold`

Then, kind of unrelatedly, I decided to take a stab at proving that `circuit_equiv` between a circuit and a spec (function) implies that the spec is equivalent to `combinational_evaluation'` of the circuit, which is a proof we eventually need for the full chain of reasoning. Turns out it was surprisingly simple and "just worked", so it's here too!

I started writing a proof that all circuits are related by `circuit_equiv` to their evaluation, but there are a couple leftover goals for loops because `circuit_equiv` doesn't have cases for those yet. I left the proof `Abort`ed so we can fill it in later; that property isn't directly needed but might come in handy, and serves as a good sanity check.

CC: @satnam6502 (who probably doesn't have time to review!)

### DIFF
--- a/cava/cava/Cava/Arrow/CircuitFunctionalEquivalence.v
+++ b/cava/cava/Cava/Arrow/CircuitFunctionalEquivalence.v
@@ -14,6 +14,7 @@
 (* limitations under the License.                                           *)
 (****************************************************************************)
 
+From coqutil Require Import Tactics.Tactics.
 From Arrow Require Import Category Arrow.
 From Cava.Arrow Require Import ArrowKind CircuitArrow CircuitSemantics
      ExprSyntax ExprLowering.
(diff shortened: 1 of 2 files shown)


cava/cava/Cava/Arrow/CircuitFunctionalEquivalence.v
```diff
@@ -26,59 +27,107 @@ Inductive circuit_equiv: forall i o, Circuit i o -> (denote_kind i -> denote_kin
...
+  circuit_equiv i o c (combinational_evaluation' c).
+Proof.
+  induction c; cbn [combinational_evaluation']; intros.
+  all:try match goal with
+          | x : ArrowStructure |- _ => destruct x
+          | x : CircuitPrimitive |- _ => destruct x
+          end.
+  all:try econstructor; eauto.
+  all:intros; destruct_products; eauto.
+  (* TODO: loop cases missing *)
```
blaxill, line 128: You could also take an `is_combinational` (defined in CircuitProp.v) that prevents `Loopl` `Looplr `Primitive Delay` from appearing in the circuit.

blaxill (approved):
LGTM!

jadephilipoom:
I took your suggestion, Ben, and used `is_combinational` to get rid of the loops. However, I had to change the definition of `is_combinational` to do it, because the `Map` case didn't recurse.

PR #225, opened 2020-10-01
Proof of list_bits_to_nat_cons

satnam6502:
This PR proves the lemma:

```coq
Lemma list_bits_to_nat_cons b bs :
  list_bits_to_nat (b :: bs) = N.b2n b + 2 * list_bits_to_nat bs.
```
which shows how a bit-vector representation of an unsigned binary number can be decomposed into a value from the LSB plus two times the rest of the bit-vector (i.e. shift right by 1). This is useful in proofs that relate bit-vector level arithmetic operations to arithmetic operations at the level of the `N` type or other types like `nat`.

The original proof was done by @jadephilipoom in b7364dc5923cac09dbc1281a1f6099b5b278cc65 and this version adds commentary and a few adjustments to make it easier (at least for me) to follow.

### DIFF
--- a/cava/cava/Cava/BitArithmetic.v
+++ b/cava/cava/Cava/BitArithmetic.v
@@ -159,6 +159,57 @@ Proof.
 Qed.
 
 (******************************************************************************)
+(* Prove that an unsigned bit-vector b::bs represents the same number         *)
+(* as b + 2 * bs i.e. the low bit represented as a number plus 2 times the    *)
+(* the rest of the bit-vector.                                                *)    
+(******************************************************************************)
+
+Local Open Scope N_scope.
+
+Lemma list_bits_to_nat_cons b bs :
+  list_bits_to_nat (b :: bs) = N.b2n b + 2 * list_bits_to_nat bs.
+Proof.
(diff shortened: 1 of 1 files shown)


cava/cava/Cava/BitArithmetic.v
```diff
@@ -158,6 +158,63 @@ Proof.
...
+(******************************************************************************)
+
+Local Open Scope N_scope.
+
+Lemma list_bits_to_nat_cons b bs :
+  list_bits_to_nat (b :: bs) = N.b2n b + 2 * list_bits_to_nat bs.
+Proof.
+  (* Unfold the definition of list_bits_to_nat to get an expression
+     in terms of BvN.
+  *)
```
jadephilipoom, line 174: Just curious, is there a particular reason for putting the `*)` on the next line down? I would probably have written:
```suggestion
     in terms of BvN. *)
```

jadephilipoom (approved):
LGTM! The proof logic is very clear now.

cava/cava/Cava/BitArithmetic.v
satnam6502: Done!

PR #227, opened 2020-10-05
Bedrock2 implementation of aes_init

jadephilipoom:
I hadn't merged my progress on this so far because it requires adding bedrock2 as a submodule (and making us depend on coqutil through bedrock2 to avoid mismatched coqutil versions), and that seemed like a large global change for a small experiment. But I think now that we've had some promising early results it's probably worth merging before I make any larger changes. What do you two think, @satnam6502 and @blaxill?


.gitmodules
```diff
@@ -1,9 +1,9 @@
-[submodule "third_party/coqutil"]
-	path = third_party/coqutil
-	url = https://github.com/mit-plv/coqutil
```
blaxill, line 3: I guess the preference of the bedrock coqutil version is that it simpler to depend on that version as we don't have to mess with the bedrock submodule?

blaxill (approved):
Lgtm!

.gitmodules
jadephilipoom: Pretty much. We don't want a situation where, for instance, we update our coqutil version and bedrock2 hasn't yet updated theirs. Then we would have two different definitions in scope for everything in coqutil -- bedrock2 would be using one kind of coqutil `map` and we would be using another. And in order to update bedrock2 to match our version of coqutil, we might have to do nontrivial work on bedrock2 to update it, which we probably want to avoid.

satnam6502 (approved):

Issue #50, opened 2020-04-05
Investigate selective Coq to Haskell extraction


blaxill:
Using `Extraction Library` works but we have little control over how the extraction happens, which means the extracted code is not readable (due to lack of type mapping and also module hierachy), and also causes a lot of files to be generated that we need to make sure are hidden from version control (`.gitignore`).

This might be the best way for us, but another way is to manually extract terms with  giving us more control. For an example [Kami](https://github.com/sifive/Kami) uses constant extract to control exactly what and how terms are extracted:

- https://github.com/sifive/Kami/blob/79f75423d7e023f45d126cb8ae2f0b0eb623286f/Extraction.v
- https://github.com/sifive/Kami/blob/8e77e58f368689d18a796c442c4dd20cff603d94/Tutorial/ExtractEx.v
- https://github.com/sifive/Kami/blob/a32841ebf7021f20a455b2a34e9a55322d621d89/Simulator/CoqSim/RegisterFile.v#L331

blaxill:
Extraction is also quite noisy right now, which adds to #305

satnam6502:
Let's punt in this until phase 2 since it is not phase 1 blocking.

PR #228, opened 2020-10-05
Remove v prefix from a number of VectorUtils functions

blaxill:
I think I named these, but now I don't think these need to be prefixed. Where there is ambiguity with list functions, the module name could/should be used to differentiate. Also simplifies the traverse definition.

### DIFF
--- a/cava/cava/Cava/Arrow/CircuitLowering.v
+++ b/cava/cava/Cava/Arrow/CircuitLowering.v
@@ -274,7 +274,7 @@ Fixpoint build_netlist' {i o}
   | Primitive (Index n o) => fun '(v,(i,_)) => index' _ _ v i
   | Primitive (Primitives.Cons n o) => fun '(x, (v,_)) =>
     ret ((x :: v)%vector)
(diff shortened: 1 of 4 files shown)


jadephilipoom (approved):
LGTM modulo a naming nitpick. I have some preference for the module name to match the standard library module name; that way I don't have to remember where a function is defined to call it. This could be solved by either creating a `Vector` module inside `VectorUtils.v`, or renaming `VectorUtils.v` to `Vector.v` (or better yet, `Util/Vector.v`).

blaxill:
> (or better yet, `Util/Vector.v`).

I think this may create an extraction issue- we were doing this kind of thing before but we ended up with Haskell like `Vector.hs` `Vector0.hs` etc. I'll experiment with it and also using a `Vector` module inside the file, before merging.

blaxill:
After investigating a little: Due to our extraction setup, using a `Vector.v` filename generates a `Vector0.hs` haskell filename due to the standard library collision, using a `Module Vector` inside the file makes extracted Haskell function names be prefixed with `_Vector__namehere`. I'm going to merge the PR as is, with the idea of fixing the extraction side of this sometime through more intelligent extraction (see #50), and then at the point unify the names

PR #229, opened 2020-10-05
`closure_conversion` decomposition lemmas

blaxill:


blaxill:
I'm not sure how useful `Var` and other non-branching expression lemmas are, but I added them anyway.

jadephilipoom (approved):
Awesome, thanks! LGTM!

PR #230, opened 2020-10-05
Ignore dirty coq-ext-lib submodule

blaxill:
I think building coq-ext-lib modifies files in its directory causing git to mention it during staging. This just removes the git notification that `coq-ext-lib` directory is dirty.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,7 @@
 [submodule "third_party/coq-ext-lib"]
 	path = third_party/coq-ext-lib
 	url = https://github.com/coq-community/coq-ext-lib
+	ignore = dirty
 [submodule "third_party/opentitan"]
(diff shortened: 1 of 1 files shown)


jadephilipoom:
Hmm, I'd be careful with this one since it could obscure local changes to the dependency. Maybe we can fix this by excluding build output files instead. When you go to the `coq-ext-lib` directory and run `git status`, what's the output?

blaxill:
`.lia.cache` is dirty within `coq-ext-lib`. As its a submodule, our `.gitignore` seems to be ignored (it has a .lia.cache entry). It looks like upstream recent commits add `.lia.cache` to their `.gitignore`. I'll try updating the submodule instead and see if there are any integration issues.

blaxill:
Closing in favor of #231

PR #232, opened 2020-10-06
Combinator equivalence proofs

jadephilipoom:
This PR includes some changes to the `circuit_spec` tactic that use the lemmas in #229 to unfold `closure_conversion'` one constructor at a time, which speeds it up considerably. I then used the sped-up tactic to prove functional equivalence for most of the circuit combinators (I had these proofs lying around already because I needed them for the AES top-level proof I'm working on, but they were due to be separated out and merged and also good test cases!)

### DIFF
--- a/cava/cava/Cava/Arrow/CircuitFunctionalEquivalence.v
+++ b/cava/cava/Cava/Arrow/CircuitFunctionalEquivalence.v
@@ -192,6 +192,8 @@ Ltac lower1 :=
       first [ rewrite (@lower_remove_context x y e ctxt)
             | change x with t1; change y with t2;
               rewrite (@lower_remove_context t1 t2 e ctxt) ]
+    | @Comp _ ?x ?y ?z ?e1 ?e2 =>
+      rewrite (@lower_comp x y z e2 e1 ctxt)
     | @Primitive _ ?p => cbv [closure_conversion']
     | @Var _ _ _ _ => cbv [closure_conversion']
     end; arrowsimpl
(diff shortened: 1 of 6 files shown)


cava/cava/Cava/Arrow/CircuitFunctionalEquivalence.v
```diff
@@ -161,22 +158,77 @@ Ltac arrowsimpl :=
...
+            | change t1 with (Tuple x y); change t2 with z;
+              rewrite (@lower_abs x y z f ctxt) ]
+    | @App _ ?x ?y ?z ?f ?e => rewrite (@lower_app x y z f e ctxt)
+    | @Let _ ?x ?y ?z ?v ?f => rewrite (@lower_let x y z f v ctxt)
+    | @RemoveContext _ ?x ?y ?e =>
+      first [ rewrite (@lower_remove_context x y e ctxt)
+            | change x with t1; change y with t2;
+              rewrite (@lower_remove_context t1 t2 e ctxt) ]
+    | @Primitive _ ?p => cbv [closure_conversion']
+    | @Var _ _ _ _ => cbv [closure_conversion']
```
blaxill, line 196: This is missing a case for `Comp`, although it doesn't matter as currently the expressions don't generate `Comp`. 

I'm not sure `Comp` is actually needed, but I've kept it around just in case... Longer story is that any `Expr` written with `Comp` should be able to be re-written without `Comp` I think.  Previously when `Expr` contained what is now `Circuit`, `Comp` was used to connect some of the `Circuit` parts.

blaxill (approved):
LGTM, the comment on `Comp` is just an FYI more than anything.

cava/cava/Cava/Arrow/CircuitFunctionalEquivalence.v
jadephilipoom: I'll add in a case for it anyway, just so we don't get surprised later. Thanks for the catch!

PR #236, opened 2020-10-07
Prove facts about identity monad and combinational

satnam6502:
This PR ports over some of the proofs about monads that @jadephilipoom wrote in https://github.com/project-oak/oak-hardware/commit/b7364dc5923cac09dbc1281a1f6099b5b278cc65 which I have tried to simplify a bit for my comprehension. I've checked that these definitions allow the ripple-carry adder correctness to still be proved (with minor adaption).

I've added comments that show the goals as they develop. I appreciate this may be poor form? Especially since we can't expect these to be stable from one release of Coq to another?


jadephilipoom (approved):
I think the goals-in-comments style is just fine! I've done the exact same thing when I've tried to write tutorial/demo proofs. Even though different versions of Coq might change the goal slightly, it probably won't change dramatically, and it helps to see what the writer of the proof was thinking.

PR #241, opened 2020-10-08
Add direct instantiation of vectors from lists

blaxill:
This fixes a quadratic type annotation blow up in expressions that
occurs due to large sequences of Cons.


cava/cava/Cava/Arrow/CavaNotation.v
```diff
@@ -243,6 +243,19 @@ Module KappaNotation.
     (in custom expr at level 2,
     x constr at level 4, no associativity) : kappa_scope.
 
+
+  Notation "{ x , y , .. , z }" :=
+    (
+      let ls := (cons x (cons y .. (cons z nil) ..)) in
+      (Primitive
+      (ConstantVec
+        (List.length ls)
```
jadephilipoom, line 252: Are you certain this is what you want? You might want to wrap it in a `Resize` to avoid getting `length` terms everywhere. You can probably even infer the size of the `Resize`, I think, i.e. `Resize _ (ConstantVec (List.length ls) ...)`. If the current version doesn't cause problems, great, but maybe this is something to keep in mind if you do find yourself having to resize a lot because of sizes in terms of `length`.

jadephilipoom (approved):
LGTM!

cava/cava/Cava/Arrow/CavaNotation.v
blaxill: The current solution is not perfect but the expressions fail to unify if the length is left to be inferred. Since the list is always statically sized I think any vector length should be able unify by computation (without any resizing) but there will definitely be `length`s floating around in the term. I'll merge this as-is for now, and self flagellate if this brings problems later.

Issue #242, opened 2020-10-08
Monad-based AES block top-level definition


jadephilipoom:
Following up on conversation with @satnam6502 :

We've discussed me trying to verify a monad-based version of the AES block, both to have a point of comparison for proof effort and to have a "backup option" in case the performance issues I'm currently running into with arrows aren't solvable. For the first pass at this, I'd want to work on a similar proof to what I'm currently working on in arrow-land, which means some version of the whole AES block defined in terms of its sub-components. It can be specialized to AES-256, and the loop can be either unrolled or not. It's okay to leave the subroutines (key expansion, `mix_columns`, `sub_bytes`, `shift_rows`, `add_round_key`) as `Axiom`s; I don't need to unfold them or inspect them for this proof. It's also okay to leave things at the level of `arrow-examples/Combinators.v` (for instance, vector reversal) as axioms for now. I've been looking at `unrolled_cipher_naive` for the current proof, so something similar to that will provide the best point of comparison in the short-term. However, if it's a lot of effort to implement, it might be better to go with the more updated, stateful version that we'd need eventually anyway.

satnam6502:
Closing this for now via #304 but please do open new issues for further changes or refinements.

PR #243, opened 2020-10-09
High-level Gallina AES spec

jadephilipoom:
AES forward cipher and equivalent inverse cipher, based on the [FIPS documentation](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.197.pdf). The equivalent inverse cipher is the version OpenTitan implements. This is about the level of abstraction I need for the top-level AES cipher block in #242.

OpenTitan implements both cipher directions by threading through a boolean argument (`op_i`) that says whether it should do the inverse cipher or not, so a more faithful version would look something like:
```
  Definition cipher (op_i : bool) (first_key last_key : key) (middle_keys : list key)
             (input : state) : state :=
    let st := input in
    let st := add_round_key st first_key in
    let st := fold_left
                (fun (st : state) (round_key : key) =>
                   let st := sub_bytes op_i st in
                   let st := shift_rows op_i st in
                   let st := mix_columns op_i st in
                   let st := add_round_key st round_key in
                   st)
                middle_keys st in
    let st := sub_bytes op_i st in
    let st := shift_rows op_i st in
    let st := add_round_key st last_key in
    st.
```
...where the subroutines also do either the forward or the inverse version depending on `op_i`.

### DIFF
--- a/cava/cava/Cava/Spec/Aes.v
+++ b/cava/cava/Cava/Spec/Aes.v
@@ -1,0 +1,131 @@
+Require Import Coq.Lists.List.
+
+Hint Rewrite @map_app @fold_left_app : listsimpl.
+Local Ltac listsimpl :=
+  repeat progress (cbn [rev map fold_left]; autorewrite with listsimpl).
+
+Module Spec.
+  Context (state key : Type)
+          (add_round_key : state -> key -> state)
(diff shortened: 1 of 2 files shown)


satnam6502 (approved):
Thank you!

PR #246, opened 2020-10-10
Update gitignore files, cleanup clean

satnam6502:
Originally I've strived to get to the point where when we do `make` at the top level followed by `make clean` and then do `git status` it should come out clean (i.e. no files should be shown). We're drifted away from that point and this PR aims to get us back to a clean git status after clean.

Also `clean` should be idempotent i.e.
```consle
$ make clean
$ make clean
```
should not give an error.


blaxill:
> should not give an error.

What error were you seeing?

satnam6502:
I think I might be wrong here... I think there might have been an error but the Makefile is set to ignore it.

PR #250, opened 2020-10-11
Add pinmux build to the system build

satnam6502:
Add the generation of the Silver Oak to the system build (and pre-submit) and also check the generated SystemVerilog for our pinmux by linting it with Verilator in situ with the dependent SystemVerilog files.

### DIFF
--- a/cava/opentitan/Makefile
+++ b/cava/opentitan/Makefile
@@ -1,5 +1,5 @@
 #
-# Copyright 2019 The Project Oak Authors
+# Copyright 2020 The Project Oak Authors
 #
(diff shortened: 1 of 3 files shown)


cava/Makefile
```diff
@@ -23,12 +23,7 @@ all:
 	cd monad-examples/xilinx && $(MAKE) extraction
 	cd tests/xilinx && $(MAKE) extraction
 	cd arrow-examples && $(MAKE)
-
-coq:
```
blaxill, line 27: I'm not sure why this was removed?

Issue #213, opened 2020-09-22
Build organization


jadephilipoom:
The story for building this project is a little jumbled right now -- lots of different `_CoqProject` files, for instance, and some directories require you to run `make` in other directories first to build dependencies. I think it would be smoother if we had:

- One top-level `_CoqProject` file, no other `_CoqProject`s
- A `make` target that makes it possible to automatically populate the `_CoqProject` file (maybe using `git ls-files` or similar)
- DIfferent top-level `make` targets for different logical subparts of the project (e.g. `make arrow-lib`, `make monad-examples`)

This is not something I'm intending on doing at this exact moment, and it should probably be prioritized behind some other things we're all busy with. But I figured I'd open an issue so that whenever one of us does have a moment, we'll have already discussed it and figured out what makes the most sense!

blaxill:
I think it also makes sense to remove remove other inactive sub directories keeping only cava, third_party and nix, to make the focus of the project clearer. It might also be worth thinking about a better structuring of directories under `/cava`, that perhaps can address lack of namespacing better (Coq seems to mostly care about the immediate module name, e.g. `A.C` `B.C` produce name collision warnings and extract poorly?)

jadephilipoom:
> I think it also makes sense to remove remove other inactive sub directories keeping only cava, third_party and nix, to make the focus of the project clearer.

:+1: 

One consideration for naming directories under `cava/` is that for COQPATH-based builds it's a requirement that the root-level source directory name matches the library name (so, for instance, the directory `arrow-lib` would need to be renamed to `Arrow` because those files are compiled and imported as `Arrow.Foo`). I think this only becomes a problem if someone using a COQPATH-based build decides they want to have us as a dependency, but it's relatively easy to do and not a bad organizational idea anyway.

satnam6502:
Soon I would like to reorganize the repo so that the Cava code is at the top level and everything else is relegated to `experiments` or some such directory. As for the `_CoqProject` level organization, this is why I did things the way they are right now:
* cava/cava/Cava is support to represent the "Cava system" as a self-contained entity containing the core combinators/DSLs and libraries of useful circuits.
* cava/monad-exampkes is supposed to represent some "examples" that use the Cava library so I gave this code its own `_CoqProject`. Likewise for `arrow` and `tests`.

Is this the right way to organize a Coq library and clients of the library? Or should we just one one `_CoqProject` file at the root?

jadephilipoom:
[...]

PR #253, opened 2020-10-12
A first cut at reorganizing the Silver Oak repo

satnam6502:
To address issue #213 this PR makes the following changes:
* The cava work is promoted to the root of the repo.
* Other work is moved into the `investigations` directory.
* `arrow-lib` for the moment a top-level library.
* A level of hierarchy has been collapsed: `cava/cava` -> `cava` which in the past has caused me some confusion.
* Some stuff has been consolidated into a `doc` directory.


blaxill (approved):
Lgtm! I wonder if we can get `cava/Cava/` down to just `cava/`?

PR #257, opened 2020-10-13
More cleanup and add Lava investigation when merging

satnam6502:


blaxill (approved):
LGTM. Perhaps call it `More cleanup and add Lava investigation` when merging

PR #259, opened 2020-10-13
Remove nix from install instructions

satnam6502:
Users should just have to type:
```console
$ make
```
to build most of the repo (that excludes Xilinx tools specific targets).
Nix is used for CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,14 +10,6 @@ requests.
 
 ## Pre-requisites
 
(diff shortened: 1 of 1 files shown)


blaxill (approved):

blaxill:
I guess an alternative could be to just relegate the instructions lower in the file as an alternative option

PR #261, opened 2020-10-13
Check parallelism

blaxill:
There appears to be some non-determinism in GCB (our CI) which sometimes keeps/creates the `.git` folder but mostly strips it. 

This PR is failing until I fix the interdependencies in the makefiles


blaxill:
Superseded by  #306

PR #262, opened 2020-10-13
Don't install globally by default

blaxill:


blaxill:
Please try this PR out and let me know if it creates problems in your environments.

satnam6502 (approved):

satnam6502:
I've merged and we can revert if required.

satnam6502:
`make clean` and `make` works for me. Thank you.
Should I now do a global uninstall of some things?

satnam6502:
Perhaps I spoke too soon, I get this error in VS Code when I try to step through code:

```coq
From Cava Require Import Kind.
From Cava Require Import Types. (* This fails *)
```

```
Compiled library Cava.Types (in file /Users/satnam/oak-hardware/cava/Cava/Types.vo) makes inconsistent assumptions over library ExtLib.Structures.Monad
Not in proof mode.
```

jadephilipoom:
I uninstalled the dependencies and re-built them, no issues :+1:

Issue #268, opened 2020-10-14
Transient CI build errors


satnam6502:
This issue tracks transient build errors on CI. Whenever I notice one I'll add a note here.

satnam6502:
https://pantheon.corp.google.com/cloud-build/builds/ffad1c6d-3abd-47e7-a41f-4643b8ea4b14?project=oak-ci

```console
Step #0 - "recover_git_folder": Cloning into 'oak-hardware-git'...
Step #0 - "recover_git_folder": POST git-upload-pack (142 bytes)
Step #0 - "recover_git_folder": POST git-upload-pack (445 bytes)
Step #0 - "recover_git_folder": mv: cannot move 'oak-hardware-git/.git' to './.git': Directory not empty
Finished Step #0 - "recover_git_folder"
ERROR
ERROR: build step 0 "nixos/nix:2.3.4" failed: step exited with non-zero status: 1
```

blaxill:
The cause is actually the start of the build, GCB is starting to more frequently initialize a git repository:

```
FETCHSOURCE
Initialized empty Git repository in /workspace/.git/
From https://github.com/project-oak/oak-hardware
 * branch            cd8c0b08123b0b318ce6dbeb12622d3971c78ee1 -> FETCH_HEAD
HEAD is now at cd8c0b0 Merge arrow-lib back into Cava
```

Previously it would not 

```
FETCHSOURCE
Fetching storage object: gs://691249393555.cloudbuild-source.googleusercontent.com/b3eb8b460810e9d7f73c1278623fc1779fc7fabc-e30ff79c-2123-43ee-b469-56dbfd84fd4c.tar.gz#1602639191658367
Copying gs://691249393555.cloudbuild-source.googleusercontent.com/b3eb8b460810e9d7f73c1278623fc1779fc7fabc-e30ff79c-2123-43ee-b469-56dbfd84fd4c.tar.gz#1602639191658367...
/ [0 files][    0.0 B/ 33.2 MiB]                                                
/ [1 files][ 33.2 MiB/ 33.2 MiB]                                                
Operation completed over 1 objects/33.2 MiB.  
```

This behaviour is non-deterministic as far as I can tell. I think for now I'll update the build steps to be resilient to either initialisation routine.

blaxill:
Rebasing on main since #269 (commit 382a11b on main branch) *should* now work.

satnam6502:
Got a build failure for Verilator.
```
[...]

PR #270, opened 2020-10-14
CI: Don't fail if .git exists

blaxill:
This will hopefully solve the transient build errors ( #268 ) 

I can't guarantee it as I cannot manually trigger the GCB build difference


satnam6502 (approved):
Thank you kindly for all your great work on our CI system.

PR #271, opened 2020-10-14
Derive specifications from Kappa representation

### LINKED ISSUE #249: COQC CombinatorProperties.v is too slow to compile

jadephilipoom:
I wasn't getting anywhere on improving arrow performance, so I backtracked a little farther and tested out what happens if I derive the spec from the `kappa` representation instead. With several strategic `Opaque` declarations, the performance issues pretty much disappeared. `CombinatorProperties.v` went down from 26s to 1.4s compilation on my machine, and I was able to derive a spec for `unrolled_cipher_naive'` and typecheck its correctness proof!

One drawback of this change is that now we will need to write a correctness proof for `combinational_evaluation`, which was not required with the previous design. I'll probably start working on that right after this is merged, just to get an early warning if that proof turns out to be complex.

@blaxill now that there's a strategy that works for deriving specs of larger circuits, is `unrolled_cipher_naive` the right circuit definition to use?

Fixes #249

### DIFF
--- a/arrow-examples/Aes/CipherProperties.v
+++ b/arrow-examples/Aes/CipherProperties.v
@@ -1,0 +1,181 @@
+(****************************************************************************)
+(* Copyright 2020 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
+(* you may not use this file except in compliance with the License.         *)
+(* You may obtain a copy of the License at                                  *)
+(*                                                                          *)
+(*     http://www.apache.org/licenses/LICENSE-2.0                           *)
+(*                                                                          *)
+(* Unless required by applicable law or agreed to in writing, software      *)
+(* distributed under the License is distributed on an "AS IS" BASIS,        *)
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+Require Import Coq.Strings.String.
+From Coq Require Import Derive.
+From Arrow Require Import Category Arrow.
+From Cava Require Import Arrow.ArrowExport Arrow.CircuitFunctionalEquivalence
+     BitArithmetic Tactics VectorUtils.
+
+From ArrowExamples Require Import CombinatorProperties Aes.cipher_round Aes.unrolled_naive_cipher.
+
+Module Vector.
+  (* matches pkg.aes_transpose; uses snoc/unsnoc instead of cons/tl *)
+  Fixpoint transpose_rev {A n m} : Vector.t (Vector.t A m) n -> Vector.t (Vector.t A n) m :=
+    match n with
+    | O => fun _ => Vector.const (Vector.nil _) _
+    | S n' =>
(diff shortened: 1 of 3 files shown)


blaxill:
> One drawback of this change is that now we will need to write a correctness proof for `combinational_evaluation`

Is this to match evaluation on the kappa expression with the evaluation of an arrow expression? 

>@blaxill now that there's a strategy that works for deriving specs of larger circuits, is unrolled_cipher_naive the right circuit definition to use?

`unrolled_naive_cipher.v` is a simpler version of `unrolled_opentitan_cipher.v` which in turn is a simpler version of the synchronous version that I haven't got around to merging.

I think the `unrolled_opentitan_cipher.v` in particular will be informative, and `unrolled_naive_cipher.v` optional - e.g. we might want to depart further from the design of the OpenTitan modules for simplicity. The `unrolled_naive_cipher.v` version should be the easiest to prove as its the most direct from the spec, e.g. the `unrolled_opentitan_cipher` mixes the key expansion with the cipher rounds.

arrow-examples/Aes/CipherProperties.v
```diff
@@ -0,0 +1,181 @@
...
+  Lemma aes_transpose_correct n m (x : Vector.t (Vector.t (Vector.t bool _) _) _) :
+    kinterp (@pkg.aes_transpose n m) (x, tt) = Vector.transpose_rev x.
+  Proof.
+    revert m x; induction n; cbn [pkg.aes_transpose Vector.transpose_rev];
+      kappa_spec; [ reflexivity | ].
+    repeat destruct_pair_let. cbn [fst snd].
+    autorewrite with vsimpl. reflexivity.
+  Qed.
+  Hint Rewrite @aes_transpose_correct : kappa_interp.
+  Opaque pkg.aes_transpose.
```
blaxill, line 97: Are these Opaque to control unfolding in proof? This looks like another place that a function environment with an external call constructor would help/be an alternative?

blaxill (approved):
LGTM

arrow-examples/Aes/CipherProperties.v
jadephilipoom: Specifically to control unfolding in the typechecking. The proof goes fine but `Qed` hangs because it tries to compute `interp_combinational` for all the callees. Quantifying over a function environment would indeed fix that.

jadephilipoom:
> Is this to match evaluation on the kappa expression with the evaluation of an arrow expression?

I typo'ed, actually meant to say `closure_conversion`. But still yes -- our ultimate objective is to connect the Gallina spec to the netlist implementation, so if we connect the kappa expression to the spec, we need to connect the kappa expression to the netlist implementation, which means we have to prove each stage in that chain.

> I think the `unrolled_opentitan_cipher.v` in particular will be informative, and `unrolled_naive_cipher.v` optional - e.g. we might want to depart further from the design of the OpenTitan modules for simplicity. The `unrolled_naive_cipher.v` version should be the easiest to prove as its the most direct from the spec, e.g. the `unrolled_opentitan_cipher` mixes the key expansion with the cipher rounds.

Okay, maybe I'll switch to working on the `unrolled_opentitan_cipher` now and move back to `naive` if I'm struggling with it. Thanks!

PR #272, opened 2020-10-15
Remove outdated import

jadephilipoom:
Caused by #269 and #271 being merged at about the same time, so the new file created in #271 had an import that #269 invalidated. This commit removes the bad import (which is now subsumed by `Arrow.ArrowExport`, so no further imports needed to be added).


blaxill:
Is this ready for review? Lgtm, feel free to merge straight away

jadephilipoom:
> Is this ready for review? Lgtm, feel free to merge straight away

Yep, my bad, just forgot to hit "request". Thanks!

PR #273, opened 2020-10-15
Proof that closure conversion preserves semantics

jadephilipoom:
This proof went...much easier than I expected! The punchline is this theorem from `ExprLoweringPreservation.v`:
```
Theorem closure_conversion_preserves_semantics i o (expr : Kappa i o) x :
  Wf expr -> NoLetRec expr ->
  interp_combinational (expr coq_func) x = combinational_evaluation
                                             (closure_conversion expr) x.
```
The restrictions are that a) the Kappa expression must be `Wf`, meaning that it doesn't change its structure based on the `var` you pass in, and b) that there can't be `LetRec` constructors in the expression (because otherwise it wouldn't be combinational).

Other changes included here are:
- change `var` type to `Kind -> Type` instead of `Kind -> Kind -> Type`. The first (input/source) type was always `Unit`, so the argument provided no useful information or generality, and would have required bookkeeping in the proofs to prove that everything in the context was `var Unit ?`
- removed the `Kappa_equivalence` axiom because it was untrue as stated (something of type `forall var, kappa var i o` could do something silly like branch on the type of `var`, so quantifying over everything of that type is too strong); added a `Wf` precondition to the one place the axiom was used
- move `rewrite_or_default_refl` proof to the new file, because it was actually unused there but was necessary for the semantics-preservation proof

### DIFF
--- a/arrow-examples/CombinatorProperties.v
+++ b/arrow-examples/CombinatorProperties.v
@@ -75,16 +75,7 @@ Section Misc.
              | _ => rewrite map2_swap, map2_drop
              end.
   Qed.
-
-  Lemma rewrite_or_default_refl A :
-    circuit_equiv _ _ (rewrite_or_default A A) (fun x => x).
-  Proof.
-    induction A; cbn [rewrite_or_default] in *; arrowsimpl;
-      circuit_spec; autorewrite with vsimpl;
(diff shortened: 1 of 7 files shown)


blaxill (approved):
Lgtm thanks Jade

PR #275, opened 2020-10-15
Add Let and RemoveContext cases to kappa_equivalence

jadephilipoom:
All `kappa` constructors are now included, which paves the way to proving that the AES cipher implementations are `Wf`.

### DIFF
--- a/cava/Cava/Arrow/ExprEquiv.v
+++ b/cava/Cava/Arrow/ExprEquiv.v
@@ -44,6 +44,18 @@ Set Asymmetric Patterns.
 
     | Prim_equiv : forall E p, kappa_equivalence E (ExprSyntax.Primitive p) (ExprSyntax.Primitive p)
 
+    | Let_equiv : forall x y z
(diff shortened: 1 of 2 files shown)


cava/Cava/Arrow/ExprEquiv.v
```diff
@@ -44,6 +44,18 @@ Set Asymmetric Patterns.
 
     | Prim_equiv : forall E p, kappa_equivalence E (ExprSyntax.Primitive p) (ExprSyntax.Primitive p)
 
+    | Let_equiv : forall x y z
```
blaxill, line 47: Oh it was the non-recursive version missing?! I must have accidentally removed it, I was thinking it was the `Letrec` case missing.

blaxill (approved):
Lgtm

cava/Cava/Arrow/ExprEquiv.v
jadephilipoom: Yep, that's the one!

PR #280, opened 2020-10-16
Prove Wf for arrow combinators and unrolled_cipher_naive

jadephilipoom:
This is not yet plugged into proofs, but serves as a sanity check that the circuits follow the preconditions needed to prove that `closure_conversion` preserves semantics (see #273). I'm leaving `Wf` admitted for the same subcircuits that I've left admitted for the equivalence proofs.


blaxill (approved):
Lgtm

PR #282, opened 2020-10-16
Prove spec and Wf of unrolled_opentitan_cipher definitions

jadephilipoom:
The proofs are analogous to the ones in `NaiveCipherProperties.v`, and are stated in terms of  of the subroutines `cipher_round`, `final_cipher_round`, `mix_columns`, and `key_expand`.

### DIFF
--- a/arrow-examples/Aes/OpenTitanCipherProperties.v
+++ b/arrow-examples/Aes/OpenTitanCipherProperties.v
@@ -1,0 +1,145 @@
+(****************************************************************************)
+(* Copyright 2020 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
+(* you may not use this file except in compliance with the License.         *)
+(* You may obtain a copy of the License at                                  *)
+(*                                                                          *)
+(*     http://www.apache.org/licenses/LICENSE-2.0                           *)
+(*                                                                          *)
+(* Unless required by applicable law or agreed to in writing, software      *)
+(* distributed under the License is distributed on an "AS IS" BASIS,        *)
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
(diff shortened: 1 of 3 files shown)


arrow-examples/Aes/OpenTitanCipherProperties.v
```diff
@@ -0,0 +1,145 @@
...
+                     (round : Vector.t bool 4),
+                      kinterp (key_expand_and_round sbox_impl)
+                              (state, (round, tt))
+                      = key_expand_and_round_spec sbox_impl state round)
+         As key_expand_and_round_correct.
+  Proof.
+    cbv [key_expand_and_round]; kappa_spec.
+    repeat destruct_pair_let. cbn [fst snd].
+    rewrite <-!surjective_pairing.
+    derive_spec_done.
```
blaxill, line 111: I might be getting confused, but is this defining the spec from the implementation?

blaxill (approved):
If I'm reading this right, it turns out mixing the key scheduling in with the rounds was not

arrow-examples/Aes/OpenTitanCipherProperties.v
jadephilipoom: Yep, exactly. The plan here is that I'm going to derive a spec, and then I'm going to prove the derived version equivalent to the pretty, transcribed-from-FIPS version in `silveroak-opentitan`.

jadephilipoom:
> If I'm reading this right, it turns out mixing the key scheduling in with the rounds is not adding any significant difficulty?

It'll probably require some bookkeeping when I want to prove the derived spec equivalent to the transcribed one, which doesn't have key expansion interleaved, but I don't think it'll be too hard. I'm imagining I would create a definition that, given a key expansion procedure, runs the key expansion repeatedly to generate a list of all the keys, and then state as a precondition to the equivalence that the list of keys being fed to the transcribed spec matches the list generated by key expansion. We'll see when I get there, but I don't expect major problems.

PR #285, opened 2020-10-19
Separate primitive semantics

blaxill:
This reduces duplication and removes the need for `ExprSemantics.v ` to depend on `CircuitSemantics.v `


jadephilipoom (approved):
LGTM! The change is nicely intuitive.

Issue #286, opened 2020-10-20
Acorn: Simplification of the monad kernel for Cava


satnam6502:
Device a stripped down simplified monad kernel to experiment with simplifying proofs.

PR #287, opened 2020-10-20
An experimental simplified monad kernel: Acorn.

satnam6502:
This PR introduces an experimental kernel for Cava that is a simplified version of our current monad kernel. I'm calling this kernel "Acorn". Simplifications include dispensing with the shape type and also removing smashing at the type level. The combinator definition of `col` is also simpler than what we currently have in the monad kernel.

The PR includes two lemmas in the file `Cava/Acorn/Lib/AcornUnsignedAdderProofs.v`: one for a list-based adder and one for a vector based adder. Note some useful monad facts are imported. It would be instructive to see:

* Is the list based adder proof any simpler than what happens in https://github.com/project-oak/oak-hardware/pull/218 ? 
* If it is possible to do the equivalent proof for the vector-based adder, then how do they compare? Or can this proof be done as an "add on" to the list-baed proof by relating vectors to lists?

Part of addressing #286 

CC: @blaxill

### DIFF
--- a/cava/Cava/Acorn/AcornCombinators.v
+++ b/cava/Cava/Acorn/AcornCombinators.v
@@ -1,0 +1,121 @@
+(****************************************************************************)
+(* Copyright 2020 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
+(* you may not use this file except in compliance with the License.         *)
+(* You may obtain a copy of the License at                                  *)
+(*                                                                          *)
+(*     http://www.apache.org/licenses/LICENSE-2.0                           *)
+(*                                                                          *)
+(* Unless required by applicable law or agreed to in writing, software      *)
+(* distributed under the License is distributed on an "AS IS" BASIS,        *)
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+From Coq Require Import Lists.List.
+Import ListNotations.
+From Coq Require Import Vector.
+From Coq Require Import Bool.Bvector.
+Import VectorNotations.
+Local Open Scope vector_scope.
+Require Import ExtLib.Structures.Monads.
+Import MonadNotation.
+Open Scope monad_scope.
+
+Require Export Acorn.AcornSignal.
+Require Export Acorn.AcornCavaClass.
+
+Open Scope type_scope.
+
+Definition fork2 {m signal} `{Cava m signal} {A : SignalType}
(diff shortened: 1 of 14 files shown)


jadephilipoom:
Created a draft PR to show the proofs: #289 
Full summary in the description there, but here are the short answers:

>     * Is the list based adder proof any simpler than what happens in #218 ?

Yes, slightly.

>     * If it is possible to do the equivalent proof for the vector-based adder, then how do they compare? Or can this proof be done as an "add on" to the list-baed proof by relating vectors to lists?

They're pretty similar, the main difference being that vector resizing is a little annoying. Working on relating vectors to lists and trying the proof that way, but that's once again proving a little tricky because of vector resizing issues.

jadephilipoom:
Vector-based proof is now proven by equivalence with the list-based proof!

cava/Cava/Acorn/AcornCombinators.v
```diff
@@ -0,0 +1,121 @@
...
+              ret (b0::bRest, aFinal)
+  end.
+
+Definition colL {m} `{Monad m} {A B C}
+                (circuit : A * B -> m (C * A))
+                (inputs: A * list B) :
+                m (list C * A) :=
+ colL' circuit (fst inputs) (snd inputs).
+
+(* TODO(satnam): Lemma about col_cons *)
```
jadephilipoom, line 119: Actually, it didn't end up being necessary for the proofs; since it's defined directly by recursion on the input list, I could just `cbn [colL']` in the inductive step.

cava/Cava/Acorn/AcornSignal.v
```diff
@@ -0,0 +1,55 @@
...
+  | Vec : forall {t : SignalType} {s : nat}, Vector.t (Signal t) s -> Signal (VecType t s)
+  | IndexSignal : forall {t : SignalType} {s : nat},
+                  Signal (VecType t s) -> nat -> Signal t
+  | Pair : forall {A B : SignalType}, Signal A -> Signal B -> Signal (PairType A B)
+  | Fst : forall {A B : SignalType}, Signal (PairType A B) -> Signal A
+  | Snd : forall {A B : SignalType}, Signal (PairType A B) -> Signal B.
+
+Definition peel {k: SignalType} {s: nat} (v: Signal (VecType k s)) :
+                Vector.t (Signal k) s :=
+  Vector.map (IndexSignal v) (vseq 0 s).
```
jadephilipoom, line 44: I don't see this used anywhere. Is it left over from code churn, or is there a reason it will be needed in the future?

jadephilipoom (approved):
LGTM!

cava/Cava/Acorn/AcornSignal.v
satnam6502: I suspect we will need this so leaving it in for now. Thank you.

PR #289, opened 2020-10-20
Acorn adder proofs

jadephilipoom:
On top of #287 -- see individual commits for the real diff

Adds proofs for the list- and vector-based adders using the Acorn kernel. Some thoughts:

- The list proof ended up almost exactly the same as with the old kernel. Changes included some slight simplification due to the fact that `colL` is defined directly, while the previous `col` was defined in terms of `below_cons`.
- The vector proof was almost exactly the same as the list proof, but there was some pain involved with resizing vectors (the adder produces a vector of length `n+1`, so the inductive case ends up with a length  `S n + 1` that needs to simplify to `S (n + 1)` to allow the right simplification. And of course a normal rewrite doesn't work because then the types don't line up. I ended up solving this by putting a `resize` in front of the adder right off the bat, so that by the time I do induction I'm actually proving:
```coq
let bitAddition := combinational (addLWithCinV cin a b) in
Bv2N (resize_default false (S n) bitAddition) =
Bv2N a + Bv2N b + (N.b2n cin).
```
- Besides vector resizing, there's one (minor) pain point that remaining around destructing lets (e.g. `let (x, y) := z in ..`). These are actually `match` statements under the hood, which means that I can't do any kind of rewrite in their continuation until I destruct the match -- which I usually can't do, because it's under a monad bind.

### DIFF
--- a/cava/Cava/Acorn/Lib/AcornUnsignedAdderProofs.v
+++ b/cava/Cava/Acorn/Lib/AcornUnsignedAdderProofs.v
@@ -49,16 +49,19 @@ Lemma fullAdder_correct (cin a b : bool) :
   (N.testbit sum 0, N.testbit sum 1).
 Proof. destruct cin, a, b; reflexivity. Qed.
 
-(* Lemma about how to decompose a list of vectors. *)
-Lemma list_bits_to_nat_cons b bs :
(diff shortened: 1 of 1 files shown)


jadephilipoom:
I rebased on top of #287, so ready to merge once CI passes, @satnam6502 !

satnam6502 (approved):

PR #290, opened 2020-10-20
Arrow modules

blaxill:
Changes `Arrow` layer to a shallow representation, removing the adhoc inductive structures.


cava/Cava/Arrow/Classes/Arrow.v
```diff
@@ -133,86 +124,57 @@ Proof.
...
+    void either (ASum: Arrow object category void either)
+    := {
+    merge {x} : (either x x) ~> x;
+    never {x} : void ~> x;
+  }.
+
+  (* A case that can never happen *)
+  Class Impossible `(A: Arrow) := {
+    impossible : forall {x y}, x ~> y;
+  }.
```
jadephilipoom, line 174: Is this for default/error cases?

cava/Cava/Arrow/Classes/Kleisli.v
```diff
@@ -1,35 +1,65 @@
...
+    ret (pair b (snd a));
+  second x y z f a :=
     b <- f (snd a) ;;
-    ret (fst a, b);
+    ret (pair (fst a) b);
 
-  assoc   x y z a := ret (fst (fst a), (snd (fst a), snd a));
-  unassoc x y z a := ret ((fst a, fst (snd a)), snd (snd a));
+  assoc   x y z := fun a => ret (pair (fst (fst a)) (pair (snd (fst a)) (snd a)));
+  unassoc x y z := fun a => ret (pair (pair (fst a) (fst (snd a))) (snd (snd a)));
```
jadephilipoom, line 48: The previous version of these two lines seems more readable to me than the new version. Why the change?

cava/Cava/Arrow/ExprLoweringPreservation.v
```diff
@@ -31,19 +30,29 @@ Require Import Cava.Arrow.ExprSyntax.
 (* TODO: this is copied from CircuitFunctionalEquivalence and extended; it
    should be moved to a more general location *)
```
jadephilipoom, line 31: This TODO can be removed; since `CircuitFunctionalEquivalence` was deleted, it's no longer a duplicate.

cava/Cava/Arrow/CavaNotation.v
```diff
@@ -340,7 +337,7 @@ Section regression_examples.
   Program Fixpoint tree
     (A: Kind)
     (n: nat)
-    (f: << A, A, Unit >> ~> A)
+    (f: << A, A, Unit >> ~> A )
```
jadephilipoom, line 340: nit: inserted extra space

cava/Cava/Arrow/ExprSemantics.v
```diff
@@ -17,7 +17,7 @@
 From Cava Require Import Arrow.Classes.Category.
 From Cava Require Import Arrow.Classes.Arrow.
 From Cava Require Import Arrow.CircuitArrow.
-From Cava Require Import Arrow.CircuitSemantics.
+(* From Cava Require Import Arrow.CircuitSemantics. *)
```
jadephilipoom, line 20: Just delete this import, if it's unused?

cava/Cava/Arrow/ExprSemantics.v
```diff
@@ -48,15 +48,16 @@ Section combinational_semantics.
...
     (expr: kappa coq_func x y)
     (i: denote_kind (remove_rightmost_unit x)): (denote_kind y) :=
     interp_combinational' expr (apply_rightmost_tt x i).
 
-  Axiom expression_evaluation_is_arrow_evaluation: forall i o (expr: Kappa i o), forall (x: denote_kind i),
-    combinational_evaluation' (closure_conversion expr) x =
-    interp_combinational' (expr _) x.
+  (* Axiom expression_evaluation_is_arrow_evaluation: forall i o (expr: Kappa i o), forall (x: denote_kind i), *)
+  (*   combinational_evaluation' (closure_conversion expr) x = *)
+  (*   interp_combinational' (expr _) x. *)
```
jadephilipoom, line 61: ditto, just delete? We can always add it back if it's needed.

silveroak-opentitan/aes/Arrow/sbox_canright.v
```diff
@@ -110,56 +110,55 @@ Definition canright_composed
...
 Axiom aes_sbox_canright_correct :
-  obeys_spec aes_sbox_canright aes_sbox_canright_spec.
-Axiom CircuitLaws : CategoryLaws CircuitCat.
-Existing Instance CircuitLaws.
+  forall x y, kinterp aes_sbox_canright (x,(y,tt)) = aes_sbox_canright_spec (x,(y,tt)).
 
-Hint Resolve aes_sbox_canright_correct CIPH_FWD_correct CIPH_INV_correct
-  : circuit_spec_correctness.
+Hint Rewrite aes_sbox_canright_correct CIPH_FWD_correct CIPH_INV_correct
+  : kappa_interp.
```
jadephilipoom, line 139:
```suggestion
Hint Rewrite aes_sbox_canright_correct CIPH_FWD_correct CIPH_INV_correct
  using solve [eauto] : kappa_interp.
```

`Hint Rewrite`s should always have a `using` clause that fails if it doesn't solve the goal -- otherwise they can do a rewrite that creates an unprovable side condition and sneakily make your goal false.

silveroak-opentitan/aes/Arrow/sbox_canright.v
```diff
@@ -110,56 +110,55 @@ Definition canright_composed
...
-  cbv [canright_composed]. circuit_spec.
-  subst canright_composed_spec.
-  instantiate_app_by_reflexivity.
+  cbn [canright_composed interp_combinational']; kappa_spec.
+  derive_spec_done.
 Qed.
+
 (* Uncomment below to see derived spec for canright_composed *)
-(* Print canright_composed_spec. *)
+Print canright_composed_spec.
```
jadephilipoom, line 150: nit: recomment this print statement to reduce build noise

jadephilipoom (commented):
I have some concerns about this change. Although some of the definitions have gotten _smaller_ in terms of code size, the long inductives with many branches seem _simpler_ conceptually. For instance, looking at `CircuitProp.v`, the new version is smaller but it's much harder to understand what's happening. Compare the old version:
```coq
Fixpoint no_delays {i o} (c: Circuit i o): bool :=
￼  match c with
￼  | Primitive (Delay _) => false
￼  | Composition _ _ _ f g => no_delays f && no_delays g
￼  | First _ _ _ f => no_delays f
￼  | Second _ _ _ f => no_delays f
￼  | Loopr _ _ _ f => no_delays f
￼  | Loopl _ _ _ f => no_delays f
￼  | Map _ _ _ f => no_delays f
￼  | _ => true
￼  end.

￼Fixpoint no_loops {i o} (c: Circuit i o): bool :=
￼  match c with
￼  | Composition _ _ _ f g => no_loops f && no_loops g
￼  | First _ _ _ f => no_loops f
￼  | Second _ _ _ f => no_loops f
￼  | Loopr _ _ _ f => false
￼  | Loopl _ _ _ f => false
￼  | Map _ _ _ f => no_loops f
￼  | _ => true
￼  end.
￼
￼Local Open Scope category_scope.
￼Local Open Scope arrow_scope.
￼
￼Definition is_combinational {i o: Kind} (c: i ~> o) :=
￼  no_loops c && no_delays c = true.
```
...with the new version:
```coq
  Local Notation all := (monoid_arrow Kind bool all_monoid Unit Tuple).
￼  Instance no_loops_loop : Arrows.Loop all := {
￼    loopl _ _ _ _ := false;
￼    loopr _ _ _ _ := false;
￼  }.

￼  Instance no_delays_primitives
￼    : Arrows.Primitive all
￼      CircuitPrimitive primitive_input primitive_output := {
￼    primitive p :=
￼      match p with
￼      | Delay _ => false
￼      | _ => true
￼      end;
￼  }.
￼
￼  Instance is_combinational_circuit : CircuitArrow := {}.
￼End combinational.
￼
￼Definition is_combinational {i o: Kind}
￼  (circuit: ExprSyntax.Kappa i o): bool :=
￼  (closure_conversion is_combinational_circuit circuit).
```

The previous version was verbose, but I could look at the `Circuit` inductive and look at the definition and understand what was happening pretty quickly. The second is (to me) much less clear; I can see that `loopl`, `loopr`, and `Delay` are returning `false`, but what are the other options? How is `is_combinational_circuit` pulling those definitions in? As far as I can tell, I now need to read about the `CircuitArrow` class and associated coercions, plus monoidal arrows, to be able to understand what's happening. And to be honest, I just finished fully reading and reviewing those files and I _still_ don't understand what's going on here.

If we're expecting only one instantiation of this representation, an inductive seems like a better fit in terms of representation, and proofs almost universally cooperate better with inductive structures than with typeclasses. Maybe I'm completely missing the mark here and there's some intuition I don't have that I could have if I read the right paper? For instance, I'm really not sure what the `monoid_arrow` is doing, and maybe if I saw it through that frame it would make sense?

Could you perhaps explain to me a little further why the change to a typeclass-based shallow embedding is helpful? What problem are you solving here, and why is this the right way to solve it?

cava/Cava/Arrow/CircuitSemantics.v
```diff
@@ -15,134 +15,181 @@
...
-  | Composition _ _ _ f g => (default_state f, default_state g)
-  | First x y z f => default_state f
-  | Second x y z f => default_state f
-  | Loopr x y z f => (default_state f, kind_default z)
-  | Loopl x y z f => (kind_default z, default_state f)
-  | Primitive (Delay o) => kind_default o
-  | Map x y n f => const (default_state f) _
-  | _ => tt
-  end.
+Set Typeclasses Unique Solutions.
```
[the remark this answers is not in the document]
blaxill: I was using this a sanity check to make sure the `CircuitArrow` instances had a single instance in scope when I used the empty/implicit definitions. I will remove it. Does this setting affect modules that import this module?

blaxill:
> I have some concerns about this change. 

That is fair and I agree. I don't think I've quite achieved what I was aiming for here in representation- for example I was hoping a lot of the boilerplate would fall away but hasn't. Let us continue discussing this PR before abandoning or merging.

> If we're expecting only one instantiation of this representation, an inductive seems like a better fit in terms of representation, and proofs almost universally cooperate better with inductive structures than with typeclasses. Maybe I'm completely missing the mark here and there's some intuition I don't have that I could have if I read the right paper? For instance, I'm really not sure what the `monoid_arrow` is doing, and maybe if I saw it through that frame it would make sense?
> 

The `Classes/Monoid.v` takes any monoid (type with a unit and a binary function) and provides default class instances that insert the unit (and `compose` that uses the binary function to combine two branches). The reasoning for having this is that a lot of interpretations only need to lift a default value into instance functions. For example, `is_combinational_circuit` this default value is `true` and so any instances from `Classes/Monoid.v` insert `true` and the combine definition will `&&` them. Not everything will fit into these monoid based instances but a number of things do, for example a naive critical path could be calculated from a `(nat,0,+)` monoid, and would have different weighting numbers for primitives. 

**I may well have been overzealous in applying this abstraction**- Its not clear how many circuit interpretations fit, but the motivation was simplifying definitions and composition of interpretations. The `is_combinational_circuit` could also not use the `monoid` stuffs to be more explicit, (although also probably longer than the head definition!)

> Could you perhaps explain to me a little further why the change to a typeclass-based shallow embedding is helpful? What problem are you solving here, and why is this the right way to solve it?

Although I opened by say shallow representation I actually see the difference as
- a single one-fits-all inductive representation used by all interpretations (main)
- each interpretation specifying its own deep or shallow embedding (this PR)

I was hoping to achieve better composition properties via the 2nd, as for example, being able to reuse interpretations for subclasses, and from being able to select nice morphism representations.

My personal 2 biggest unknowns/concerns with this PR approach is that: 

- **There is no longer a canonical arrow inductive structure to prove with:** Its unclear to me the cons/benefits of this- We seem to have had more success proving at the kappa expression level, and I would presume if we had a 2nd rules based language, it would also be easier to prove those expressions at their definition level. In both this PR regime and main, I can imagine you could compose two different types of expressions (via arrows) and then a proof of the whole component would decompose via lemma into their expression-type specific proofs, importantly not _using the arrow structuring_. 

- **Without a shared inductive structure, different term representations are no longer connected**: If you look at the head sequential semantics `circuit_evaluation'`, having the shared inductive `Circuit` structure allows a straightforward implementation, where as the typeclass version required sigma-types. This is somewhat obvious in retrospect but surprised me as I was writing it. 

As to "why this change at this moment?"- this was already a non-trivial LoC change, I would expect it to be even more in the future as we add interpretations. That said, It might be best to file this representation under experiments/investigations for now.

cava/Cava/Arrow/CircuitSemantics.v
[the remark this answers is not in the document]
jadephilipoom: I'm actually not sure about this one, but if you put `Local` in front then it definitely shouldn't!

jadephilipoom:
Thanks for the explanation of the `Monoid` classes; I think I see better now what's happening! It's pretty cool that so many things fit this structure, and I can see that removing the repetitive recursive structure is appealing. I'm still a little concerned about the effect on proofs.

> * **There is no longer a canonical arrow inductive structure to prove with:** Its unclear to me the cons/benefits of this- We seem to have had more success proving at the kappa expression level, and I would presume if we had a 2nd rules based language, it would also be easier to prove those expressions at their definition level. In both this PR regime and main, I can imagine you could compose two different types of expressions (via arrows) and then a proof of the whole component would decompose via lemma into their expression-type specific proofs, importantly not _using the arrow structuring_.

This one is my main concern as well. If we're never doing proofs that involve the arrow structure, then it's not a problem, but I don't want to back us into a corner where we can't do proofs if we need them. How certain are you that we won't need to do proofs at arrow level?

PR #291, opened 2020-10-21
Move Aes folder into silveroak-opentitan

### LINKED ISSUE #245: Move all OpenTitan development work under the cava/opentitan/ directory

jadephilipoom:
I believe this is the last piece not already in the OpenTitan-specific folder, so fixes #245

This is a necessary step for me to link the existing proofs about the top-level AES block up with the spec we wrote earlier. I'm aiming (although no promises) to have that link-up done for tomorrow!


jadephilipoom:
Sorry for the immediate code churn -- I realized I hadn't committed the last steps of fixing this up, so I converted to draft and pushed them. All settled now (I think)!

jadephilipoom:
The ArrowExamplesSV file had some dependencies on AES stuff that I just removed -- long-term, we'd probably want to revive those bits and put them in the opentitan folder also. @blaxill , could I lean on you to do that part separately?

blaxill (approved):
Lgtm thanks Jade

PR #292, opened 2020-10-22
Time indexed sequential semantics experiments

satnam6502:
A "note" to check in which illustrates an alternative way to attribute sequential circuits by viewing signals as functions from a time index to a value, rather than as a list of values or stream of values. Next step would be to model a circuit with a loop e.g.. a counter.

### DIFF
--- a/investigations/modeling/Makefile
+++ b/investigations/modeling/Makefile
@@ -1,0 +1,29 @@
+#
+# Copyright 2020 The Project Oak Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


jadephilipoom (approved):
Very interesting! A couple of thoughts:

1. The inputs to any circuit need to be complete functions from `nat->bool`, which means that their values across all time have to be known at the outset (or, I suppose, we could quantify over all functions from `nat->bool` that have certain values at the first `n` timesteps). In any case, they won't be able to depend on the circuit body.
2. Does this structure assume everything is using the same clock, or is "timestep" a vague enough notion to save us from that restriction?

blaxill (approved):
Lgtm. It would be interesting to investigate how this performs on larger circuits over larger time steps.

blaxill:
> 2\. Does this structure assume everything is using the same clock, or is "timestep" a vague enough notion to save us from that restriction?

Definitely our first sequential explorations should assume a single clock domain. A timestep increment unassociated to a clock is too vague as `delay` needs to trigger on a defined clock edge. Modelling arbitrary multiple clocks is going to require some thought and possibly restriction.

satnam6502:
Thanks for the great comments. It's not clear how scalable this specific formulation is. Let's experiment.

PR #297, opened 2020-10-26
Remove redundant arrow terms

blaxill:
This might have been the one pointed out previously by @jadephilipoom, but it looks like we didn't fix it yet


jadephilipoom (approved):
Yep, that's the one! I guess I never ended up merging that branch. LGTM!

PR #298, opened 2020-10-26
Helper lemmas about lists, vectors, and denote_kind_eqb

jadephilipoom:
I have managed to prove that the arrow-based AES implementation, assuming subroutines are correct, matches the top-level spec! :tada: But the punchline will come in a later PR. To get there, I had to do quite a bit of basic standard-library-level lemmas for vectors and lists, which were sizable enough to be their own PR. Changes included here:
- Move some of the proofs about `to_list` from the Acorn adder proofs to `ListUtils` so we can import them easily elsewhere
- Create some new autorewrite databases for list/vector proofs (try the new tactic `length_hammer` on your next annoying list-lengths subgoal!)
- Definition and several proofs about `fold_left_accumulate`, a variant of `fold_left` that accumulates a list of (a projection of) all the states it passed through (this was helpful when I was proving that the version of the cipher with key-generation interleaved is equivalent to the one that pre-generates all the keys)
- A few helper proofs about `N`/`positive`/`N2Bv_sized`, added to `BitArithmetic.v`
- Helper proofs about `denote_kind_eqb`, the boolean decider for `denote_kind` types

CC @satnam6502

### DIFF
--- a/cava/Cava/ListUtils.v
+++ b/cava/Cava/ListUtils.v
@@ -2,6 +2,7 @@ Require Import Coq.Arith.PeanoNat.
 Require Import Coq.Lists.List.
 Require Import Coq.micromega.Lia.
 Import ListNotations.
+Local Open Scope list_scope.
 
 Section Length.
   Lemma nil_length {A} : @length A nil = 0.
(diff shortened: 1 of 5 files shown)


cava/Cava/ListUtils.v
```diff
@@ -18,7 +41,261 @@ Section Folds.
...
 End Folds.
+Hint Rewrite @fold_left_cons @fold_left_nil
+     using solve [eauto] : push_list_fold.
+
+(* Defines a version of fold_left that accumulates a list of (a
+   projection of) all the states it passed through *)
+Section FoldLeftAccumulate.
+  Definition fold_left_accumulate' {A B C}
+             (f : B -> A -> B) (g : B -> C) acc0 ls b : list C * B :=
+    fold_left (fun '(acc, b) a => (acc ++ [g (f b a)], f b a)%list)
```
blaxill, line 127: Possibly use `Local Open Scope list_scope.`? Then you could remove the `%list`s

cava/Cava/ListUtils.v
```diff
@@ -18,7 +41,261 @@ Section Folds.
...
+    rewrite fold_left_accumulate'_length.
+    reflexivity.
+  Qed.
+
+  Lemma map_fold_left_accumulate {A B C D}
+        (f : B -> A -> B) (g : B -> C) (h : C -> D) ls b :
+    map h (fst (fold_left_accumulate f g ls b)) =
+    fst (fold_left_accumulate f (fun x => h (g x)) ls b).
+  Proof.
+    revert b.
```
blaxill, line 219: Very minor: `revert b` could possibly be removed by moving the `b` in definition to a type on the right hand side of the `:`?

blaxill (approved):
Lgtm

blaxill:
> I have managed to prove that the arrow-based AES implementation, assuming subroutines are correct, matches the top-level spec! 🎉 But the punchline will come in a later PR.

A cliffhanger PR !

cava/Cava/ListUtils.v
jadephilipoom: It could indeed; I used this pattern all over the place because I have an irrational aesthetic preference for putting forall-quantified things before the `:` and then `revert`ing what I need for induction. Is it okay if I leave it as-is?

blaxill: Fine with me!

PR #299, opened 2020-10-26
Proof that arrow-based AES top-level spec and FIPS-based spec are equivalent

jadephilipoom:
Resolves the cliffhanger I teased in #298! I first defined an alternative spec-level implementation that has key-generation interleaved, and then proved that equivalent to the original spec. Then, I proved that the derived top-level behavior of the `unrolled_opentitan_cipher` matches the alternative version.

There are a few oddities I'd like to address that came up while doing the proof; I've marked them as TODOs in the code.
1. The implementation transposes the key before performing `add_round_key`. Why?
2. It looks like the implementation does 15 rounds instead of 14 -- or am I misreading the FIPS spec?
3. The starting key pair seems to be reversed.

This PR completes phase 1 of the AES verification plan! :tada: :tada: :tada:

### DIFF
--- a/silveroak-opentitan/aes/Arrow/OpenTitanCipherEquivalence.v
+++ b/silveroak-opentitan/aes/Arrow/OpenTitanCipherEquivalence.v
@@ -1,0 +1,152 @@
+(****************************************************************************)
+(* Copyright 2020 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
+(* you may not use this file except in compliance with the License.         *)
+(* You may obtain a copy of the License at                                  *)
+(*                                                                          *)
+(*     http://www.apache.org/licenses/LICENSE-2.0                           *)
+(*                                                                          *)
+(* Unless required by applicable law or agreed to in writing, software      *)
+(* distributed under the License is distributed on an "AS IS" BASIS,        *)
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+Require Import Coq.Arith.PeanoNat.
+Require Import Coq.Lists.List.
+Require Import Coq.micromega.Lia.
+Require Import Coq.NArith.NArith.
+Require Import Coq.Vectors.Vector.
+Require Import Cava.Arrow.ArrowExport.
+Require Import Cava.Arrow.CombinatorProperties.
+Require Import Cava.Arrow.DeriveSpec.
+Require Import Cava.BitArithmetic.
+Require Import Cava.ListUtils.
+Require Import Cava.Tactics.
+Require Import Cava.VectorUtils.
+Require Import AesSpec.Cipher.
+Require Import AesSpec.ExpandAllKeys.
+Require Import AesSpec.InterleavedCipher.
+From Aes Require Import OpenTitanCipherProperties CipherRoundProperties
+     unrolled_opentitan_cipher.
+Import VectorNotations.
+Import ListNotations.
+
+Section Equivalence.
+  Local Notation byte := (Vector.t bool 8) (only parsing).
+  Context (sbox : pkg.SboxImpl)
+          (aes_key_expand_spec :
+             pkg.SboxImpl -> bool -> Vector.t bool 4 -> byte ->
+             Vector.t (Vector.t byte 4) 8 ->
+             byte * Vector.t (Vector.t byte 4) 8)
+          (aes_key_expand_correct :
+             forall sbox_impl op_i round_id rcon key_i,
+               kinterp (aes_key_expand sbox_impl)
+                       (op_i, (round_id, (rcon, (key_i, tt))))
+               = aes_key_expand_spec sbox_impl op_i round_id rcon key_i)
+          (aes_sub_bytes_correct :
+             forall sbox_impl op_i state,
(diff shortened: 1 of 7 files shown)


blaxill:
Awesome work! I will review it shortly, but to respond to your questions immediately:

> 1. The implementation transposes the key before performing `add_round_key`. Why?

This comment in OpenTitan has a good description:

https://github.com/lowRISC/opentitan/blob/94e374bd9d7c8c489f187e02678ff58e11c09316/hw/ip/aes/rtl/aes_cipher_core.sv#L19

I think it corresponds to equations 3.2/3.3/3.4 in the spec.

> 2. It looks like the implementation does 15 rounds instead of 14 -- or am I misreading the FIPS spec?

I think this is because the function `key_expand_and_round` in `unrolled_opentitan_cipher.v` handles the initial `AddRoundKey(state, w[0, Nb-1]) // See Sec. 5.1.4` as round 0, so the fold is over 14 elements rather than 13 in the spec. 

> 3. The starting key pair seems to be reversed.

The reversing at 

~~~Coq
210     let forward_keys = concat key_i[:7:4] key_i[:3:0] in
~~~

is to put the first key in the second position, so that `key_expand` does not need to mux to decided where to pull the key from. IIRC this is what happens in the OpenTitan version, but let me double check...

silveroak-opentitan/aes/Arrow/OpenTitanCipherEquivalence.v
```diff
@@ -0,0 +1,152 @@
...
+  Notation rconst := (Vector.t bool 8) (only parsing).
+  Notation keypair := (Vector.t (Vector.t (Vector.t bool 8) 4) 8) (only parsing).
+
+  Notation nat_to_bitvec size n := (Ndigits.N2Bv_sized size (N.of_nat n)).
+
+  (* TODO: this transpose seems odd *)
+  Definition add_round_key : state -> key -> state :=
+    fun st k =>
+    @bitwise (Vector (Vector (Vector Bit 8) 4) 4) (fun a b => xorb a b)
+             st (PkgProperties.Vector.transpose_rev k).
```
blaxill, line 73: To expand a bit on my other reply, the comment at the bottom of Figure 15. in FIPS 197 says 

```
Note that, since InvMixColumns operates on a two-dimensional array of bytes

while the Round Keys are held in an array of words, the call to

InvMixColumns in this code sequence involves a change of type (i.e. the

input to InvMixColumns() is normally the State array, which is considered

to be a two-dimensional array of bytes, whereas the input here is a Round

Key computed as a one-dimensional array of words).
```

silveroak-opentitan/aes/Arrow/OpenTitanCipherEquivalence.v
```diff
@@ -0,0 +1,152 @@
...
+    cbn [splitat fst snd Vector.append].
+    autorewrite with vsimpl. reflexivity.
+  Qed.
+
+  Definition add_round_key_pair : state -> keypair -> state :=
+    fun st kp => add_round_key st (sndkey kp).
+
+  Lemma unrolled_cipher_spec_equiv
+        init_keypair first_key last_key middle_keys input :
+    let nrounds := 15 in (* TODO: is this the right # rounds? *)
```
blaxill, line 102: (expanding on original reply) Merging the initial `AddRoundKey` as a pseudo round allows a smaller circuit in the sequential version, although since this version is unrolled it adds no advantage (apart from being more representational of the sequential version).

blaxill (approved):
Lgtm! I'm curious on your opinion on how the proof turned out and any issues encountered, e.g. any surprises(?), any issues with proof time/size anymore?

jadephilipoom:
I wanted to be absolutely certain we didn't have a sneaky off-by-one on the number of rounds, so I looked over FIPS again. The conclusion is "we're all good", but I'll record my reasoning here in case we need to reference it sometime.

Appendix B has a detailed unrolled walkthrough of the cipher for AES-128 (`Nr`=10), which shows:
- round 0 is only `add_round_key`
- rounds 1-9 (inclusive) perform all transformations
- round 10 skips `mix_columns`

So. counting both the first `add_round_key` and the final non-`mix_columns` as rounds, we should expect `Nr+1` rounds. `Nr` is 14 for AES-256. Put another way, we should do `Nr -1` = 13 "normal" rounds, so the `middle_keys` passed to our top-level spec should have length 13 and `all_keys` needs length 15; both of those are provable by the `length_all_keys` lemma. I've updated the name of the argument for # rounds to make it match FIPS.

jadephilipoom:
> Lgtm! I'm curious on your opinion on how the proof turned out and any issues encountered, e.g. any surprises(?), any issues with proof time/size anymore?

No issues with proof time or size! `OpenTitanCipherEquivalence.v` builds in 3.9s on my machine, and the `InterleavedCipher.v` in 0.7s. The main complexity was the interleaved key-generation, which was just a pure-Coq, DSL-independent proof that the two specs were equivalent. Then it was just pretty standard loop invariant reasoning; I didn't even have to unfold the loop.

jadephilipoom:
> This reversing _should_ only be internal to the implementation though. IIRC this is what happens in the OpenTitan version, but let me double check...

Yeah, I'm a little unhappy with the fact that reversing shows up in the top-level proof but not sure how to fix it. I have to give the key expansion and the cipher keys that are reversed from each other to get the proof to work, which makes it hard to avoid.

In experimenting with fixing this problem I pushed a change that means we no longer use key pairs at the top-level proof, but just keys -- the change pushes a little more complexity into the equivalence-between-specs proof, but I think we'd rather have the complexity there than at the top level.

PR #300, opened 2020-10-27
Add counter and fir arrow sequential examples

blaxill:
- Adds two sequential arrow examples that pass systemverilog testbench
- Updates arrow sequential semantics (but doesn't yet use them to create testbench outputs)


arrow-examples/Fir.v
```diff
@@ -0,0 +1,77 @@
...
+
+  Definition adder
+    : << Byte, Byte, Unit >> ~> Byte :=
+    <[ \x y => x +% y ]>.
+
+  Definition fir
+    : << Byte, Unit >> ~> Byte :=
+    <[ \byte =>
+      letrec window = byte :: (!drop_last window) in
+      (* TODO(blaxill): map2 adder should be mult, but we haven't added it yet *)
```
satnam6502, line 49: Please make an issue and file it against me?

arrow-examples/Counter.v
```diff
@@ -0,0 +1,59 @@
...
+     tt (mkPort "count" (Kind.Vec Kind.Bit 3)) [].
+
+Definition counter_3_netlist :=
+  makeNetlist counter_3_Interface (build_netlist' (closure_conversion (counter 3))).
+
+Definition counter_3_tb_inputs : list unit :=
+ [tt; tt; tt; tt; tt; tt; tt; tt; tt].
+
+Definition counter_3_tb_expected_outputs : list (Bvector.Bvector 3) :=
+  (* TODO(blaxill): replace with 'circuit_evaluation' *)
```
satnam6502, line 53: Make an issue and add it to the TODO e.g. `TODO(blaxill, #1066): replace...`

arrow-examples/Counter.v
```diff
@@ -0,0 +1,59 @@
...
+  Local Open Scope category_scope.
+  Local Open Scope kind_scope.
+
+  Definition counter n
+    : << Unit >> ~> Vector Bit n :=
+    <[
+      letrec counter = counter +% #1 in
+      counter
+    ]>.
+End notation.
```
satnam6502, line 35: Taking a guess, this generates a synthesiable adder that has no bit-growth and unsigned `mod` semantics.

arrow-examples/Counter.v
```diff
@@ -0,0 +1,59 @@
...
+
+Open Scope kind_scope.
+
+Require Import Cava.Types.
+Require Import Cava.Netlist.
+
+Definition counter_3_Interface :=
+   sequentialInterface "counter_3" "clk" PositiveEdge "rst" PositiveEdge
+     tt (mkPort "count" (Kind.Vec Kind.Bit 3)) [].
```
satnam6502, line 44: Add a comment to say this circuit has no inputs, hence `tt`.

satnam6502 (approved):
I'm probably missing context here... I can see why we need to focus on a counter, but why do we need a FIR example just yet?

arrow-examples/Counter.v
```diff
@@ -0,0 +1,59 @@
...
+
+Definition counter_3_netlist :=
+  makeNetlist counter_3_Interface (build_netlist' (closure_conversion (counter 3))).
+
+Definition counter_3_tb_inputs : list unit :=
+ [tt; tt; tt; tt; tt; tt; tt; tt; tt].
```
jadephilipoom, line 50: Minor, but this could be changed to `repeat tt 8` if you don't need/want it fully computed.

jadephilipoom (approved):
LGTM!

blaxill:
> I'm probably missing context here... I can see why we need to focus on a counter, but why do we need a FIR example just yet?

I'm not sure I follow, the FIR example is pretty trivial, do you not want to many examples? From a regression PoV vs the counter example it covers (recursive) non-trivial state, that is dependent on some non-recursive input. The counter alone wasn't sufficient to cover some of the netlist/semantic evaluation logic I wanted to check.

arrow-examples/Counter.v
blaxill: Yeah, the notation was added in #162. I'm open to removing it though and only having sugar for one adder to simplify.

PR #303, opened 2020-10-28
Use 'circuit_evaluation' to generate the sequential tb outputs

### LINKED ISSUE #301: Replace constant values in arrow sequential examples with evaluation via `circuit_evaluation`

blaxill:
Also leave `rewrite_or_default` to arrow interpreter, rather than performing `rewrite_or_default` as a construction of arrow constructors.

Fixes #301

### DIFF
--- a/cava/Cava/Arrow/ArrowKind.v
+++ b/cava/Cava/Arrow/ArrowKind.v
@@ -236,6 +236,8 @@ Fixpoint apply_rightmost_tt (x: Kind)
   | _ => fun x => x
   end.
 
+(* Avoid eq_rect type equality rewriting by inductively matching the Kind
+* structure. *)
 Fixpoint rewrite_or_default (x y: Kind): denote_kind x -> denote_kind y :=
   match x as x' return denote_kind x' -> denote_kind y with
   | Unit =>
--- a/cava/Cava/Arrow/ArrowKind.v
+++ b/cava/Cava/Arrow/ArrowKind.v
@@ -236,3 +236,27 @@ Fixpoint apply_rightmost_tt (x: Kind)
   | _ => fun x => x
   end.
 
+Fixpoint rewrite_or_default (x y: Kind): denote_kind x -> denote_kind y :=
+  match x as x' return denote_kind x' -> denote_kind y with
+  | Unit =>
+      match y with
+      | Unit => fun a => a
+      | _ => fun _ => kind_default _
+      end
+  | Tuple l r =>
+      match y with
+      | Tuple ll rr => fun '(a,b) => (rewrite_or_default l ll a, rewrite_or_default r rr b)
+      | _ => fun _ => kind_default _
+      end
+  | Vector t n =>
+      match y with
+      | Vector t2 n2 => fun a => VectorUtils.resize_default (kind_default _) _ (Vector.map (rewrite_or_default t t2) a)
+      | _ => fun _ => kind_default _
+      end
+  | Bit =>
+      match y with
+      | Bit => fun a => a
+      | _ => fun _ => kind_default _
+      end
+  end.
+
--- a/cava/Cava/Arrow/CircuitArrow.v
+++ b/cava/Cava/Arrow/CircuitArrow.v
@@ -64,8 +64,7 @@ Inductive Circuit: Kind -> Kind -> Type :=
   | Loopr: forall x y z, Circuit (Tuple x z) (Tuple y z) -> Circuit x y
   | Loopl: forall x y z, Circuit (Tuple z x) (Tuple z y) -> Circuit x y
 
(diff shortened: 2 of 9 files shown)


cava/Cava/Arrow/ArrowKind.v
```diff
@@ -236,3 +236,27 @@ Fixpoint apply_rightmost_tt (x: Kind)
...
+      match y with
+      | Vector t2 n2 => fun a => VectorUtils.resize_default (kind_default _) _ (Vector.map (rewrite_or_default t t2) a)
+      | _ => fun _ => kind_default _
+      end
+  | Bit =>
+      match y with
+      | Bit => fun a => a
+      | _ => fun _ => kind_default _
+      end
+  end.
```
satnam6502, line 261: A comment to explain what this does?

cava/Cava/Arrow/CircuitLowering.v
```diff
@@ -130,6 +130,31 @@ Fixpoint map2M (f: Signal Kind.Bit -> Signal Kind.Bit -> Instance) (ty: Kind)
...
+      match y with
+      | Vector t2 n2 => fun a => VectorUtils.resize_default (const_wire _ (kind_default _)) _ (Vector.map (rewrite_or_default t t2) a)
+      | _ => fun _ => (const_wire _ (kind_default _))
+      end
+  | Bit =>
+      match y with
+      | Bit => fun a => a
+      | _ => fun _ => (const_wire _ (kind_default _))
+      end
+  end.
```
satnam6502, line 155: How and why is this different from the earlier `rewrite_or_default`?

cava/Cava/Arrow/CircuitSemantics.v
```diff
@@ -132,10 +129,28 @@ Definition combinational_evaluation {x y: Kind}
...
+  | x :: xs =>
+    let '(o,ns) := circuit_evaluation circuit x state
+    in o :: unroll_circuit_evaluation' circuit ns xs
+  end%list.
+
+Definition unroll_circuit_evaluation {x y: Kind}
+  (circuit: x ~> y)
+  (i: list (denote_kind (remove_rightmost_unit x)))
+  : list (denote_kind y) :=
+  unroll_circuit_evaluation' circuit (default_state _) i.
```
satnam6502, line 155: So this "simulates" the circuit as a sequential circuit, giving a default initial state?

cava/Cava/Arrow/CircuitArrow.v
```diff
@@ -64,8 +64,7 @@ Inductive Circuit: Kind -> Kind -> Type :=
   | Loopr: forall x y z, Circuit (Tuple x z) (Tuple y z) -> Circuit x y
   | Loopl: forall x y z, Circuit (Tuple z x) (Tuple z y) -> Circuit x y
 
-  | Map: forall x y n, Circuit x y -> Circuit (Vector x n) (Vector y n)
-  | Resize: forall x n nn, Circuit (Vector x n) (Vector x nn)
+  | RewriteTy: forall x y, Circuit x y
```
jadephilipoom, line 67: Curious about why this ended up being generalized (the first version only works on `Vector`, the new one works for any types). Is this because of nested vectors?

cava/Cava/Arrow/CircuitArrow.v
```diff
@@ -64,8 +64,7 @@ Inductive Circuit: Kind -> Kind -> Type :=
   | Loopr: forall x y z, Circuit (Tuple x z) (Tuple y z) -> Circuit x y
   | Loopl: forall x y z, Circuit (Tuple z x) (Tuple z y) -> Circuit x y
 
-  | Map: forall x y n, Circuit x y -> Circuit (Vector x n) (Vector y n)
```
jadephilipoom, line 67: Was this really the only usage of `Map`? If so, this change is definitely a win for simplicity!

jadephilipoom (commented):
LGTM!

cava/Cava/Arrow/CircuitLowering.v
blaxill: The typing is different, this is in terms of netlist wiring, the other is in terms of regular coq terms.

cava/Cava/Arrow/CircuitSemantics.v
blaxill: Yep!

cava/Cava/Arrow/CircuitArrow.v
blaxill: When we were addressing the performance issues of working at the arrow layer I was thinking of adding combinators (map fold etc) as constructors to the Arrows, but since we went in a different direction this wasnt necessary. As your comment says below, it seems a win for simplicity to remove these two constructors and instead have a `resize_or_default` constructor and handle that individually.

PR #313, opened 2020-10-29
OpenTitan aes_cipher_core replacement

blaxill:
~~Requires #341 or similar changes before this will pass CI~~

This is the `aes_cipher_core` to be a drop in for the OpenTitan version. There are a few outstanding items I will track in #357

### DIFF
--- a/silveroak-opentitan/.gitignore
+++ b/silveroak-opentitan/.gitignore
@@ -1,0 +1,10 @@
+Makefile.coq
+.Makefile.coq.d
+*.sv
+!testbench/*
+*.vcd
+*.vvp
+*.hs
+*.cpp
+*.tcl
+obj_dir/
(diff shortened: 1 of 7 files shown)


silveroak-opentitan/.gitignore
```diff
@@ -0,0 +1,10 @@
+Makefile.coq
```
jadephilipoom, line 1: Adding this `.gitignore` seems like a good move, but it seems to me like it's not logically part of this change. Unless the change creates a bunch more generated files?

jadephilipoom (approved):

silveroak-opentitan/.gitignore
blaxill: I copied one of the existing .gitignore contents, to try and hopefully not miss additional files (alternatively, perhaps its too aggressive!)

PR #314, opened 2020-10-30
Cleanup primitive structures

blaxill:
- This seems to be a performance degradation somewhat, I am investigating options

### DIFF
--- a/silveroak-opentitan/aes/Arrow/_CoqProject
+++ b/silveroak-opentitan/aes/Arrow/_CoqProject
@@ -18,7 +18,6 @@ unrolled_naive_cipher.v
 unrolled_opentitan_cipher.v
 cipher_round.v
 key_expand.v
-cipher_control.v
 netlists.v
 aes_test.v
 CipherRoundProperties.v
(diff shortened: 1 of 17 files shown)


silveroak-opentitan/aes/Arrow/_CoqProject
```diff
@@ -18,7 +18,6 @@ unrolled_naive_cipher.v
 unrolled_opentitan_cipher.v
 cipher_round.v
 key_expand.v
-cipher_control.v
```
jadephilipoom, line 21: Is this supposed to be removed?

jadephilipoom (approved):
LGTM!

silveroak-opentitan/aes/Arrow/_CoqProject
blaxill: Yeah, type unification is particularly flakey in the combination of this definition and the custom notation it uses. I will introduce a better fix for this in upcoming notational changes

jadephilipoom:
This actually broke `silveroak-opentitan/aes/`; the CI didn't catch it because of the issue mentioned in https://github.com/project-oak/oak-hardware/pull/328#issuecomment-720485345

CI should be able to catch the issue now.

PR #317, opened 2020-10-30
Acorn-based AES equivalence

### LINKED ISSUE #315: Prove that monad-based AES matches spec
Now that #242 is complete, prove that the implementation there matches the AES specification.

jadephilipoom:
Resolves #315 

This proof was very easy! Probably had a lot to do with the fact that the implementation was based on the spec. A good follow-up to this would be to create an Acorn-based AES implementation that's closer to OpenTitan, and see how we do with that.

Also included in this PR is a typo fix and a couple of proofs about the new `foldLM` definition -- only one is currently used, but the other might be useful later.

### DIFF
--- a/cava/Cava/Acorn/AcornSignal.v
+++ b/cava/Cava/Acorn/AcornSignal.v
@@ -39,13 +39,13 @@ Inductive Signal : SignalType -> Type :=
   | Fst : forall {A B : SignalType}, Signal (Pair A B) -> Signal A
   | Snd : forall {A B : SignalType}, Signal (Pair A B) -> Signal B.
 
-Fixpoint denoteCombinaional (t : SignalType) : Type :=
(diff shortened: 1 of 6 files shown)


cava/Cava/Acorn/AcornSignal.v
```diff
@@ -39,13 +39,13 @@ Inductive Signal : SignalType -> Type :=
   | Fst : forall {A B : SignalType}, Signal (Pair A B) -> Signal A
   | Snd : forall {A B : SignalType}, Signal (Pair A B) -> Signal B.
 
-Fixpoint denoteCombinaional (t : SignalType) : Type :=
+Fixpoint denoteCombinational (t : SignalType) : Type :=
```
satnam6502, line 42: Ah, thank you!

satnam6502 (approved):

Issue #318, opened 2020-10-30
Remove m from typeclass context of Acorn circuits


satnam6502:
The `m` typeclass variable should really be a member of the typeclass with only `signal` being the principle parameter.

PR #319, opened 2020-10-30
Remove m as typeclass variable

satnam6502:
This removes `m` as a typeclass variable and makes it a member instead. This simplifies the type context for Acorn circuits.
Addresses #318

### DIFF
--- a/cava/Cava/Acorn/AcornCavaClass.v
+++ b/cava/Cava/Acorn/AcornCavaClass.v
@@ -20,7 +20,8 @@ From Cava Require Import Acorn.AcornSignal.
 
 Open Scope type_scope.
 
-Class Cava m `{Monad m} (signal : SignalType -> Type) := {
+Class Cava (signal : SignalType -> Type) := {
+  m : Type -> Type;
   one : signal Bit;
   zero : signal Bit;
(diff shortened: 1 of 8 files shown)


cava/Cava/Acorn/AcornCavaClass.v
```diff
@@ -20,7 +20,8 @@ From Cava Require Import Acorn.AcornSignal.
 
 Open Scope type_scope.
 
-Class Cava m `{Monad m} (signal : SignalType -> Type) := {
+Class Cava (signal : SignalType -> Type) := {
+  m : Type -> Type;
```
blaxill, line 24: (To your preference) You could likely put the `Monad m` constraint inside the Cava class if you'd choose to. Although I'm not sure on exactly the notation. The straightforward version might work:

~~~
  m : Type -> Type;
  monad_constraint : Monad m;
~~~

blaxill (approved):
Lgtm!

cava/Cava/Acorn/AcornCavaClass.v
satnam6502: I did experiment with that at first, but I think that causes some other issues so backtracked to just having `m` in the class. Thank you.

PR #306, opened 2020-10-28
Run `make` in parallel mode for presubmit

jadephilipoom:
If it works, fixes #248 

I have no idea if this works, but I need to create a PR for it for the build to run.

PR #328, opened 2020-11-02
Clean up logic in Makefiles

jadephilipoom:
I wanted to get rid of some of the repetitive code in our Makefiles. In particular, we have a lot of sub-makes, and logic was repeated for every subdirectory. I also wanted to make the Makefiles a little more uniform. This PR:
- Inserts a `SUBDIRS` variable in each Makefile with sub-makes that contains all the subdirectories, and makes all the logic dependent on that variable. Now, to add a subdirectory, you don't need to go through and copy code in several targets; you need only to add its name to `SUBDIRS` and, if it has dependencies on other subdirectories, add a `foo: bar` target to declare `bar` should be built before `foo`.
- Adds `.vo` targets to every Makefile, including the higher-level ones; you can now call `make cava/Cava/Acorn/AcornCombinators.vo` on the top-level directory and it will run `make -C cava Cava/Acorn/AcornCombinators.vo`
- No longer cleans `third_party` when you run `make clean` at the top level; you have to run `make cleanall` to clean `third_party` -- I know I very rarely need to clean `third_party`, and it takes a while to build

### DIFF
--- a/third_party/Makefile
+++ b/third_party/Makefile
@@ -34,6 +34,8 @@ install: update
 	cd bedrock2 && $(MAKE) install_coqutil && $(MAKE) install_bedrock2
 	cd coq-ext-lib && $(MAKE) install
 
+coq: all
+
 clean:
 	-cd bedrock2 && $(MAKE) clean_all
 	-cd coq-ext-lib && $(MAKE) clean
--- a/Makefile
+++ b/Makefile
@@ -29,13 +29,6 @@ SUBDIRS = third_party cava tests monad-examples arrow-examples silveroak-opentit
 
 .PHONY: all coq clean subdirs $(SUBDIRS)
 
-# if the "make clean" or "make coq" targets were called, pass these to subdirs
-ifeq ($(MAKECMDGOALS), coq)
-SUBDIRTARGET="coq"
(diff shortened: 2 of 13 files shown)


jadephilipoom:
This has increased build time by 40s because I made a mistake in #306 that meant we weren't actually building `silveroak-opentitan/aes` at all. The `pinmux` and `aes` targets weren't declared as `.PHONY`, so `make` looked to see whether the targets already existed (which of course they did, because they're directories), and decided there was nothing to do. (This change declares all `$(SUBDIRS)` as `.PHONY`, so we should be protected from this particular brand of mistake in the future!)

third_party/Makefile
```diff
@@ -34,6 +34,8 @@ install: update
 	cd bedrock2 && $(MAKE) install_coqutil && $(MAKE) install_bedrock2
 	cd coq-ext-lib && $(MAKE) install
 
+coq: all
```
blaxill, line 37: I guess this is for consistency ?

blaxill (approved):
Lgtm

third_party/Makefile
jadephilipoom: Yes, because the `coq` target is passed to all subdirectories including `third_party`.

PR #330, opened 2020-11-02
De-noise `make`

### LINKED ISSUE #305: `make` is very noisy
Sometimes it's hard to tell what's going on because `make` produces so much output. There are three main sources, as far as I can tell:

1. Complaints about files with duplicate names where the names are not fully-qualified (this is pretty bad, because which file gets loaded is [undefined behavior](https://coq.inria.fr/distrib/current/refman/language/core/modules.html#libraries-and-filesystem):
```
*** Warning: in file Cava/Signal.v, 
    required library String matches several files in path
    (found String.v in ../third_party/coq-ext-lib/theories/Data and ../third_party/bedrock2/deps/coqutil/src/coqutil/Datatypes; used the latter)
```
2. Entering and leaving directories (I think there's a `make` option to suppress these):
```
cd cava && make coq
make[1]: Entering directory '/home/jadep/oak-hardware/cava'
make -f Makefile.coq
make[2]: Entering directory '/home/jadep/oak-hardware/cava'
make[3]: Nothing to be done for 'real-all'.
make[2]: Leaving directory '/home/jadep/oak-hardware/cava'
make[1]: Leaving directory '/home/jadep/oak-hardware/cava'
```
3. Extraction-related warnings:
```
File "./Extraction.v", line 64, characters 0-28:
Warning: The following logical axioms were
encountered: aes_sbox_canright_correct BitArithmetic.nat_of_bits_sized_n.
Having
invalid logical axiom in the environment when extracting may lead to
incorrect or non-terminating ML terms.
 [extraction-logical-axiom,extraction]
```

It would be nice to do something about these. Number (3) is a warning we might actually want to pay attention to and continue receiving, but perhaps we can reduce the sheer number of those that are produced or offer a `make quiet` target.

jadephilipoom:
Resolves #305 

- Suppresses `extraction-opaque-accessed` warnings using `Makefile.coq.local` files
- Removes/comments out some unused admitted lemmas and axioms so they don't cause extraction warnings
- Fully qualifies *all* imports to avoid warnings and ambiguities (most of the LOC changes are this)
- Inserts `INSTALLDEFAULTROOT` to every `_CoqProject` so that `coq_makefile` doesn't complain
- Reorders the `-R` arguments in `_CoqProject`s with the current directory first to avoid `not a subdirectory of the current directory` warnings
- Changes some non-recursive `Fixpoint`s to `Definition`s

Much quieter `make` now!

### DIFF
--- a/cava/Cava/Monad/Combinators.v
+++ b/cava/Cava/Monad/Combinators.v
@@ -75,7 +75,7 @@ Local Open Scope monad_scope.
 -------------------------------------------------------------------------------
 *)
 
-Fixpoint below `{Monad m} {A B C D E F G}
+Definition below `{Monad m} {A B C D E F G}
              (r : A * B -> m (D * G)%type)
              (s : G * C -> m (E * F)%type)
              (abc : A * (B * C)) : m ((D * E) * F)%type :=
@@ -307,7 +307,7 @@ Fixpoint treeList {T: Type} {m} `{Monad m}
            circuit aS bS
   end.
 
-Fixpoint treeWithList {T: Type} {m bit} `{Cava m bit}
+Definition treeWithList {T: Type} {m bit} `{Cava m bit}
                       (circuit: T -> T -> m T) (def: T)
                       (n : nat) (v: Vector.t T (2^(n+1))) : m T :=
   treeList circuit def n (to_list v).
--- a/cava/Cava/Monad/Combinators.v
+++ b/cava/Cava/Monad/Combinators.v
@@ -14,16 +14,15 @@
 (* limitations under the License.                                           *)
 (****************************************************************************)
 
-From Coq Require Import Vector.
+From Coq Require Import Vectors.Vector.
 Import VectorNotations.
 
-From Coq Require Import Lists.List Lia.
+From Coq Require Import Lists.List micromega.Lia.
 Import ListNotations.
 
 Require Import ExtLib.Structures.Monads.
 Require Import ExtLib.Structures.MonadFix.
-From Coq Require Arith.PeanoNat.
-Require Import Omega.
+From Coq Require Import Arith.PeanoNat.
 
 Export MonadNotation.
 
@@ -35,7 +34,6 @@ Require Import Cava.Tactics.
 
 Generalizable All Variables.
 
-From Coq Require Import Lia.
 Require Import ExtLib.Structures.MonadLaws.
 
 Local Open Scope monad_scope.
--- a/cava/Cava/Signal.v
+++ b/cava/Cava/Signal.v
@@ -14,9 +14,9 @@
 (* limitations under the License.                                           *)
 (****************************************************************************)
 
-From Coq Require Import Ascii String.
-From Coq Require Import ZArith.
-From Coq Require Import Vector.
+From Coq Require Import Strings.Ascii Strings.String.
+From Coq Require Import ZArith.ZArith.
+From Coq Require Import Vectors.Vector.
 
 From Cava Require Import Kind.
 From Cava Require Import VectorUtils.
--- a/arrow-examples/ArrowAdderTutorial.v
+++ b/arrow-examples/ArrowAdderTutorial.v
@@ -19,7 +19,7 @@ From Cava Require Import Arrow.ArrowExport.
 Require Import Coq.Strings.String.
 Local Open Scope string_scope.
 
-From Coq Require Import Lists.List NArith Lia.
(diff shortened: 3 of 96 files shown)


cava/Cava/Monad/Combinators.v
```diff
@@ -77,7 +75,7 @@ Local Open Scope monad_scope.
 -------------------------------------------------------------------------------
 *)
 
-Fixpoint below `{Monad m} {A B C D E F G}
+Definition below `{Monad m} {A B C D E F G}
```
satnam6502, line 78: Ah, thanks. I kept meaning to fix this!

cava/Cava/Signal.v
```diff
@@ -14,9 +14,9 @@
 (* limitations under the License.                                           *)
 (****************************************************************************)
 
-From Coq Require Import Ascii String.
-From Coq Require Import ZArith.
-From Coq Require Import Vector.
+From Coq Require Import Strings.Ascii Strings.String.
+From Coq Require Import ZArith.ZArith.
+From Coq Require Import Vectors.Vector.
```
satnam6502, line 19: What's the reason for this qualification?

satnam6502 (approved):
Thank you very much for this great cleanup. Not building third_party by default is great!

cava/Cava/Signal.v
jadephilipoom: If two files have the same name, which one gets loaded is undefined behavior. From the [manual](https://coq.inria.fr/distrib/current/refman/language/core/modules.html#libraries-and-filesystem):
> If several files with identical base name are present in different subdirectories of a recursive loadpath, which of these files is found first may be system- dependent and explicit qualification is recommended. The From argument of the Require command can be used to bypass the implicit shortening by providing an absolute root to the required file (see Compiled files).

The manual suggests that `From Coq` should constrain this enough, but it doesn't seem to. From this very file, we got:
```
*** Warning: in file Cava/Signal.v, 
    required library String matches several files in path
    (found String.v in ../third_party/coq-ext-lib/theories/Data and ../third_party/bedrock2/deps/coqutil/src/coqutil/Datatypes; used the latter)
```
For some reason, Coq actually ended up pulling `String` imports from `coqutil` instead of the standard library (which ended up okay because `coqutil.Datatypes.String` exports `String.String`, but still not great). Still not quite sure why this is happening, but I went ahead and fully-qualified everything, even imports that weren't generating warnings, just to be absolutely sure there's no ambiguity even if a new file pops up somewhere with a matching name.

satnam6502:
So I think we should avoid this situation by ensuring we never use the same name for a file as one that already exists or is imported?

jadephilipoom:
The third-party libraries are exactly the case where it's hard. I'm not aware of any `qualified Import` analogue in Coq, unfortunately.

I do agree that it's a bit of a waste of developer time to have everyone write `Require Import Nat`, and then `Locate Nat` to figure out where it's coming from, and then `Require Import Coq.Init.Nat`. Not to mention if they don't fully-qualify the import and then it gets mentioned in a code review, so they have to fix it and their change may be delayed. So I have another idea that I've been thinking about -- what if we had a script that could run as a githook or just exist in `tools` for people to run before submitting a PR, and it looked for the imports, figured out what their fully-qualified name was, and inserted all the `Require Import` statements in a fully-qualified and agreed-upon format (maybe even in alphabetical order/removing duplicates/_possibly_ removing unused imports). There might be some work already in this direction we could use -- I haven't looked into it much yet. But I think that could be a nice win/win: avoiding ambiguity, getting neat code, and avoiding developer time wastage.

jadephilipoom:
I thought I remembered that Jason Gross had something of the sort. This repo has a `minimize_requires` script that seems to do exactly what we want (remove unneeded imports and fully-qualify names), and there's a `normalize_requires` function (in the `replace_imports` script) separates each into their own line and orders them: https://github.com/JasonGross/coq-tools

MIT licensed, so I don't think there should be an issue with us using it. Although it would be another dependency, which is something to consider.

satnam6502:
> I thought I remembered that Jason Gross had something of the sort. This repo has a `minimize_requires` script that seems to do exactly what we want (remove unneeded imports and fully-qualify names), and there's a `normalize_requires` function (in the `replace_imports` script) separates each into their own line and orders them: https://github.com/JasonGross/coq-tools
> 
> MIT licensed, so I don't think there should be an issue with us using it. Although it would be another dependency, which is something to consider.

I was going to ask if such a thing existed (for GHC there is a lint rule for this). This seems like a good plan.

PR #332, opened 2020-11-04
Cipher.v: state/key/add_round_key

smore-lore:
This uses Vector.t to implement state and key, and it adds a
definition of add_round_key and proves it is its own inverse.

- suggestions for using VectorUtils.v appreciated (export vnil?)
- Adding nc/nb to context instead of making them dependently typed was causing some issues in my proofs where
  it couldn't match 'nc' and 'nc0' with each other.

### DIFF
--- a/silveroak-opentitan/aes/Spec/AddRoundKey.v
+++ b/silveroak-opentitan/aes/Spec/AddRoundKey.v
@@ -1,0 +1,61 @@
+(* Copyright 2020 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
+(* you may not use this file except in compliance with the License.         *)
+(* You may obtain a copy of the License at                                  *)
+(*                                                                          *)
+(*     http://www.apache.org/licenses/LICENSE-2.0                           *)
+(*                                                                          *)
+(* Unless required by applicable law or agreed to in writing, software      *)
+(* distributed under the License is distributed on an "AS IS" BASIS,        *)
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+Require Import Coq.Bool.Bvector Coq.Vectors.VectorDef.
+Require Import Cava.VectorUtils.
+
+Section Spec.
+  Context (nc : nat) (nb : nat).
+  Definition state := t (Bvector nb) nc.
+  Definition key := Bvector nb.
+
+  (* FIPS 197, 5.1.4 AddRoundKey() Transformation *)
+  Definition add_round_key (input : state) (k : key) : state :=
+    map (fun col => BVxor _ col k) input.
+
+  Section Properties.
+    Lemma inverse_xorb (b1 : bool) (b2 : bool) : xorb (xorb b1 b2) b2 = b1.
+    Proof.
+      destruct b1, b2; reflexivity.
+    Qed.
+
+    Lemma inverse_bvxor n (a : t bool n) (b : t bool n) :
+                        BVxor n (BVxor n a b) b = a.
+    Proof.
+      induction a.
+      { eapply case0 with (v:=b). reflexivity. }
+      { unfold BVxor.
+        do 2 rewrite map2_cons.
+        simpl.
+        rewrite inverse_xorb.
+
+        unfold BVxor in IHa.
+        rewrite IHa.
+        reflexivity. }
+    Qed.
+
+    Theorem inverse_add_round_key (x : state) (k : key) :
+      add_round_key (add_round_key x k) k = x.
+    Proof.
+      unfold add_round_key.
+      cbv [state] in *.
+      induction x; [ reflexivity | ].
+      { simpl.
+        rewrite IHx.
+        rewrite inverse_bvxor.
+        reflexivity. }
+    Qed.
+  End Properties.
+End Spec.
--- a/silveroak-opentitan/aes/Spec/_CoqProject
+++ b/silveroak-opentitan/aes/Spec/_CoqProject
@@ -3,6 +3,7 @@ INSTALLDEFAULTROOT = AesSpec
 -R ../../../cava/Cava Cava
 -R ../../../third_party/coq-ext-lib/theories ExtLib
 -R ../../../third_party/bedrock2/deps/coqutil/src/coqutil coqutil
+AddRoundKey.v
 Cipher.v
 ExpandAllKeys.v
 InterleavedCipher.v


silveroak-opentitan/aes/Spec/AddRoundKey.v
```diff
@@ -0,0 +1,64 @@
...
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+Require Import Coq.Bool.Bool Coq.Vectors.VectorDef.
+Require Import Cava.VectorUtils.
+
+Section Spec.
+  Definition state (nc : nat) (nb : nat) : Type := t (t bool nb) nc.
+  Definition key (nb : nat) : Type := t bool nb.
```
satnam6502, line 21: This is a vector of bool so you could write:
```coq
Vector.t bool nb
```
Even better would be to import the `Bvector` package and write:
```coq
From Coq Require Import Bool.Bvector.

Definition key (nb : nat) := Bvector nb.
```
Or just make a type synonym for `key` where you need it e.g. using `Notation`.

silveroak-opentitan/aes/Spec/AddRoundKey.v
```diff
@@ -0,0 +1,64 @@
...
+
+  Definition xor_vec n (a : t bool n) (b : t bool n) : t bool n :=
```
satnam6502, line 23: I suspect this is in one of the standard libraries that has a `Bvector` instance.

silveroak-opentitan/aes/Spec/AddRoundKey.v
```diff
@@ -0,0 +1,64 @@
...
+    map2 xorb a b.
+
+  (* FIPS 197, 5.1.4 AddRoundKey() Transformation *)
+  Definition add_round_key nc nb (input : state nc nb) (k : key nb) :
```
satnam6502, line 27: I would make `nb` an implicit argument.
```coq
Definition add_round_key nc {nb : nat} (input : state nc nb) (k : key nb) :
```

satnam6502:
Thanks! That's a great start! I'll defer to the others to complete this review.

silveroak-opentitan/aes/Spec/AddRoundKey.v
smore-lore: Thanks, done, nicer.

silveroak-opentitan/aes/Spec/AddRoundKey.v
smore-lore: Done

silveroak-opentitan/aes/Spec/AddRoundKey.v
smore-lore: Nice suggestion, thanks & done.

satnam6502:
Thanks for putting several of us down as reviewers. I propose that since @jadephilipoom  is leading the space work as soon as she gives GitHib Approval it is fine to merge if that's fine with @cipher1024 and @blaxill 
I don't think we have a policy on what to do with PRs with multiple reviewers?

satnam6502 (approved):

smore-lore:
> I don't think we have a policy on what to do with PRs with multiple reviewers?

In general - I like multiple reviewers or being included on changes even if I'm not necessarily the one to approve, since reading a lot of code gives you context for what is going on. I expected anyone with access and who was comfortable with my change to approve and merge without necessarily everyone reviewing.

blaxill:
> I don't think we have a policy on what to do with PRs with multiple reviewers?

The SWE team has a policy that one review is required before merging, but you can request as many as you like to get feedback. I think we should go with that.

silveroak-opentitan/aes/Spec/AddRoundKey.v
```diff
@@ -0,0 +1,61 @@
...
+    Lemma inverse_xorb (b1 : bool) (b2 : bool) : xorb (xorb b1 b2) b2 = b1.
+    Proof.
+      destruct b1, b2; reflexivity.
+    Qed.
+
+    Lemma inverse_bvxor n (a : t bool n) (b : t bool n) :
+                        BVxor n (BVxor n a b) b = a.
+    Proof.
+      induction a.
+      { eapply case0 with (v:=b). reflexivity. }
```
blaxill, line 38: (Your approach is fine) Another approach that can sometimes work in the face of `destruct` failing due to dependent types is to refine with a match on the variable:

```Coq
    Goal forall b,
     BVxor 0 (BVxor 0 [] b) b = [].
    Proof.
      intros.
            intros.
      refine match b with
             | nil _ => _
             end; auto.
    Qed.
```

blaxill (approved):

cipher1024:
Because `BVxor` is defined in terms of `map`, I was suggesting to @smore-lore to use the functorial laws to reduce the proof 
to the property that `xor` on single bits is idempotent. What is your opinion on this approach @jadephilipoom, @blaxill and @satnam6502?

silveroak-opentitan/aes/Spec/AddRoundKey.v
```diff
@@ -0,0 +1,61 @@
...
+      { unfold BVxor.
+        rewrite map2_cons, map2_cons.
```
blaxill, line 40: (Again totally optional, might be useful for future proofs) Some alternative ways to write this :

```Coq
        rewrite !map2_cons. (* or *)
        rewrite ?map2_cons. (* or *)
        do 2 rewrite map2_cons. (* or *)
        repeat rewrite map2_cons. 
```

satnam6502:
> Because `BVxor` is defined in terms of `map`, I was suggesting to @smore-lore to use the functorial laws to reduce the proof
> to the property that `xor` on single bits is idempotent. What is your opinion on this approach @jadephilipoom, @blaxill and @satnam6502?

Sure, you could `unfold BVxor` and see if you can apply the `map2_*` lemmas?

silveroak-opentitan/aes/Spec/AddRoundKey.v
```diff
@@ -0,0 +1,61 @@
...
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+Require Import Coq.Bool.Bvector Coq.Vectors.VectorDef.
+Require Import Cava.VectorUtils.
+
+Section Spec.
+  Definition state (nc : nat) (nb : nat) := t (Bvector nb) nc.
+  Definition key (nb : nat) := Bvector nb.
```
jadephilipoom, line 21: Thanks for taking the suggestion to move this to its own file. Just to respond to your last comment on the topic [(discussion from the earlier post of this change)](https://github.com/smore-lore/oak-hardware/commit/09f910d7c16f9a90f3e38b87c8bc73b29e330e88#r43689206):

> This makes sense - I'm curious how exactly we keep all the types aligned then? If one file declares a key which is incompatible with another file, we'd only find out in a 3rd file which tries to use the functionality together?
> 
> Should there be some basic type outline we check all these files against?

Yes, there will be files that thread these pieces together and will fail if they are incompatible -- the places where we prove that some implementation is equivalent to the specification will plug type arguments into the specification, and rely on proofs that plug type arguments into any subroutine specifications.

silveroak-opentitan/aes/Spec/AddRoundKey.v
```diff
@@ -0,0 +1,61 @@
...
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+Require Import Coq.Bool.Bvector Coq.Vectors.VectorDef.
+Require Import Cava.VectorUtils.
+
+Section Spec.
+  Definition state (nc : nat) (nb : nat) := t (Bvector nb) nc.
```
jadephilipoom, line 20: `nc` and `nb` could probably be declared with `Context` here, instead of being passed individually to each definition. `Context` will insert the variable as an argument only for definitions within the section that _use_ the variable, so the type signatures here would not be changed.

silveroak-opentitan/aes/Spec/AddRoundKey.v
```diff
@@ -0,0 +1,61 @@
...
+        rewrite IHa.
+        reflexivity. }
+    Qed.
+
+    Theorem inverse_add_round_key nc nb (x : state nc nb) (k : key nb) :
+      add_round_key nc (add_round_key nc x k) k = x.
+    Proof.
+      unfold add_round_key.
+      induction x.
+      { reflexivity. }
```
jadephilipoom, line 54: It's completely fine as-is, but when I have a very easily-solved induction base case I often save myself curly braces this way:
```suggestion
      induction x; [ reflexivity | ].
```

The `[ reflexivity | ]` syntax means "expect two goals, and run `reflexivity` on the first one".

jadephilipoom (approved):
Great work! Thanks for putting this together.

jadephilipoom:
> Adding nc/nb to context instead of making them dependently typed was causing some issues in my proofs where
it couldn't match 'nc' and 'nc0' with each other.

I recommended this change in my review, and am curious where issues popped up. Is it possible that the proof introduced another `nc`? Proofs that use context variables will end up forall-quantified over those variables, so writing a proof that says `forall nc` within a section that already declared `nc` will insert two competing versions of `nc` and possibly produce the issue you described.

I see that in `inverse_add_round_key` you do induction on a `state`, which could cause issues with the `nc` variable not matching up as well. If that's where you're running into issues, feel free to ignore the suggestion and leave as-is!

jadephilipoom:
> Because `BVxor` is defined in terms of `map`, I was suggesting to @smore-lore to use the functorial laws to reduce the proof
> to the property that `xor` on single bits is idempotent. What is your opinion on this approach @jadephilipoom, @blaxill and @satnam6502?

I think this is a great approach! And actually, I believe you could now move the proofs that are just about bitvectors to `cava/Cava/BitArithmetic.v`, the file for generic proofs about bit vectors, so as to minimize complexity in `Spec`.

smore-lore:
re jadephilipoom - this is the error I get when I move nc/nb into a context in inverse_add_round_key when I try to do induction on x. Couldn't figure out why I was introducing another nc0, so this is why I made it dependently typed. It's probably a limitation of my coq knowledge. Perhaps we could change it when/if it becomes more important, unless you want to help out!

```
Abstracting over the terms "nc" and "x" leads to a term
fun (nc0 : nat) (x0 : t (Bvector nb) nc0) =>
map (fun col : Vector.t bool nb => BVxor nb col k)
  (map (fun col : Vector.t bool nb => BVxor nb col k) x0) = x0
which is ill-typed.
Reason is: Illegal application: 
The term "@eq" of type "forall A : Type, A -> A -> Prop"
cannot be applied to the terms
 "state" : "Set"
 "map (fun col : Vector.t bool nb => BVxor nb col k)
    (map (fun col : Vector.t bool nb => BVxor nb col k) x0)" : "t (Vector.t bool nb) nc0"
 "x0" : "t (Bvector nb) nc0"
The 2nd term has type "t (Vector.t bool nb) nc0" which should be coercible to 
"state".
```

silveroak-opentitan/aes/Spec/AddRoundKey.v
smore-lore: Thanks for the tip!

silveroak-opentitan/aes/Spec/AddRoundKey.v
smore-lore: Done.

silveroak-opentitan/aes/Spec/AddRoundKey.v
smore-lore: Perfect, that makes sense.

silveroak-opentitan/aes/Spec/AddRoundKey.v
smore-lore: Done, thanks for the tip.

smore-lore:
> I think this is a great approach! And actually, I believe you could now move the proofs that are just about bitvectors to `cava/Cava/BitArithmetic.v`, the file for generic proofs about bit vectors, so as to mi

These proofs are pretty specific to this operation, and I didn't want to carve out a whole new section there for inverse_xorb. Let me keep them here for now, and if they are needed by something else, we can move them. I'm also worried I may want to change the key types used here after we start looking at the other operations.

jadephilipoom:
> re jadephilipoom - this is the error I get when I move nc/nb into a context in inverse_add_round_key when I try to do induction on x. Couldn't figure out why I was introducing another nc0, so this is why I made it dependently typed.

The issue is that the context variable "fixes" the size of the vector, so doing induction doesn't work because the proof doesn't have a notion of a `state` with any other length. Try putting in `cbv [state] in *` just before doing induction, so `x` is seen as `Vector.t` instead of `state`?

silveroak-opentitan/aes/Spec/AddRoundKey.v
smore-lore: done, w/ your help

smore-lore:
I think this is ready to merge as discussed (if someone can - I don't have perms). Will talk to Simon about cleaning up these proofs in a separate PR.

PR #333, opened 2020-11-05
Clean up key-expansion-interleaving proofs

jadephilipoom:
Reduces the complexity of the definitions and proofs in `silveroak-opentitan/aes/Spec/InterleavedCipher.v`, which defines a version of the top-level AES cipher that interleaves key expansion with cipher rounds and proves that it's equivalent to the version with keys precomputed (`Cipher.v`). The key trick here is a generic proof about changing the representation of keys in the top-level spec.

Previously, we proved that the interleaved version was equivalent to the precomputed-keys version _defined over a different key type_, which was necessary because the existing AES implementation operates on pairs of keys instead of single keys. That introduced a lot of extra boilerplate like a different definition of `add_round_key` for each key type. With a generic key-representation change lemma, we can instead write the interleaved key expansion <-> precomputed keys proof assuming the two have the same key type, and change the representation before using the lemma.

This change also let me simplify the definitions and proofs in `ExpandAllKeys.v`, which previously had to pass around a round constant. With less painful key representation changes, we can effectively treat the AES implementation's key type as `rconst * keypair`, so tracking it when precomputing the keys is not necessary.

### DIFF
--- a/silveroak-opentitan/aes/Arrow/OpenTitanCipherEquivalence.v
+++ b/silveroak-opentitan/aes/Arrow/OpenTitanCipherEquivalence.v
@@ -89,10 +89,13 @@ Section Equivalence.
     let init_rcon := nat_to_byte 1 in
     (* initial key pair reversed so key_expand doesn't have to mux *)
     let init_keypair_rev := sndkey init_keypair ++ fstkey init_keypair in
-    let all_keypairs := all_keys key_expand Nr (init_rcon, init_keypair_rev) in
-    (* project out the forward key from the pair and transpose it *)
-    let all_keys := List.map (fun kp => PkgProperties.Vector.transpose_rev (sndkey kp))
-                             (List.map snd all_keypairs) in
+    (* key_expand state is rconst * keypair *)
+    let init_rk := (init_rcon, init_keypair_rev) in
+    let all_rcons_and_keypairs := all_keys key_expand Nr init_rk in
+    (* representation change: project out the forward key and transpose it *)
+    let all_keys := List.map (fun x : rconst * keypair =>
+                                PkgProperties.Vector.transpose_rev (sndkey (snd x)))
+                             all_rcons_and_keypairs in
     all_keys = (first_key :: middle_keys ++ [last_key])%list ->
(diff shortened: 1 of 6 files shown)


silveroak-opentitan/aes/Arrow/OpenTitanCipherEquivalence.v
```diff
@@ -88,18 +89,31 @@ Section Equivalence.
...
+    map_inversion Hall_keys; subst.
+    match goal with H : @eq (list (_ * keypair)) _ (_ :: _ ++ [_])%list |- _ =>
+                    rename H into Hall_keys end.
+
+    (* representation change; use full key-expansion state (rconst * keypair) *)
+    erewrite cipher_change_key_rep with
+        (projkey := fun x => PkgProperties.Vector.transpose_rev (sndkey (snd x)))
+      by reflexivity.
+
+    erewrite <-cipher_interleaved_equiv by eassumption.
```
jadephilipoom, line 116: Just to point out the relevant code, this is the part of the "arrow-based-implementation <-> spec" lemma where we use the representation change. The first rewrite changes the RHS from `cipher state key add_round_key sub_bytes shift_rows mix_columns` to `cipher state (rconst * keypair) (fun st k => add_round_key st (transpose (sndkey (snd k)))) sub_bytes shift_rows mix_columns`; it's carrying around the extra information needed for key expansion, but it's ignoring that information. However, then the types match what key expansion expects, so the second rewrite can replace the new RHS with the interleaved-key-expansion version of the cipher, which is closer to the implementation.

blaxill (approved):
Lgtm! By representation change- are you talking about projecting the elements from the key tuple? Or are there additional transforms? (Just curious)

jadephilipoom:
> Lgtm! By representation change- are you talking about projecting the elements from the key tuple? Or are there additional transforms? (Just curious)

The keys actually need to be transposed also, like you helped me figure out a few weeks ago when we were going through the AES code. So I take the full key-expansion state -- `(round_constant, key1 ++ key2)` -- and I return e.g. `transpose key2`, which is all the spec cares about.

Issue #334, opened 2020-11-05
Proofs for top-level inverse cipher (both Arrow- and Acorn-based)


jadephilipoom:
#233 and #315 were for proving that the _forward_ AES cipher matches the spec; for both the Arrow and Acorn kernels, we should also prove that the _inverse_ version matches the spec.

PR #336, opened 2020-11-09
Interleaved inverse cipher spec

jadephilipoom:
This is progress towards #334 ; the proof that the AES inverse cipher implementation matches the main spec (which uses precomputed keys) will be done by proving that the implementation matches this interleaved version, and then using the proof that the interleaved version is equivalent to the precomputed-keys version.

Includes:
- Version of the equivalent inverse cipher with key expansion interleaved with cipher rounds
- Proof that this definition is equivalent to the original spec with precomputed keys (the proof actually is in terms of the keys from the forward direction)
- Improvements/more lemmas added to `fold_left_accumulate` section in `ListUtils.v`

 The proof of equivalence with the precomputed-keys is a bit more complex than it was for the forward direction, primarily because, instead of saying that the keys used by the precomputed version are generated by starting with a particular key (the first decryption key/last encryption key) and running inverse key expansion, I'm saying that they are generated by starting with the first _encryption_ key, running the cipher forward, and then reversing the list. This should help later proofs because the precondition will be the same as the forward version. However, it might end up being a better idea to simplify the proof by using the reverse-key-expansion precondition, and then using `all_keys_inv_eq` to prove that the preconditions are equivalent.

CC @satnam6502

### DIFF
--- a/cava/Cava/ListUtils.v
+++ b/cava/Cava/ListUtils.v
@@ -21,6 +21,13 @@ Hint Rewrite @nil_length @cons_length @seq_length @repeat_length @rev_length
 Ltac length_hammer :=
   autorewrite with push_length; eauto; lia.
 
+(* Miscellaneous proofs about lists *)
+Section Misc.
+  Lemma rev_seq_S start len :
+    rev (seq start (S len)) = (start + len) :: rev (seq start len).
+  Proof. rewrite seq_S, rev_app_distr; reflexivity. Qed.
+End Misc.
+
 (* Proofs about fold_right and fold_left *)
 Section Folds.
   Lemma fold_left_nil {A B} (f : B -> A -> B) b :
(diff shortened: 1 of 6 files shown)


silveroak-opentitan/aes/Spec/InterleavedInverseCipher.v
```diff
@@ -0,0 +1,199 @@
+(* Copyright 2020 The Project Oak Authors                                   *)
```
blaxill, line 1: minor nit: there should be an extra `(****************)` line at the top here, it looks like the existing spec files also miss it

blaxill (approved):
Lgtm!

cava/Cava/ListUtils.v
```diff
@@ -148,6 +155,28 @@ Section FoldLeftAccumulate.
 
   Hint Rewrite @fold_left_accumulate'_cons : push_fold_acc.
 
+  Lemma fold_left_accumulate'_snd {A B C}
```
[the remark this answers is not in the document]
jadephilipoom: This has long been a bit of a [pain point](https://github.com/coq/coq/issues/9113) -- Coq devs are (my perception) chronically undermanned and usually focused on parts of the code that are not the standard library. That's how we get coqutil and stdlib2 and coq-ext-lib. It's a bit annoying, but on the other hand there are so many standard-library-level lemmas that could be merged that it would probably result in an absolutely massive standard library if they were more willing to take contributions.

silveroak-opentitan/aes/Spec/InterleavedInverseCipher.v
jadephilipoom: Oops, thanks for the catch! Fixed now

PR #337, opened 2020-11-10
Move transpose_rev to VectorUtils and prove it is involutive

jadephilipoom:
This is also progress towards #334 

I had to prove several sublemmas for this about custom vector definitions like `snoc`/`unsnoc`, and I had to move a few existing proofs earlier in the file so I could put, for instance, `map2` sublemmas with other sublemmas about `map2`.


blaxill (approved):
Lgtm!

PR #339, opened 2020-11-11
AddRoundKey: clarify context parameters.

smore-lore:
Before, this haphazardly used 'nc' for num columns and 'nb' for num
bytes per column, but in order to avoid confusion, I am replacing these
with variable names from the actual standard.

nc - is actually Nb
nb - this is actually Nk * 32-bits, so calling it bits_per_column
     for now


smore-lore:
This is still wrong.

PR #340, opened 2020-11-11
AddRoundKey: clarify variable names

smore-lore:
Lines these up with the standard, also renames key to
key_schedule_round, since this is actually a round-specific piece of the
key schedule.

### DIFF
--- a/silveroak-opentitan/aes/Spec/AddRoundKey.v
+++ b/silveroak-opentitan/aes/Spec/AddRoundKey.v
@@ -26,7 +26,7 @@ Section Spec.
      For this standard, Nb = 4. *)
   Variable nb : nat.
 
-  Definition state := t word nb.
+  Definition state := Vector.t word nb.
   Definition key := word.
 
   (* FIPS 197, 5.1.4 AddRoundKey() Transformation *)
@@ -39,7 +39,7 @@ Section Spec.
       destruct b1, b2; reflexivity.
     Qed.
 
-    Lemma inverse_bvxor n (a : t bool n) (b : t bool n) :
+    Lemma inverse_bvxor n (a : Vector.t bool n) (b : Vector.t bool n) :
                         BVxor n (BVxor n a b) b = a.
     Proof.
       induction a.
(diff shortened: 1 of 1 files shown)


silveroak-opentitan/aes/Spec/AddRoundKey.v
```diff
@@ -17,12 +17,20 @@ Require Import Coq.Bool.Bvector Coq.Vectors.VectorDef.
...
+  (* FIPS 197, 2.1: Word A group of 32 bits that is treated either as a single
+     entity or as an array of 4 bytes. *)
+  Variable bits_per_word : nat.
+  Definition word := (Bvector bits_per_word).
+
+  (* FIPS 197, 2.2: Number of columns (32-bit words) comprising the State.
+     For this standard, Nb = 4. *)
+  Variable nb : nat.
+
+  Definition state := t word nb.
```
satnam6502, line 29: I would propose:
```coq
Vector.t word nb
```

satnam6502 (approved):
I know that `Nb` used in the AES spec document but `nb` does seem a bit cryptic. I wonder if `NumBytes` or some such thing might be better? `nb` is fine for now.

silveroak-opentitan/aes/Spec/AddRoundKey.v
smore-lore: Agreed - I added a commit doing this in the pull request.

smore-lore: note - specifically did not squash because it changes the whole file

jadephilipoom (approved):
LGTM!

The one comment I have is that I think it'd be okay to stick with the `key` name; I think FIPS talks about this round-specific part of the schedule as the "round key". But it's also fine with me to change it!

smore-lore:
> The one comment I have is that I think it'd be okay to stick with the `key` name; I think FIPS talks about this round-specific part of the schedule as the "round key". But it's also fine with me to change it!

You are right, from 5.1.4, for instance:

> Each Round Key consists of Nb words from the key schedule (described in Sec. 5.2).

And I'd also like to make it consistent between Cipher.v and here, eventually. Let me avoid changing it here for now.

PR #341, opened 2020-11-12
Use anonymous modules instead of plain kappa

blaxill:
_(Updated)_ smaller than it looks, some churn on expression type definitions, sorry about that ...


blaxill:
Currently messy, I will clean up before asking for a review proper.

I'd like to make this change as it simplifies manipulation of Expr AST and also how combinators can be defined (there is no differentiation between a module and a fragment). Unfortunately this requires more rewriting in prove_Wf/kappa_spec, @jadephilipoom perhaps there is a better way? Rewriting seems to be required now as previously everything was bare `Kappa` and so "calls" didnt need particularly special handling.

cava/Cava/Arrow/CombinatorProperties.v
```diff
@@ -147,19 +147,32 @@ Ltac kequiv_step :=
...
-    eapply RemoveContext_equiv
-  | |- kappa_equivalence _ (CallModule _) (CallModule _) =>
-    eapply CallModule_equiv
+  | |- kappa_equivalence _
+      (CallModule ?F)
+      (CallModule ?F) =>
+    eapply CallModule_equiv;
+    match goal with
+    | H: Wf (?F) |- _ => apply H
+    | _ => idtac (* "No hypotheses matching" F *)
```
jadephilipoom, line 156: Why comment out the error message?

cava/Cava/Arrow/CombinatorProperties.v
```diff
@@ -147,19 +147,32 @@ Ltac kequiv_step :=
...
-  | |- kappa_equivalence _ (RemoveContext _) (RemoveContext _) =>
-    eapply RemoveContext_equiv
-  | |- kappa_equivalence _ (CallModule _) (CallModule _) =>
-    eapply CallModule_equiv
+  | |- kappa_equivalence _
+      (CallModule ?F)
+      (CallModule ?F) =>
+    eapply CallModule_equiv;
+    match goal with
+    | H: Wf (?F) |- _ => apply H
```
jadephilipoom, line 155:
```suggestion
    lazymatch goal with
    | H: Wf  (mkModule F) |- _ => apply H
```

This way you'll get to see the error message from `apply` if you find the right `Wf` in context but the `apply` fails.

cava/Cava/Arrow/CombinatorProperties.v
```diff
@@ -147,19 +147,32 @@ Ltac kequiv_step :=
...
+  | H: Wf ?X, H1: Wf ?X -> _ |- _ => specialize (H1 H)
   | |- kappa_equivalence _ _ _ =>
     first [ kequiv_step
           | solve [eauto with Wf] ]
   | |- List.In _ _ => cbn [List.In]; tauto
   end.
-Ltac prove_Wf := cbv [Wf module_body]; intros; repeat prove_Wf_step.
+
+Ltac prove_Wf :=
+  intros; cbv [Wf module_body module_to_expr]; intros; repeat prove_Wf_step.
```
jadephilipoom, line 171:
```suggestion
  intros; cbn [Wf module_body module_to_expr]; intros; repeat prove_Wf_step.
```
If the slowdown you mentioned is due to too much getting unfolded, particularly matches on called modules, then switching to `cbn` here might help.

cava/Cava/Arrow/DeriveSpec.v
```diff
@@ -21,8 +21,12 @@ Require Import Cava.Arrow.ArrowExport.
 (* This file contains tactics and notations designed to simplify proofs that
    derive or prove specifications for kappa-level circuits. *)
 
-(* convenient notation *)
-Notation kinterp x := (interp_combinational' (x coq_func)).
+Lemma reduce_module_match i o (f: ModuleK i o):
+  interp_combinational' ((f: Kappa _ _) coq_func) =
```
jadephilipoom, line 25: Ooh, I don't know this syntax (the `(f:Kappa _ _)` part). What does it do?

silveroak-opentitan/aes/Arrow/NaiveCipherProperties.v
```diff
@@ -69,6 +73,7 @@ Section Equivalence.
        @aes_shift_rows_correct @aes_mix_columns_correct : kappa_interp.
   Opaque aes_256_naive_key_expansion mix_columns.aes_mix_columns.
 
+
```
jadephilipoom, line 76: nit:extra line

silveroak-opentitan/aes/Arrow/NaiveCipherProperties.v
```diff
@@ -79,6 +84,7 @@ Section Equivalence.
          As unrolled_cipher_naive'_correct.
   Proof.
     cbv [unrolled_cipher_naive']; kappa_spec.
+
```
jadephilipoom, line 87: nit:extra line

jadephilipoom (approved):
LGTM. Looking at the build history, it seems like this introduces a ~30s slowdown, which is not ideal but is also not big enough, I think, to block merging.

cava/Cava/Arrow/CombinatorProperties.v
blaxill: Its relatively noisy- it gets emitted before trying the hints in Wf. Its useful to debug if hypothesis are correctly matching though.

cava/Cava/Arrow/DeriveSpec.v
blaxill: I'm just forcing the coercion to happen before applying `coq_func`, as `coq_func` cannot be applied to `ModuleK`

blaxill:
> If the slowdown you mentioned is due to too much getting unfolded, particularly matches on called modules, then switching to cbn here might help.

`cbv [Wf]` is currently needed, to progress. Any changes to increase performance would be appreciated, but I think I'm getting diminishing returns with my skills :)

PR #342, opened 2020-11-17
Major overhaul of Acorn/Monad kernel

satnam6502:
This is a major overhaul of how the circuit interface descriptions work for the Acorn monad-kernel for Cava. Changes include:

* A different representation for the circuit's input and output ports, which are now expressed with a *list* of port declarations. The corresponding circuit works over inputs and outputs that correspond to a flat _n_-tuple that matches the length port specification _n_-list.
* The types for netlist generation, testbench generation, hierarchical instantiation and black box components are impact as a result.
* The type changes do away with vector smashing and all top-level circuits.

### DIFF
--- a/arrow-examples/Mux2_1.v
+++ b/arrow-examples/Mux2_1.v
@@ -28,8 +28,9 @@ Section notation.
   Local Open Scope kind_scope.
 
   Definition mux2_1
-    : << ArrowKind.Bit, ArrowKind.Bit, ArrowKind.Bit, Unit >> ~> ArrowKind.Bit :=
-    <[ \ sel a b =>
+    : << Bit, << Bit, Bit >>, Unit >> ~> Bit :=
+    <[ \ sel ab =>
+      let '(a,b) = ab in
       let sel_a = and sel a in
       let inv_sel = not sel in
       let sel_b = and inv_sel b in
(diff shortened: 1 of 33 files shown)


satnam6502:
@blaxill please can you add a commit to fixup the arrow examples so they work with the new netlist layer? Thank you kindly.

arrow-examples/Mux2_1.v
```diff
@@ -27,8 +27,10 @@ Section notation.
   Local Open Scope category_scope.
   Local Open Scope kind_scope.
 
+About Kind.
```
jadephilipoom, line 30: nit: remove `About`

jadephilipoom (approved):
LGTM!

arrow-examples/Mux2_1.v
satnam6502: Done, thanks.

cava/Cava/Signal.v
```diff
@@ -15,44 +15,147 @@
...
 
 From Cava Require Import Kind.
 From Cava Require Import VectorUtils.
 
-Inductive Signal : Kind -> Type :=
+(******************************************************************************)
+(* The types of signals that can flow over wires, used to index signal        *)
+(******************************************************************************)
+
+Inductive SignalType :=
```
blaxill, line 30: This definition is the same as the existing `Kind` in `cava/Kind.v`, one of them could be removed (I don't mind which)

satnam6502: I plan several sweeps of cleanup and code removal/refactoring after this PR lands.

satnam6502: I would vote for removing `Kind` because `SiganlType` is a more focused and specific name, and I want to avoid confusion with other kinds of kind e.g. kinds in Haskell.

blaxill (approved):

PR #343, opened 2020-11-17
Inverse cipher equivalence (arrow-based)

jadephilipoom:
Resolves the arrow-based part of #334 

- **Proof that the arrow-based inverse cipher implementation matches the inverse cipher spec**
- A few additional proofs about arrow combinators that I needed to do the proof
- Refactoring of some common structures from forward/inverse equivalence proofs into a separate, mutually imported file
- Extra lemmas about `all_keys` that aren't currently used but might be helpful in the future

I was really spinning my wheels on this one for a while, mostly because I was trying to write the equivalence proof by only providing an expression for the full list of forward keys, and then writing the precomputed keys for the inverse function as the reverse of this list. However, that makes the proof much trickier, because on top of reasoning about the equivalence between the two different functions you're also dealing with the fact that one side gets the next key by doing inverse key expansion and the other gets the next key by stepping backwards in the forward key list. You *can* prove equivalence between these in various ways, but dealing with that while also dealing with the equivalence proof gets tricky because there are parts of the structure you really don't want to inline and transform.

I ended up getting unstuck by providing an expression for the full list of *inverse* keys, and then also requiring that the key that the inverse key expansion list starts with is the same as the last element of the forward list. This precondition should be pretty reasonable to prove -- or even just compute, if we instantiate the key expansion spec.

CC @satnam6502

### DIFF
--- a/silveroak-opentitan/aes/Arrow/OpenTitanInverseCipherEquivalence.v
+++ b/silveroak-opentitan/aes/Arrow/OpenTitanInverseCipherEquivalence.v
@@ -37,13 +37,28 @@ Import VectorNotations ListNotations.
 Local Notation nat_to_bitvec size n := (Ndigits.N2Bv_sized size (N.of_nat n)).
 Local Notation nat_to_byte n := (nat_to_bitvec 8 n).
 Local Notation byte := (Vector.t bool 8) (only parsing).
+Local Notation state := (Vector.t (Vector.t byte 4) 4) (only parsing).
+Local Notation key := (Vector.t (Vector.t byte 4) 4) (only parsing).
+Local Notation rconst := byte (only parsing).
+Local Notation keypair := (Vector.t (Vector.t byte 4) 8) (only parsing).
+
+(* aes_key_expand operates on a pair of keys; this is a sliding window of
+   (previous key ++ current key). The following definitions project out
+   keys from the pair. *)
+Definition fstkey : keypair -> key :=
+  @slice_by_position
+    (t (t bool 8) 4) 8 3 0 (kind_default (Vector (Vector Bit 8) 4)).
+Definition sndkey : keypair -> key :=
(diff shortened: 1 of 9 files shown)


silveroak-opentitan/aes/Arrow/OpenTitanInverseCipherEquivalence.v
```diff
@@ -0,0 +1,238 @@
...
+        (Nr:=14) (init_rcon := nat_to_byte 1) (final_rcon:=nat_to_bitvec _ 64)
+        init_keypair final_keypair first_key last_key middle_keys input :
+    (* key_expand state is rconst * keypair *)
+    let all_rcons_and_keypairs_fwd := all_keys key_expand Nr (init_rcon, swap_keys init_keypair) in
+    let all_rcons_and_keypairs_inv := all_keys inv_key_expand Nr (final_rcon, swap_keys final_keypair) in
+    (* representation change: project out the relevant part of the key and transpose it *)
+    let all_keys_fwd := List.map projkey2 all_rcons_and_keypairs_fwd in
+    let all_keys_inv := List.map projkey1 all_rcons_and_keypairs_inv in
+    all_keys_inv = (first_key :: middle_keys ++ [last_key])%list ->
+    (exists keys, all_rcons_and_keypairs_fwd = (keys ++ [(final_rcon, final_keypair)])%list) ->
```
jadephilipoom, line 162: This is the precondition I mentioned in the description

blaxill (approved):
Lgtm! I was trying to see if you can abstract some tactics, but the usages look quite unique

jadephilipoom:
There's definitely more I can do in terms of automation, but I don't want to specialize too hard on the structure we have right now while everything is still in flux. Once we do some more equivalence proofs, I'll try to find common patterns and make things more automated!

Issue #344, opened 2020-11-17
Make AES spec testable


jadephilipoom:
Summary from discussion at weekly meeting: As the AES subroutine specifications are written/proven, it would be nice to have the ability to run the specs against some tests. This might be useful for developing the AES implementations also; both the Acorn and Arrow expressions can be interpreted as Gallina functions. It would be a good source of quick feedback if there was a way to take an AES that does pass the test and swap out the implementation of a subroutine for (the Gallina representation of) your own implementation, then see if it still passes. And of course, having checks on our spec is always good.

I think the best way to do this is to:
- [x] Write a simple, not-necessarily-nice-to-prove version of each subroutine we don't yet have a spec for (`shift_rows`, `mix_columns`, `sub_bytes`, `key_expand`, and their inverses) -- these can be replaced later
- [ ] Create a file that has all the test vectors
- [ ] Implement functions `aes_cipher_step` and `aes_inverse_cipher_step` that, given Gallina implementations of each subroutine, a round index, and state information, runs one round of AES and returns state information
- [ ] Implement a function `aes_test` that, given a Gallina implementation that runs AES one round at a time (i.e. the one produced by the `_step` functions, but leaving enough flexibility to accept some other custom implementation) and a test vector, checks that the implementation matches expected values on every round
- [ ] Run this function on all the specs/placeholders with all the test vectors and make sure it passes (in a way such that the CI will fail if it doesn't)

To test some definition of e.g. `sub_bytes`, you would run `aes_test` on `aes_cipher_step` with every subroutine *except* `sub_bytes` instantiated with the original spec/placeholder version, and with your own `sub_bytes`. If the test fails, then you know something is off with the implementation.

CC @smore-lore @cipher1024

PR #345, opened 2020-11-17
Make add_round_key use `nb`-word key

jadephilipoom:
I was trying to prove equivalence of the Acorn-based add-round-key with the spec, and realized that there was a minor error; the spec was xor-ing each word of the state with a one-word key, while it should be xor-ing *corresponding* words from the state and an `nb`-word key. The change is very small, so kudos to @smore-lore for a nice and flexible design!

Definitely a point in favor of #344

### DIFF
--- a/silveroak-opentitan/aes/Spec/AddRoundKey.v
+++ b/silveroak-opentitan/aes/Spec/AddRoundKey.v
@@ -27,11 +27,11 @@ Section Spec.
   Variable nb : nat.
 
   Definition state := Vector.t word nb.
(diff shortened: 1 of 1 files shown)


smore-lore:
Hmmm, yeah, thanks, not sure how I missed this now.

> Each Round Key consists of Nb words from the key schedule (described in Sec. 5.2).

This definitely shows the importance of having somewhere to make sure all the types are consistent as well. Thanks again!

smore-lore (approved):

PR #346, opened 2020-11-17
Acorn add round key equivalence

jadephilipoom:
Before doing the Acorn-based inverse cipher equivalence proof for #334, I did some cleanup of the forward proof. I separated out the add-round-key/xor part, and used our brand-new add-round-key spec to prove equivalence for that part of the cipher.

### DIFF
--- a/silveroak-opentitan/aes/Acorn/AddRoundKeyEquivalence.v
+++ b/silveroak-opentitan/aes/Acorn/AddRoundKeyEquivalence.v
@@ -30,9 +30,9 @@ Require Import AcornAes.AddRoundKey.
 Existing Instance Combinational.
 
 Section Equivalence.
-  Local Notation byte := (t bool 8).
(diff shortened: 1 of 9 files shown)


silveroak-opentitan/aes/Acorn/AddRoundKeyEquivalence.v
```diff
@@ -0,0 +1,50 @@
...
+Require Import Cava.Acorn.Lib.AcornVectors.
+Import VectorNotations.
+
+Require Import AesSpec.AddRoundKey.
+Require Import AcornAes.AddRoundKey.
+
+Existing Instance Combinational.
+
+Section Equivalence.
+  Local Notation byte := (t bool 8).
```
satnam6502, line 33: Perhaps it is better to write these out as qualified types to make them easier to read i.e. `Vector.t` ?

satnam6502 (approved):

PR #348, opened 2020-11-19
Placeholder operations

jadephilipoom:
Progress towards #334 

This PR adds placeholder, runnable versions of `sub_bytes`, `shift_rows`, and `mix_columns`, and their inverses. Each one has some minimal tests, although of course the real test will be to run all of them with the full test vectors. Currently, they are unused.

Note: about 500 lines of diff here are the s-boxes, which I wrote by copying the tables from FIPS and using a python script to convert them into 256-line match statements on Coq's native bytes. Not the most readable -- I think we should definitely use a different strategy long-term so it's easier to verify by eye that the s-box says what it's supposed to. Based on email discussion, perhaps @smore-lore might be interested?

@cipher1024 would you mind taking a look at the `mix_columns` placeholder? Because even the byte-level multiplication is multiplication modulo a polynomial in a polynomial field, I ended up writing some minimal infrastructure for polynomials based mostly on the tricks I remember using for fiat-crypto. If we can get an algebra library that can prove these operations form a field, that would be great! Of course, another representation is also fine, but this might be a good starting point/testable example.

Next up: key expansion placeholder version, infrastructure to pull in all the placeholders/specs and run test vectors on them.


smore-lore:
Yeah, I would like to try getting everything hooked together w/ the data flow through the whole algorithm. I'm not sure how hard it will be to prove isomorphisms between all the different state representations, but having that flow will at least test out writing those transformations.

blaxill (approved):
Lgtm. Minor: I'm not sure about putting it under `Placeholder`, this isn't intended to be removed right? If its distinctly not intended to be part of the spec, perhaps we could have something like

- `aes/Spec`
- `aes/Gallina` (instead of spec/placeholder)
- `aes/Monad`
- `aes/Arrow`

If it is part of the spec then just directly `aes/Spec` seems fine too

jadephilipoom:
> I'm not sure about putting it under `Placeholder`, this isn't intended to be removed right?

Actually, it sort of is. The difference between this and the definitions directly under `Spec` is that these are not necessarily nice to do proofs with -- I haven't done proofs for them, and there might well be different representations that we need to switch to once we start trying to do proofs. However, the focus here was just to get something runnable so we can achieve the fully runnable, testable spec version for debugging purposes -- that's why it's `Placeholder`. Does that make sense?

jadephilipoom:
> Yeah, I would like to try getting everything hooked together w/ the data flow through the whole algorithm. I'm not sure how hard it will be to prove isomorphisms between all the different state representations, but having that flow will at least test out writing those transformations.

I agree -- that's the next thing to do! I just need to write key expansion, and then I can hook everything up into a single definition. But this seemed like a large enough diff that I wanted to merge it first.

jadephilipoom:
Oops, put the wrong issue number in the initial description -- just for archival purposes, the actual issue this is progress towards is #344.

PR #351, opened 2020-11-20
Add generation of interface tuple default values

satnam6502:
Needed for blackBox module instantiation to become an overloaded operation, with the combinational and sequential instantiations yielding default values for the behaviour of an unspecified external block (as currently used in the pinmux).


satnam6502:
Self-merging minor change.

PR #352, opened 2020-11-20
Port AES Acorn top-level to monad version

satnam6502:
This PR re-wires the AES top-level blocks in the silveroak directory to use the mainline monad API which allows us to both simulate (i.e. verify) circuits and generate circuit implementations. As part of this PR the work done in the Acorn prototype is absorbed into the main Cava infrastructure. Addresses #364


satnam6502:
@jadephilipoom this PR ports the results of the experiments we did in the Acorn directory to rework the Cava typeclass to the mainline Cava monad implementation.

cava/Cava/Lib/BitVectorOps.v
```diff
@@ -14,49 +14,28 @@
...
-From Coq Require Import btauto.Btauto.
+From Coq Require Import Vectors.Vector.
+Import VectorNotations.
+Local Open Scope vector_scope.
 
 Require Import ExtLib.Structures.Monads.
-
-Local Open Scope list_scope.
-Local Open Scope monad_scope.
+Require Import ExtLib.Structures.Traversable.
```
jadephilipoom, line 22: Import possibly unused?

jadephilipoom (approved):
LGTM!

cava/Cava/Lib/BitVectorOps.v
satnam6502: It's needed for `mapT`.

PR #354, opened 2020-11-23
Testable spec for AES

### LINKED ISSUE #344: Make AES spec testable

jadephilipoom:
Resolves #344 

- Fixes a transcription error in the forward s-box (this is what I get for using a hacky python script to parse the FIPS table)
- Adds infrastructure for creating and processing tests with lots of intermediate results
- Copies over the two (forward and inverse) AES-256 tests from FIPS section C.3
- Strings together all the placeholder/spec definitions and tests them so if we break the spec we'll know (they work!)

To test something you're actively working on, you can import `AesSpec.Tests.CipherTest` and write something like:
```
Compute (aes_test_encrypt false (fun (step : AESStep) key =>
                                   match step with
                                   | ShiftRows => my_shift_rows
                                   | _ => aes_impl step key
                                   end)).
```
If `my_shift_rows` is correct, this prints `Success`. Otherwise, it prints an error. For instance, if `my_shift_rows` is the identity function, you get:
```
Error in round 1:
  after ShiftRows with (state=0x63cab7040953d051cd60e0e7ba70e18c),
    expected 0x6353e08c0960e104cd70b751bacad0e7
    got      0x63cab7040953d051cd60e0e7ba70e18c

Error in round 2:
  after ShiftRows with (state=0x84fb386f1ae1ac97df5cfd237c49946b),
    expected 0x84e1fd6b1a5c946fdf4938977cfbac23
    got      0x84fb386f1ae1ac97df5cfd237c49946b

Error in round 3:
  after ShiftRows with (state=0xadcb0f257e9c63e0bc557e951c15ef01),
    expected 0xad9c7e017e55ef25bc150fe01ccb6395
    got      0xadcb0f257e9c63e0bc557e951c15ef01
[...]
```
Note that the tests don't propagate errors; each step is tested individually, so if you have a problem in `shift_rows` it doesn't mess up future steps.

The `false` flag passed to `aes_test_encrypt` is for how to print the state in error messages; set it to `true` if you want them printed as a matrix, which is more verbose but can be very helpful:
```
Error in round 1:
  after ShiftRows with (state=
0x63 | 0x9 | 0xcd | 0xba
0xca | 0x53 | 0x60 | 0x70
0xb7 | 0xd0 | 0xe0 | 0xe1
0x4 | 0x51 | 0xe7 | 0x8c),
    expected 
0x63 | 0x9 | 0xcd | 0xba
0x53 | 0x60 | 0x70 | 0xca
0xe0 | 0xe1 | 0xb7 | 0xd0
0x8c | 0x4 | 0x51 | 0xe7
    got      
0x63 | 0x9 | 0xcd | 0xba
0xca | 0x53 | 0x60 | 0x70
0xb7 | 0xd0 | 0xe0 | 0xe1
0x4 | 0x51 | 0xe7 | 0x8c
0x6b | 0x6f | 0x97 | 0x23
```
You can also use the slightly lower-level `run_test` definition to run your test on a different test vector, or with a custom print function.

@cipher1024 @smore-lore I requested your review, but there's no need to do a code review (I know your time is limited for this project) -- rather, I'd like you to look at the description and tell me if this meets your needs, or if there's any other helpful thing I should add!


silveroak-opentitan/aes/Spec/Tests/CipherTest.v
```diff
@@ -122,11 +134,21 @@ Definition print_matrix (st : state) : string :=
                  rows in
   newline ++ String.concat newline (to_list lines).
 
-(* Run AES test; if there are errors, print state/keys as hex *)
-Goal (run_test state_eqb Bv2Hex Bv2Hex fips_c3_forward aes_impl = Success).
+(* Shortcut/convenience definitions *)
+Definition aes_test_encrypt
+           (print_state_as_matrix : bool)
```
smore-lore, line 139: Pref custom `Inductive Format : Set := Hex | Matrix` so semantics are encoded in caller visibly?

smore-lore (approved):
The output looks great. Perfect!

silveroak-opentitan/aes/Spec/Tests/Common.v
```diff
@@ -0,0 +1,164 @@
...
+                                              state_before_step := st;
+                                              state_after_step := st' |} in
+                                  (st', (acc ++ [d])%list))
+                               exp (st,[])) in
+            let st' := List.last (List.map snd exp) st in
+            (S i, st', List.app acc exp'))
+         (combine test.(round_expected_states) test.(round_ksch))
+         (0%nat, test.(plaintext), [])).
+
+(* TODO: create a string-based Utils file? *)
```
blaxill, line 88: I don't suppose you are aware of any we could pull in as a dependency ?

blaxill (approved):
LGTM when CI linker error is fixed (should be simple). The error likely means we need to add one of the `Extracted` haskell modules to `Cava2HDL.cabal` (error line is):

~~~
/builder/home/.cabal/lib/x86_64-linux-ghc-8.6.5/Cava2HDL-0.1.0.0-5PRQfZiqRpHDWcDIajF6Gy/libHSCava2HDL-0.1.0.0-5PRQfZiqRpHDWcDIajF6Gy.a(BitArithmetic.o)(.data+0x578): error: undefined reference to 'Cava2HDLzm0zi1zi0zi0zm5PRQfZZiqRpHDWcDIajF6Gy_Byte0_zdwtozuN_closure'
~~~

Your testing utilities are much more thorough than anything I've added so I've also opened #356

silveroak-opentitan/aes/Spec/Tests/CipherTest.v
```diff
@@ -0,0 +1,154 @@
...
+Require Import AesSpec.Tests.TestVectors.
+Require Import AesSpec.Placeholder.MixColumns.
+Require Import AesSpec.Placeholder.Sbox.
+Require Import AesSpec.Placeholder.ShiftRows.
+Require Import AesSpec.Placeholder.SubBytes.
+Require Import AesSpec.AddRoundKey.
+Local Open Scope string_scope.
+
+Local Notation byte := Byte.byte (only parsing).
+Local Notation round_key := (Vector.t bool 128) (only parsing).
```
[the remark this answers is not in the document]
jadephilipoom: They're not all unique, but there's something like three or four different options -- for instance, this file represents bytes as Coq native bytes, but many other files have `Vector.t bool 8` for bytes, and some just have flat words with no byte separation. Maybe once the spec has stabilized a bit more, it'll turn out to only use a few configurations, and then we can make our notations more uniform, but in the meantime keeping it file-local is helpful because I can just glance up and know which representation exactly I'm using.

silveroak-opentitan/aes/Spec/Tests/Common.v
jadephilipoom: I looked into it a bit more and found that `coq-ext-lib` has the newline *character* defined, so I was able to replace `"101"%char` with `chr_newline` so at least it's not just a magic number. But it doesn't seem like either `coq-ext-lib` or `coqutil` have the string version defined, and I'm hesitant to pull a specifically string-focused dependency given we don't use strings very much.

silveroak-opentitan/aes/Spec/Tests/CipherTest.v
```diff
@@ -0,0 +1,154 @@
+(* Copyright 2020 The Project Oak Authors                                   *)
```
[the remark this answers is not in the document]
jadephilipoom: Darn, I thought I fixed those. Must be another one somewhere, or just a copy-paste error on my part. Thanks for the catch!

silveroak-opentitan/aes/Spec/Tests/CipherTest.v
jadephilipoom: Done!

jadephilipoom:
> LGTM when CI linker error is fixed (should be simple). The error likely means we need to add one of the `Extracted` haskell modules to `Cava2HDL.cabal`

This is "fixed", but not very satisfyingly. The issue seems to be that all of `BitArithmetic.v` is extracted, and I've added some definitions that use `Coq.Init.Byte`, so `Byte` needs to be added to `other-modules` in `Cava2HDL.cabal`. So far, so good -- I added it. But for some reason I also needed to add `Byte0` -- not sure where that's coming from. @blaxill, do you know how to resolve this?

jadephilipoom:
In the name of not holding things up too long, I'll merge this, but create an issue for the `Byte0` part.

blaxill:
> > LGTM when CI linker error is fixed (should be simple). The error likely means we need to add one of the `Extracted` haskell modules to `Cava2HDL.cabal`
> 
> This is "fixed", but not very satisfyingly. The issue seems to be that all of `BitArithmetic.v` is extracted, and I've added some definitions that use `Coq.Init.Byte`, so `Byte` needs to be added to `other-modules` in `Cava2HDL.cabal`. So far, so good -- I added it. But for some reason I also needed to add `Byte0` -- not sure where that's coming from. @blaxill, do you know how to resolve this?

This is because we are now using definitions from `Coq.Strings.Byte` and `Coq.Init.Byte` in extracted files and they both extract to `Byte.hs`.

PR #355, opened 2020-11-24
Add Kappa based sequential semantics

blaxill:
There are existing arrow sequential semantics, but this adds list based
(not single time step) kappa semantics, and shows they are at least
equivalent on the counter example for a particular input.

There is an inefficiency in simulation of `Letrec`, I think this will
crop up with `loopBit` too. There may be a performant simple way to
express the evaluation I may have missed. Writing a more complex
performant version I will leave till later.

Also adds 'SignalingCounter' to help explore sequential semantic proofs


arrow-examples/SignalingCounter.v
```diff
@@ -0,0 +1,70 @@
...
+Definition signaling_counter_Interface :=
+   sequentialInterface "signaling_counter" "clk" PositiveEdge "rst" PositiveEdge
+     [mkPort "is_valid" Signal.Bit
+     ;mkPort "value" (Signal.Vec Signal.Bit 4)]
+     [mkPort "count" (Signal.Vec Signal.Bit 4)] [].
+
+Definition signaling_counter_netlist :=
+  build_netlist (closure_conversion counter) "signaling_counter" ("is_valid", "value") "count".
+
+(* first for cycles have 'is_valid = false' and should not increment the counter *)
```
jadephilipoom, line 55: nit: typo
```suggestion
(* first four cycles have 'is_valid = false' and should not increment the counter *)
```

jadephilipoom (approved):
LGTM. Is this ready to try proofs about the counter, or should I hold off for now?

Issue #320, opened 2020-10-30
Make Acorn implementation of sub_bytes


satnam6502:

satnam6502:
Do LUT-based S-box first.

PR #359, opened 2020-11-24
Acorn and Arrow tests

### LINKED ISSUE #356: Reuse Spec AES Test Routines for Arrow / Monad
Jade has added much more thorough testing for AES which we might be able to reuse for monad/arrow constructions (likely this will be easier for monad than arrow). Of course this is not a blocker.

jadephilipoom:
Resolves #356 

I went ahead and took a stab at using the new testing infrastructure for Acorn/Arrow expressions as @blaxill suggested, and it just works! Makes me very happy.

Currently, all the Arrow subroutines except for `add_round_key` and `key_expand` are tested. `add_round_key` not tested because it has no separate definition; it's just an inline xor in `cipher_round`. The `key_expand` routine is not tested because we don't have tests for it -- I could create key expansion tests, but I'm not sure if it's a top priority. @blaxill , thoughts?

The only Acorn subroutine currently tested is `add_round_key`, because it's the only one we currently have -- but hopefully the test can be used to help develop the other routines in #320, #321, and #322.

Each of the tests takes 1-2 seconds. They would be slightly faster (about 0.5-1s) with `native_compute`, but the CI doesn't like it because "the ocaml compiler is not installed" -- not sure what's going on there, but shaving off a couple of seconds didn't seem worth examining it too far.

### DIFF
--- a/silveroak-opentitan/aes/Acorn/AddRoundKey.v
+++ b/silveroak-opentitan/aes/Acorn/AddRoundKey.v
@@ -22,6 +22,8 @@ From Cava Require Import VectorUtils.
 From Cava Require Import Acorn.Acorn.
 From Cava Require Import Acorn.Lib.AcornVectors.
 From AcornAes Require Import Common.
+From AesSpec Require Import Tests.CipherTest.
+From AesSpec Require Import Tests.Common.
 Import Common.Notations.
 
 Section WithCava.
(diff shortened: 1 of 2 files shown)


blaxill (approved):
Lgtm!

>  I could create key expansion tests, but I'm not sure if it's a top priority. @blaxill , thoughts?

Yeah not a priority for now.

PR #361, opened 2020-11-27
Trying out some sequential proofs

blaxill:
Testing out single step semantics (semantics on list exists already).

@jadephilipoom :
- The structured state is perhaps less useful as (without attempting to reduce the structure) its large with lots of `unit` values. A state based on lists might be more accessible when writing proofs.
- I added a small example lemma `state_increases_on_input` in the `signalingcounter` without finishing the proof. I wonder if providing some invariant in gallina that you can compute might be cleaner (or any other ideas you have!)

### DIFF
--- a/arrow-examples/SignalingCounter.v
+++ b/arrow-examples/SignalingCounter.v
@@ -17,6 +17,8 @@
 
 From Cava Require Import Arrow.ArrowExport.
 From Coq Require Import Lists.List NArith.NArith Strings.String.
+
+Require Import Cava.Tactics.
 Import ListNotations.
 
 Local Open Scope string_scope.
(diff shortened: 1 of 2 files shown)


arrow-examples/SignalingCounter.v
```diff
@@ -59,11 +61,60 @@ Definition signaling_counter_tb_inputs : list (bool * Bvector.Bvector 4) :=
...
+Lemma expr_single_step_counter_semantics_agree:
+  (map Bv2N signaling_counter_tb_expected_outputs) =
+  (map Bv2N (unroll_sequential_step first_order_counter signaling_counter_tb_inputs)).
+Proof. vm_compute; reflexivity. Qed.
+
+(* perhaps the structured state is actually not useful, as you then need to find
+* it in the tupled structure ... *)
+Definition proj_state_counter (state: skappa_state first_order_counter)
+  : Bvector.Bvector 4
+  := fst state.
```
blaxill, line 82: luckily in this trivial example, the state variable we care about is at the outermost layer

arrow-examples/SignalingCounter.v
```diff
@@ -59,11 +61,60 @@ Definition signaling_counter_tb_inputs : list (bool * Bvector.Bvector 4) :=
...
+  cbn [proj_state_counter fst snd].
+
+  (* next steps ?????
+    context [(sequential_step
+              match mux_item with
+              ...]
+    is equal to some gallina function effectively (+1)
+    which should reduce to prove the goal
+    *)
+Admitted.
```
blaxill, line 116: If there were lemmas tying existing derived gallina from the combinational interpretation to `sequential_step` (which should hold for any valid for any combinational circuit), there should be some rewrite to solve this goal.

cava/Cava/Arrow/ExprSemantics.v
```diff
@@ -110,6 +117,193 @@ Section combinational_semantics.
     (i: list_func (remove_rightmost_unit x)): (list_func y) :=
     interp_sequential' expr (map (denote_apply_rightmost_tt x) i).
 
+  Inductive skappa: Kind -> Kind -> Type :=
```
jadephilipoom, line 120: Why do we need to redefine `kappa`?

jadephilipoom (commented):
I'm a bit surprised that this was such a big structural change -- what prevents it from being just another interpretation function on regular `kappa`?

Do you think we can tie together these semantics and multi-step semantics?

cava/Cava/Arrow/ExprSemantics.v
blaxill: I was originally experimenting with a first order version that contained the state itself, but it was trickier than I expected so stopped. `skappa` could be removed.

blaxill:
> Do you think we can tie together these semantics and multi-step semantics?

I can change this to a simpler representation, and it should be possible to move between single/multi-step, but the conversion of the letrec/loops might be difficult to prove correct (or perhaps not?)

satnam6502:
Likewise, close this since it seems old?

PR #362, opened 2020-11-30
Fix interp_combinational to not require matching/rewriting in derive

blaxill:


blaxill:
Hmm, decreases time for the derive instances in naivecipher/opentitan but looks slower overall

satnam6502:
Close this since it seems old?

PR #367, opened 2020-12-02
Sequential Count-by Puzzle

satnam6502:
This PR includes the implementation of the `countBy` circuit which acts as the first sequential circuit proof as described in #371.


cava/Cava/Monad/Sequential.v
```diff
@@ -0,0 +1,259 @@
...
+(* Sequential denotion of the SignalType and default values.                  *)
+(******************************************************************************)
+
+Fixpoint seqType (t: SignalType) : Type :=
+  match t with
+  | Void => list unit
+  | Bit => list bool
+  | Vec vt sz => (Vector.t (seqType vt) sz)
+  | ExternalType _ => list unit (* No semantics for sequential interpretation. *)
+  end.
```
jadephilipoom, line 49: I think it might be easier to write the match statements if you had the outer list exposed, i.e. `seqType t = list (combType t)`. That way you could actually peel off your first input without breaking down its type.

jadephilipoom: The crux of the problem is that you really want `loopSeq'` to be a fixpoint that decreases on `a`, but you don't know the type of `a` until you destruct the SignalType `A`. Another workaround would be to define `loopSeq'` separately for each type.

jadephilipoom:
I tried:

```coq
Definition seqType t := list (combType t).
Definition defaultSeqValue t := [defaultCombValue t].

Fixpoint loopSeq' {A B C : SignalType}
                  (f : combType A * combType C -> ident (combType B * combType C))
                  (a : seqType A) (feedback: combType C)
  : ident (seqType B) :=
  match a with
  | [] => ret []
  | x :: xs =>
    '(y, nextState) <- f (x, feedback) ;; (* One step of f. *)
    ys <- loopSeq' f xs nextState ;; (* remaining steps of f *)
    ret (y :: ys)
  end.
```
That lets me define `loopSeq`, but if I try to say that `loopBit` is a specialization of `loopSeq` I get a type mismatch:
```diff
- (combType a * combType b -> ident (combType Bit * combType b)) -> seqType a -> ident (seqType Bit)
+ (seqType a * seqType Bit -> ident (seqType b * seqType Bit)) -> seqType a -> ident (seqType b)
```
Some of that is just argument order, but I think the `loopBit` field generally expects the looped circuit to be sequential instead of combinational. Do we need to define loops that loop sequential circuits? If not, we could work around this (for instance, making a wrapper that turns a sequential circuit combinational by just peeling off the first input and then making a singleton list for the output), but if we do want to have loops/delays within loops I think there are some very interesting representation questions to consider.

For instance, if I have a loop whose body is, say, add the feedback to the input and then a delay (and then an extra delay before feeding back, as I understand we're doing with all loops):
```
        _______
    ---| delay |--------------
   |   |_______|             |
   |   ____      _______     |
    --| + | -- | delay | --------- out
 in --|___|    |_______| 
```

..and my input is a stream of 1s, then  (I think) the expected output of the loop is `[0,1,1,2,2,3,3,...]`, based on:
```
f[t+1] = out[t] # feedback
out[t+1] = in[t] + f[t] = 1 + f[t]
out[0] = f[0] = 0
```

| t  | 0 | 1 | 2 | 3 | 4 | 5 | 6 |
| -- | -- | -- | -- | -- | -- | -- | -- |
| f | 0 | 0 | 1 | 1 | 2 | 2 | 3 |
| out | 0 | 1 | 1 | 2 | 2 | 3 | 3 |

However, one iteration of the looped par will always give me `[0,f+in]`. Somehow I need to know to not ignore the `0` the first time, but ignore it the other times. What would be a good way to do that?

cava/Cava/Monad/Sequential.v
satnam6502: I actually did try `seqType t = list (combType t)` but I got the same error message.

satnam6502:
I ideally want the higher order circuit argument to be a sequential circuit, especially since that lets us build up state in a hierarchical manner. Here I made it a combinational circuit because I thought it would help with debugging. Perhaps that was a mistake.

satnam6502:
As for `loopBit`, my plan was the get rid of it once `loopSeq` was defined.

satnam6502:
I thought I did try to write pretty much this:
```coq
 Fixpoint loopSeq' {A B C : SignalType}
                  (f : combType A * combType C -> ident (combType B * combType C))
                  (a : seqType A) (feedback: combType C)
  : ident (seqType B) :=
  match a with
  | [] => ret []
  | x :: xs =>
    '(y, nextState) <- f (x, feedback) ;; (* One step of f. *)
    ys <- loopSeq' f xs nextState ;; (* remaining steps of f *)
    ret (y :: ys)
  end.
```
but I must have got something wrong.... thinking...

cava/Cava/Monad/Sequential.v
jadephilipoom: Were you still destructing `A`?

jadephilipoom (approved):
LGTM. I'll fix up the conflicts and merge, then get to work on the puzzle!

PR #368, opened 2020-12-02
Absolutize imports

jadephilipoom:
Partially addresses https://github.com/project-oak/oak-hardware/issues/338

Best reviewed commit-by-commit -- the first two add coq-tools dependency, the third adds a script `tools/absolutize-requires.sh` to call the coq-tools scripts on our repo safely, and the last commit is the huge change that results from running the script.

### DIFF
--- a/tools/absolutize-requires.sh
+++ b/tools/absolutize-requires.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+#
+# Copyright 2020 The Project Oak Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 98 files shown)


satnam6502 (approved):

tools/absolutize-requires.sh
```diff
@@ -0,0 +1,40 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+TOPLEVEL=$(git rev-parse --show-toplevel)
+
+# Use python2 because absolutize-imports.py is not executable
+ABSOLUTIZE="python2 $TOPLEVEL/third_party/coq-tools/absolutize-imports.py --in-place -v"
```
JasonGross, line 21: @jadephilipoom I've updated `absolutize-imports.py` to be executable, and to accept `--update-all -f _CoqProject` just like `minimize-requires.py`.  (`--update-all` instead of `--all` because `--all` has another meaning for `absolutize-imports.py`; I've modified `minimize-requires.py` to accept `--update-all` as a synonym of `--all`)  Let me know if you want any other features.

jadephilipoom: Cool, thanks for this! I'll update us to the latest version and try it out.

PR #370, opened 2020-12-02
SubBytes.v: proof of inverse property.

### LINKED ISSUE #309: Prove that inverse-sub-bytes implements inverse of sub-bytes

Once we have a specification for `sub_bytes` and `inv_sub_bytes`, we need to prove the usual inverse property: `forall x, inv_sub_bytes (sub_bytes x) = x`.

The `sub_bytes` operation is parameterized on an "s-box" which is essentially a lookup table that declares which bytes are substituted. From the OpenTitan [AES docs](https://docs.opentitan.org/hw/ip/aes/doc/):
_The AES unit currently uses a LUT-based S-Box implementation (default) but also supports the implementation proposed by [Canright: “A very compact Rijndael S-Box”](https://calhoun.nps.edu/handle/10945/25608) (selectable by a compile-time parameter)._

So we'll need to support at least those two s-boxes. Ideally, the proof would be parameterized over the s-box and proved in terms of some property of the s-box itself. Since there are only 256 possible entries, that property is likely to be easily computable with `vm_compute` or similar.

smore-lore:
This moves sub-bytes out of placeholder algorithms and proves the
inverse property for it.

closes #309


cava/Cava/VectorUtils.v
```diff
@@ -560,6 +560,15 @@ Section VectorFacts.
...
 
+  Lemma map_map A B C (n : nat) (l : Vector.t A n) (f : A -> B) (g : B -> C) :
+    map g (map f l) = map (fun x => g (f x)) l.
+  Proof.
+    induction l; [reflexivity|].
+    simpl.
+    rewrite IHl.
+    reflexivity.
+  Qed.
+
```
satnam6502, line 571: The Coq Vector standard library has:
```coq
Lemma map_map A B C: forall (f:A->B) (g:B->C) n (v : t A n),
  map g (map f v) = map (fun x => g (f x)) v.
```
https://coq.inria.fr/distrib/current/stdlib/Coq.Vectors.VectorSpec.html

smore-lore: I thought I had seen it before! thanks.

satnam6502 (approved):

satnam6502:
Thanks for using a few lemmas from the `Vector` library: good to reuse already proved things!

jadephilipoom (approved):
LGTM!

Issue #371, opened 2020-12-04
Sequential Proofs


satnam6502:
The purpose of this issue is to track work on the development of sequential proofs for Cava circuits.

The first proof we consider is for a synthetic example of a circuit made with a counter which counts-by the current input values:

```
        _______
    ---| delay |------------
   |   |_______|            |
   |   ___                  |
    --| + ---------------------- out
 in --|___|   
```

The countBy circuit takes the current 8-bit input (in) and the current state value represented by the register (delay) which is output in the current cycle as the output of the circuit, and this value is also the next state value for the delay. The adder is an 8-bit adder with no bit-growth i.e. it computes (a + b) mod 256.

Currently this circuit description is in the PR #367 in the file [tests/Delay.v](https://github.com/project-oak/oak-hardware/blob/0ed207672075912c20711e22d5c125f046799269/tests/Delay.v).

The implementation makes use of the `loop` combinator which implicitly places a register on the feedback path:
```coq
Section WithCava.
  Context {signal} `{Cava signal} `{Monad cava}.

  Definition countFork (i : signal (Vec Bit 8) * signal (Vec Bit 8))
                       : cava (signal (Vec Bit 8) * signal (Vec Bit 8)) :=
    newCount <- addN (fst i) (snd i) ;;
    ret (newCount, newCount).

  Definition countBy : signal (Vec Bit 8) -> cava (signal (Vec Bit 8))
    := loop countFork.

End WithCava.
```

We can test the circuit with some concrete values to confirm that it seems to work as intended:
```coq
Example countBy_ex1: sequential (countBy [b14; b7; b3; b250]) = [b14; b21; b24; b18].
Proof. reflexivity. Qed.
```

We can write a specification for the intended behaviour:
```coq
Fixpoint countBySpec' (state: N) (i : list (Bvector 8)) 
                      : list (Bvector 8) :=
  match i with
  | [] => []
  | x::xs => let newState := (Bv2N x + state) mod 256 in
             N2Bv_sized 8 newState :: countBySpec' newState xs 
  end.

Definition countBySpec := countBySpec' 0.
```

Now the challenge is to prove that the implementation `countBy` has equivalent behaviour to the specification `countBySpec`:
```coq
Lemma countByCorrect: forall (i : list (Bvector 8)),
                      sequential (countBy i) = countBySpec i.
```

[...]

PR #372, opened 2020-12-04
Count by proofs

jadephilipoom:
Progress on #371 

With a minor modification (using a bitvector as the state instead of an `N`), and depending on an admitted proof that `addN` is correct, I was able to take the counter and spec in #367 and prove that they correspond. I put in a little automation as well -- review commit-by-commit to see the proofs before I sprinkled tactic magic on them.

Next step is to prove the admitted `addN` lemma, but I thought it was worth getting feedback on this step first.

### DIFF
--- a/tests/Delay.v
+++ b/tests/Delay.v
@@ -22,9 +22,12 @@ Import ListNotations.
 Require Import ExtLib.Structures.Monads.
 Export MonadNotation.
 
+Require Import Cava.Acorn.Identity.
 Require Import Cava.Cava.
+Require Import Cava.Tactics.
 Require Import Cava.Monad.CavaMonad.
 Require Import Cava.Lib.UnsignedAdders.
+
 
 From Coq Require Vector.
 From Coq Require Import Bool.Bvector.
@@ -142,7 +145,7 @@ bit-growth i.e. it computes (a + b) mod 256.
 *)
 
 Section WithCava.
-  Context {signal} `{Cava signal} `{Monad cava}.
+  Context {signal} {semantics:Cava signal} `{Monad cava}.
 
   Definition countFork (i : signal (Vec Bit 8) * signal (Vec Bit 8))
                        : cava (signal (Vec Bit 8) * signal (Vec Bit 8)) :=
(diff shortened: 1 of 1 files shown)


tests/Delay.v
```diff
@@ -142,7 +145,7 @@ bit-growth i.e. it computes (a + b) mod 256.
 *)
 
 Section WithCava.
-  Context {signal} `{Cava signal} `{Monad cava}.
+  Context {signal} {semantics:Cava signal} `{Monad cava}.
```
jadephilipoom, line 148: I named this argument because in some places I pass it in explicitly, and that looks neater (and is less fragile) if the argument has a name.

tests/Delay.v
```diff
@@ -159,16 +162,62 @@ Proof. reflexivity. Qed.
...
+
+(* TODO: rename typeclass arguments *)
+Lemma addNCorrect n (a b : list (Bvector n)) :
+  sequential (addN (H:=SequentialCombSemantics) a b) = addNSpec a b.
+Admitted.
+Hint Rewrite addNCorrect using solve [eauto] : seqsimpl.
+
+Lemma countForkStep:
+  forall (i : Bvector 8) (s : Bvector 8),
+    sequential (countFork (semantics:=SequentialCombSemantics) ([i], [s]))
```
jadephilipoom, line 200: This is where I pass the instance explicitly; if the argument is auto-named, it would be `(H:=SequentialCombSemantics)`, and the `H` would change if the order of auto-named arguments changed.

tests/Delay.v
```diff
@@ -159,16 +162,62 @@ Proof. reflexivity. Qed.
...
-  | x::xs => let newState := (Bv2N x + state) mod 256 in
-             N2Bv_sized 8 newState :: countBySpec' newState xs 
+  | x::xs =>
+    let newState := N2Bv_sized 8 (Bv2N x + Bv2N state) in
+    newState :: countBySpec' newState xs
+  end.
+
+Definition countBySpec := countBySpec' (N2Bv_sized 8 0).
+
+(* TODO: addN sequential seems to only return one result; shouldn't it be a map2? *)
```
jadephilipoom, line 176: @satnam6502 I think this might be happening because `addN` has a constant zero, and that constant zero is instantiated with `[false]` as per `zero` from `SequentialCombSemantics`. So it only runs for one timestep. This doesn't affect the counter because we run `addN` one step at a time, but I suspect it might be a problem in the future. We really want an infinite sequence of zeroes that we can truncate to match the input, which is hard to encode.

satnam6502: Ah! I currently have an incorrect definition of `zero` for the sequential semantics. In the Haskell version `zero` is defined as an infinite list of zero values, which works fine with lazy evaluation. Here I need to know somehow "how many zeros" to generate.

tests/Delay.v
```diff
@@ -159,16 +162,57 @@ Proof. reflexivity. Qed.
...
+  match a,b with
+  | a :: _, b :: _ => [N2Bv_sized n ((Bv2N a + Bv2N b))]
+  | _,_ => []
   end.
 
-Definition countBySpec := countBySpec' 0.
+(* TODO: rename typeclass arguments *)
+Lemma addNCorrect n (a b : list (Bvector n)) :
+  sequential (addN (H:=SequentialCombSemantics) a b) = addNSpec a b.
+Admitted.
```
satnam6502, line 195: Since you've already done the proof for the ripple carry adder before, I think once we resolve the `zero` issue this should be easy?

satnam6502 (approved):

tests/Delay.v
satnam6502: I could have a go at porting the proofs. I think now the mainline monad kernel has absorbed all of what we used/developed for Acorn so it should be straight forward (famous last words).

PR #373, opened 2020-12-04
Remove unused imports

### LINKED ISSUE #338: Normalize imports automatically
See https://github.com/project-oak/oak-hardware/pull/330#issuecomment-721030565

In #330, I went through and fully-qualified every single import in our repo to remove undefined behavior and import-ambiguity warnings. But this is also painful to maintain. We should use existing tools, for instance from [coq-tools](https://github.com/JasonGross/coq-tools), to create a script that can just be run before submitting a PR. We could even add it as a git hook.

jadephilipoom:
Resolves #338 

Mixed feelings on this one. It turns out that the minimize-requires script from coq-tools is a bit fragile. In particular, I ran into the following issues:
- The script doesn't handle "transitive" requires -- if you import Bar, and Bar imports Foo, you can access things from Foo as e.g. `Foo.x`. But the minimize-requires script considers itself free to remove the Bar import, causing errors because now `Foo.x` is undefined.
- Following the same style as absolutizing (see #368) was really tricky because it kept trying to rebuild files that were out of sync; I had to make sure to build, run the minimization, and then rebuild in every directory. Unless you do this in dependency order, you end up running a full build twice for every directory. So I've added it as a simple wrapper meant to be run on a single subdirectory, and then a recursive make target so that it can do the builds in dependency order.
- One definition in `VectorUtils.v` referenced `vcombine` as `VectorUtils.vcombine` but this caused an error in the minimization script because it saves different versions of the `.v` file under temporary file names and therefore can't handle self-referencing.
- The script mysteriously removed a single newline, putting a tactic on the same line as the comment preceding it

I've handled the fragility for now by just printing a big warning over the script that it might over-eagerly remove imports and hotfixing might be required, and in general I think it's fine to add it to the repo in this form. Definitely not one I would advocate for having as a CI check, though.


satnam6502 (approved):
Thank you very much for this cleanup. Much needed!

PR #374, opened 2020-12-07
Alternative sequential semantics with time-indexed values

satnam6502:
This PR creates another sequential semantics interpretation for `CavaClass` that instead of using a list to represent the values over wires time instead using a function that takes a time index and maps it to a value:
```coq
Definition seqVType ticks t := Vector.t (combType t) ticks.
```
This experiment is in `SequentialV.v` for now, with a view to replacing the list based semantics with this formulation as the canonical sequential semantics. Now we know how many `one` and `zero` constant signal values we need to generate and we can enforce that all the input streams are the same length via dependent types:
```coq
 Instance SequentialVectorSemantics {ticks: nat} : Cava (seqVType ticks) :=
  { cava := ident;
    zero := ret (Vector.const false ticks);
    one := ret (Vector.const true ticks);
...
```


jadephilipoom:
I think the most natural way to make this encode the vector for all ticks would be to make the data type a function from the number of ticks to a vector of that length -- but at that point you might as well just make it a function from the number of ticks to the combinational datatype, which is a natural generalization of that encoding. Then `zero` would be `(fun _ : nat => ret false)`, and `xorb` would be e.g. `fun (ticks : nat) (a b : nat -> bool) => ret (xorb (a ticks) (b ticks))`.

jadephilipoom:
If what you want is for the instance to be specialized to a particular number of maximum ticks, you could also make it a field in the class, so `Class Cava {signal : Type} := { ticks : nat }`, and then you'd pass in the ticks along with all the other parameters. I don't think there's a strong reason for it to show up in the type of the Cava instance -- maybe it would help typeclass inference later on, but we can always move it up.

jadephilipoom:
I managed to fix the loopSeqV and delay definitions here with some extra pattern-matching to make the dependent types work:
https://github.com/jadephilipoom/oak-hardware/commit/2593863cc652868e1ef08ddc30f914e3a6f92db1

However, there's still a type error when plugging in `loopSeqV`, to the `Cava` instance, because the loop only takes one input and it's expected to take `ticks`:

```
The term "loopSeqV ticks" has type
 "(seqVType 1 ?A * seqVType 1 ?C -> ident (seqVType 1 ?B * seqVType 1 ?C)) ->
  seqVType ticks ?A -> ident (seqVType ticks ?B)" while it is expected to have type
 "(seqVType ticks s * seqVType ticks s1 -> ident (seqVType ticks s0 * seqVType ticks s1)) ->
  seqVType ticks s -> ident (seqVType ticks s0)".
```

jadephilipoom:
Finished instantiating the cava class here:
https://github.com/jadephilipoom/oak-hardware/commit/9130d33b3ceb0296990322b8098193b17fc66bc6

I worked around the type error in the previous comment by creating a helper function that takes a circuit running for at least one step and runs it for only one step, then using that to have loopSeqV take a multi-step higher-order argument but call loopSeqV' with the single-step version.

Also made a definition for the tuple interface that works with the `seqVType` so that `blackbox` could be instantiated.

jadephilipoom (approved):

PR #377, opened 2020-12-07
Vector counter

jadephilipoom:
Progress for #371
Related to #374 

Proofs of `countBy` using the vector-based sequential semantics. The first two commits are just reorganizing; I moved the `countBy` example out of `Delay.v` and into its own file, and created a directory for all the `countBy` files since I anticipate we'll have a few more tests as well. So the only commit needing review is the third and final one, which adds `tests/CountBy/VectorProofs.v`.

The proof worked out pretty easily. The vector representation does trip up the proof automation a little bit, so I have to sometimes explicitly pass in the number of ticks to rewrites. But I think we can work around that, and all in all I was able to reuse most of the automation from lists, which I think is a good sign.

### DIFF
--- a/tests/CountBy/VectorProofs.v
+++ b/tests/CountBy/VectorProofs.v
@@ -17,7 +17,6 @@
 From Coq Require Import NArith.NArith Vectors.Vector.
 Require Import Coq.Bool.Bvector.
 Import VectorNotations.
-Local Open Scope vector_scope.
 
 Require Import ExtLib.Structures.Monads.
 Export MonadNotation.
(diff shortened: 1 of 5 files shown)


tests/CountBy/VectorProofs.v
```diff
@@ -0,0 +1,104 @@
...
+(* distributed under the License is distributed on an "AS IS" BASIS,        *)
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+From Coq Require Import NArith.NArith Vectors.Vector.
+Require Import Coq.Bool.Bvector.
+Import VectorNotations.
+Local Open Scope vector_scope.
```
satnam6502, line 20: If you moved this to just before the `countBySpec'` definition you can get rid of the `%vector` qualifications later.

satnam6502 (approved):
Great! Thanks.

PR #378, opened 2020-12-07
Cava class hierarchy

jadephilipoom:
Following up on discussion with Satnam (and my own ambitions to do more sequential representation experiments), this PR makes the sequential Cava semantics into a class that "extends" the combinational one.

We had previously discussed putting the combinational semantics inside the class as a field, i.e.
```coq
Class CavaSeq signal :=
  { combinationalSemanatics :> Cava signal;
    ....
   }
```
However, I ran into two problems with that approach in contexts that looked like:
```
Context {signal} {semantics : Sequential signal}.
```
The first problem was that typeclass inference wanted to plug in every concrete instance it could possibly find before trying `(combinationalSemantics semantics)`. Since there are several such instances exported by `CavaMonad`, it was picking up all of those. However, I was able to work around this by decreasing the priority of those instances with `| 10`.

Then I ran into the second problem. In the pipelined nand example, we define a combinational nand and then compose it with delays, so we have a sequential instance in context and pull the combinational instance out of that. However, that now means that the nand we defined expects a _sequential instance_, meaning we can't call `combinational` on it. Already a little unhappy with the `| 10`s, I restructured a bit to make the combinational semantics an _argument_:
```coq
Class CavaSeq {signal}  (combinationalSemanatics : Cava signal) :=
   { ....
   }
```
Later, it can be added to context with
```
Context `{semantics : CavaSeq}.
```
...so the overhead didn't end up being terrible, although it's less than ideal. I ended up plugging in some explicit names for the count-by definition because I wanted to reference them for proofs later.

### DIFF
--- a/cava/Cava/Monad/CavaClass.v
+++ b/cava/Cava/Monad/CavaClass.v
@@ -87,9 +87,9 @@ Class Cava (signal : SignalType -> Type) := {
   blackBox : forall (intf: CircuitInterface),
              tupleInterface signal (map port_type (circuitInputs intf)) ->
              cava (tupleInterface signal ((map port_type (circuitOutputs intf))));
-  (* These members should really be in a class that extends Cava with sequential
-     operations, but I (satnam6502) can't get that to work, so they are linlined
-     here instrad.  *)
+}.
+
+Class CavaSeq {signal : SignalType -> Type} (combinationalSemantics : Cava signal) := {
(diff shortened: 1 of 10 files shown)


satnam6502 (approved):
Thanks! It's good to get the stratification back.
I have some questions about how the overloading works for the pipelined NAND example but I shall follow up verbally.

PR #379, opened 2020-12-08
Sequential semantics in terms of a "timed" monad

jadephilipoom:
I've been playing around with a sequential semantics that uses a monad to represent an infinite stream of values. There are several examples in `TimedMonad.v`. I can then define `loopSeq` in terms of `timedFold`, a generic fold that takes in a starting state, (combinational) loop body, and an input stream and returns an output stream. I was then able to prove the `countBy` example by, instead of doing a proof about looping `countFork`, using a generic lemma to relate `timedFold` to a list `fold_left` and then applying loop-invariant logic.

I was able to prove against this list-based specification:
```coq
Definition countBySpec (i : list (Bvector 8)) :=
  fold_left (fun a b => N2Bv_sized 8 (Bv2N a + Bv2N b)) i (N2Bv_sized 8 0).
```

All in all, my guess right now is that this representation is a little more complex to work with than the list/vector representations. But it still seems worth experimenting on, especially to see if it helps us with sequential loop bodies. I have been trying that some, but no success yet.


satnam6502 (approved):
An interesting addition to our canon of our sequential interpretations. Thank you.

PR #380, opened 2020-12-09
 Change addN to take a pair for better use in combinators

satnam6502:
This PR attempts to change the type of `addN` to take a pair rather than be a curried function, which lets me write `countBy` the way I want to:
```coq
  Definition countBy : signal (Vec Bit 8) -> cava (signal (Vec Bit 8))
    := loop (addN >=> fork2).
```
However, I run into trouble when I try to adapt the List proof. Specifically at:
```coq
Lemma countForkCorrect:
  forall (i : list (Bvector 8)) (s : Bvector 8),
    sequential (loopSeq' (addN >=> fork2) i [s])
    = countBySpec' s i.
```
Before the `sequential` line was:
```coq
    sequential (loopSeq' (countFork (combsemantics:=SequentialCombSemantics)) i [s])
```
which seems to use an annotation to make sure the `countFork` function is interpreted in the `ident` monad.
I can't work out how to write the equivalent thing for `addN >=> fork2` and currently get the error:
```
File "./CountBy/ListProofs.v", line 77, characters 26-40:
Error:
In environment
i : list (Bvector 8)
s : Bvector 8
The term "addN >=> fork2" has type
 "seqType (Vec Bit ?n) * seqType (Vec Bit ?n) ->
  cava (seqType (Vec Bit ?n) * seqType (Vec Bit ?n))"
while it is expected to have type
 "seqType (Vec Bit ?n) * seqType (Vec Bit ?n) ->
  ident (seqType ?B * seqType (Vec Bit ?n))".
```
@jadephilipoom do you know how to fix this? Thank you kindly.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -4,6 +4,7 @@
 -Q third_party/coq-ext-lib/theories ExtLib
 -Q third_party/bedrock2/bedrock2/src/bedrock2 bedrock2
 -Q third_party/bedrock2/deps/coqutil/src/coqutil coqutil
+-Q tests Tests
 -Q monad-examples MonadExamples
 -Q monad-examples/xilinx XilinxExamples
 -Q arrow-examples ArrowExamples
--- a/cava/Cava/Acorn/Identity.v
+++ b/cava/Cava/Acorn/Identity.v
@@ -67,4 +67,4 @@ Ltac simpl_ident :=
                                   instantiate_app_by_reflexivity ]
           | erewrite fold_left_ext; [ | intros; progress simpl_ident;
                                         instantiate_app_by_reflexivity ]
-          | progress cbn [fst snd mcompose bind ret Monad_ident unIdent] ].
+          | progress cbn [fst snd bind ret Monad_ident unIdent] ].
--- a/cava/Cava/Lib/UnsignedAdders.v
+++ b/cava/Cava/Lib/UnsignedAdders.v
@@ -64,11 +64,12 @@ Section WithCava.
             cava (Vector.t (signal Bit) (n + 1)) :=
     adderWithGrowthNoCarryInV (vcombine a b).
 
-  (* Curried add with no carry in and no bit-growth, as signal. *)
+  (* Adder with a pair of inputs of the same size and no bit-growth, as signal. *)
   Definition addN {n : nat}
-            (a b: signal (Vec Bit n)) :
+            (ab: signal (Vec Bit n) * signal (Vec Bit n)) :
             cava (signal (Vec Bit n)) :=
     const0 <- zero ;;
+    let (a, b) := ab in
     '(sum, _) <- unsignedAdderV (const0, vcombine (peel a) (peel b)) ;;
     ret (unpeel sum).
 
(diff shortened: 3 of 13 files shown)


satnam6502:
I have also wondered about changing `addN` to take an explicit `(n: nat)` argument for the size of the vectors to be added, which might help Coq with the type inference.

jadephilipoom:
The problem is that `loopSeq'` is specialized to the identity monad. If we plug in no implicit arguments to `mcompose`, then typeclass resolution ends up trying to unify `@cava seqType ?H = ident`, and thinks they don't match. It works if we explciitly pass the class to `mcompose`:
```coq
Lemma countForkCorrect:
  forall (i : list (Bvector 8)) (s : Bvector 8),
    sequential (loopSeq' (@mcompose (@cava _ SequentialCombSemantics) _ _ _ _ addN fork2) i [s])
    = countBySpec' s i.
```
However, a better solution that keeps the notation working is just to parameterize `loopSeq'` so it works with any monad:
```diff
- Fixpoint loopSeq' {A B C : SignalType}
+ Fixpoint loopSeq' {m} `{Monad m} {A B C : SignalType}
```
I've pushed the latter solution (along with another minor fix to the list proofs -- since some subproofs are now in terms of `mcompose`, we need to not simplify `mcompose` in the `simpl_ident` tactic because it'll get rid of the `mcompose` before we can rewrite with it.

satnam6502:
Thank you! Let me look into the CI failure...

jadephilipoom:
> Thank you! Let me look into the CI failure...

The CI failure is probably the vector proofs -- I didn't change those, but a similar strategy (parameterizing `loopSeqV'` over the monad) should work!

satnam6502:
> The CI failure is probably the vector proofs -- I didn't change those, but a similar strategy (parameterizing `loopSeqV'` over the monad) should work!

So I've tried to write:
```coq
       ((addN >=> fork2) (combsemantics:=SequentialVectorCombSemantics))
```
which turned out to be a wrong guess. I also put back a definition for `countFork` here and tried to `unfold` it later on but that did not work either.

satnam6502:
Now trying:
```coq
Lemma countForkStepOnce (ticks: nat) (i s : Vector.t (Bvector 8) 1) :
  sequentialV
    (stepOnce
       (ticks:=ticks)
       (addN >=> fork2)
       (i, s))
  = (countBySpec' (Vector.hd s) i, countBySpec' (Vector.hd s) i).
```

satnam6502:
So I've got this as far as:
```coq
Lemma countForkCorrect ticks :
  forall(i : Vector.t (Bvector 8) ticks) (s : Vector.t (Bvector 8) 1),
    sequentialV
      (loopSeqV' ticks (stepOnce (ticks:=ticks)
                                 (addN >=> fork2))
                 i s)
    = countBySpec' (Vector.hd s) i.
Proof.
  cbv [sequentialV]; unfold mcompose.
  induction ticks; intros; [ reflexivity | ].
  seqsimpl. cbn [countBySpec']. rewrite IHticks. reflexivity.
Qed.
Hint Rewrite countForkCorrect using solve [eauto] : seqsimpl.
```
But I am blocked at the last `reflexivity`. I was trying to apply `unfold mcompose` at the right place to make the goal look like it did before when it used `countFork`. @jadephilipoom ~ am I going in the wrong direction?

jadephilipoom (approved):

PR #381, opened 2020-12-09
Loop with sequential subcircuit

jadephilipoom:
I think (although not thoroughly tested by any means) I have figured out how to run sequential subcircuits inside of loops for the list-based representation. The key here is the observation that, because delays are represented in the value list, running the loop body on a single input gives me the values on the feedback wire from t=1 onward (1 because of the loop-feedback delay). If there are already values queued on the feedback wire, then my circuit is malformed if it produces anything but delays in those cases -- I can safely overwrite them with the values I already have.

The operation `overlap` performs this chaining of loop step outputs. I've added a test (no proofs yet) that is pretty much `countBy` but with an extra delay inserted inside the loop body.

What do people think of this representation? Especially keen to hear if either of you have any insights about whether it will generalize to subcircuits that are more complicated (I did try with two delays also, and it seems to work, but have not tried loops-within-loops or anything with variable delay).

### DIFF
--- a/cava/Cava/Monad/Sequential.v
+++ b/cava/Cava/Monad/Sequential.v
@@ -29,35 +29,54 @@ From Cava Require Import Signal.
 Require Import Cava.Monad.CavaClass.
 Require Import Cava.Monad.CombinationalMonad.
 
+(* Given two sequential inputs, combine them by combining all the elements of
+   the first with the elements of the second that *do not overlap* when the
+   second is offset from the first by the specified offset. For example:
+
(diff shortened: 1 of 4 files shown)


satnam6502 (approved):
At first sight this looks promising. Let's merge this change and try a few loop within loop examples, some of which should contain pipeline delays. If that works then I think we are in pretty good shape.

jadephilipoom:
Related to #371 , just so we can keep everything tracked on that issue

PR #386, opened 2020-12-14
Loop invariant infrastructure for sequential loops

### LINKED ISSUE #383: Loop -> fold lemmas for sequential semantics
Currently, in the sequential semantics proofs, we need lemmas like:
https://github.com/project-oak/oak-hardware/blob/9f8800357be363bd2982ba1fd666428a9d5bf196/tests/CountBy/ListProofs.v#L80-L83

Ideally, we shouldn't need to refer to `loopSeq'` at all -- we should be able to use an existing lemma that relates the loop body to a Gallina function, and relate the behavior of the full loop to a fold over the behavior of that function. Then we can use our existing loop-invariant lemmas to reason about it in a "friendlier" fashion that doesn't involve the implementation details of our sequential semantics.

We already have this lemma for the timed monad, but need analogous ones for the vector and list representations.

### LINKED ISSUE #384: Simpl in countForkCorrect in ListProofs for countBy causes slowdown
When I adapted the list-base CountBy proof to use `addN >=> fork2` I added a call top `simpl` to make the proof go through but this has caused the proof to go a lo slower:
```coq
Lemma countForkCorrect:
  forall (i : list (Bvector 8)) (s : Bvector 8),
    sequential (loopSeq' (addN >=> fork2) i [s])
    = (countBySpec' s i, countBySpec' s i).
Proof.
  unfold mcompose.
  cbv [sequential]; induction i; intros; [ reflexivity | ].
  seqsimpl. cbn [countBySpec']; simpl; rewrite IHi; reflexivity.
Qed.
```
Is there a better way of making the proof work out using something other than `simpl` which works faster?
See [commit de95434](https://github.com/project-oak/oak-hardware/pull/380/commits/de95434412b61aa27ec644fbe0e7ca5faa86ccd6#diff-b88aa7170b246664c46ac00f4e7efcf670d81eccbf281ad8e225097742009ef8R81-R90).

jadephilipoom:
Resolves #383 
Resolves #384 (the proof was removed entirely!)

Progress on #371

- Redefines loops in terms of Coq's `fold_left`
- Changes the behavior of `overlap` to correctly handle the case where the offset is larger than the length of the first list
- Factors out some lemmas about `overlap` and list-based `loop`, as well as the `seqsimpl` tactic, into a new file `SequentialProperties.v` (only applies to list-based semantics)
- Defines a few lemmas that allow reasoning about loops using invariant logic, in the process improving some of the existing infrastructure for `fold_left` (`loop_invariant` is the main one, but I suspect some cases like `AddWithDelay` might need the slightly modified `loop_invariant_alt`)
- Uses the invariant lemma to prove  the count-by example against a less mechanistic, more intuitive specification

The new `countBy` specification is:
```coq
Definition countBySpec (i : list (Bvector 8)) : list (Bvector 8) :=
  map (fun t => bvsum (firstn t i)) (seq 1 (length i)).
```
Here, `bvsum` sums a list of bitvectors, so this specification says that the output is composed of, for each element of the input list, the sum of all the inputs up to that element's index. It is not stateful.

We can prove that the loop obeys this specification by applying the `loop_invariant` lemma like this:
```coq
  eapply (loop_invariant (B:=Vec Bit 8) (C:=Vec Bit 8)) with
      (I:=fun t acc feedback =>
            feedback = [bvsum (firstn t i)]
            /\ acc = countBySpec (firstn t i)).
```
The `loop_invariant` lemma expects an invariant in terms of the timestep, accumulator for previous outputs, and current unprocessed feedback queue. In this case, the invariant says that the feedback queue contains exactly one element and that it's the sum of the previous inputs, while the accumulator must match the specification for the previous inputs. This `eapply` produces three subgoals: the invariant is satisfied at the start, the invariant holds through the loop, and the invariant implies the postcondition.


cava/Cava/Monad/SequentialProperties.v
```diff
@@ -0,0 +1,257 @@
...
+      reflexivity. }
+    { (* R holds through loop *)
+      subst R. cbv beta zeta.
+      intros. repeat destruct_pair_let. subst.
+      reflexivity. }
+  Qed.
+
+  Lemma loop_invariant {A B C}
+        (f : seqType A * seqType C -> ident (seqType B * seqType C))
+        (I : nat -> seqType B -> seqType C -> Prop)
```
blaxill, line 192: Minor: Perhaps reordering `f`&`I` so you don't need to use `with` in your example `eapply ... with (I:=...`. I guess this only makes sense if it is expected to "know"/or have on hand `I` but not `f` in most use cases. (Up to you)

cava/Cava/Monad/Sequential.v
```diff
@@ -43,39 +44,39 @@ Require Import Cava.Monad.CombinationalMonad.
...
-    '(b', c') <- loopSeq' f xs ys ;; (* remaining steps of f *)
-    ret (overlap 1 b b', overlap 1 c c')
-  | _, _ => ret ([], [])
-  end.
+(* loopSeq' performs a single loop step, given a state consisting of
+the timestep, the accumulator for (past) output values, and the queue
+for (future) feedback values. The nth value in the output accumulator
+is always the output value for timestep n, and the nth value in
+feedback queue is the feedback to be consumed at timestep (t + n),
+where t is the current timestep. *)
```
blaxill, line 66: I'm struggling a bit with trying to wrap my head around this- can you explain the "queue for future feedback" a bit more? Also is the feedback queue of arbitrary size- if it can be `t + n` for any n? (possibly stupid questions)

tests/CountBy/ListProofs.v
```diff
@@ -14,83 +14,84 @@
...
+                                             push_list_fold push_firstn ].
+    f_equal.
+    { apply map_ext_in; intros.
+      match goal with H : _ |- _ => apply in_seq in H end.
+      autorewrite with push_length natsimpl push_firstn push_list_fold.
+      reflexivity. }
+    { cbv [bvadd]. rewrite N.add_comm. reflexivity. } }
+  { (* invariant implies postcondition *)
+    intros *; intros [Hfeedback Hacc]; seqsimpl; subst.
+    rewrite firstn_all; reflexivity. }
```
blaxill, line 96: Is the intention to try and simplify this a bit again? I'm thinking that this is now more intimidating (at least for me!), and presumably would be even more complex for more complex examples?

blaxill (approved):
Lgtm

cava/Cava/Monad/SequentialProperties.v
jadephilipoom: I think it's a good idea to swap the arguments so `with` is not *required*, but I'm probably going to continue using it in most places because that way the proofs don't depend on argument order (and I think in this particular case `I:=` helps to signal "this part is the invariant")

cava/Cava/Monad/Sequential.v
jadephilipoom: Let me see if I can explain it here and then maybe you can recommend a rephrasing based on what parts of the description are helpful?

The feedback queue is arbitrary size because the subcircuit can have an arbitrary amount of delay in it; it's `t + n` where `n` is the amount of delay in the subcircuit. In the list-based semantics, if I have a circuit  with three delays between input and output and I feed it one input `[i]`, it will give me a 4-element list: three default values, and one output that corresponds to the input (`[0;0;0;f(i)]`). In the context of the loop, these represent the values on the feedback wire at `t`, `t+1`, `t+2`, and `t+3`. When I run the next input at timestep `t+1`, I get another 4-element list that gives me values for `t+1` through `t+4` (again, three delays and then a "real" value). Now I have a conflict -- I already supposedly knew what was on the feedback wire at `t+1` through `t+3`. This is where the `overlap` operation comes in. I declare that the circuit would be malformed if the second run produced anything but delays for `t+1` through `t+3`, and I overwrite them with the values from the first four-element list. For the three-delay circuit, we'd get `[0;0;0;f(i0);f(i1)]`.

Is that any clearer now?

blaxill: Yeah thats a bit clearer thanks

tests/CountBy/ListProofs.v
jadephilipoom: It's more complex because the specification is now a lot farther away from the implementation, and we need to now reason about things like `firstn` logically instead of just simplifying until the two sides are equal. I think it wouldn't necessarily be more complex for more complex examples, because in those cases you could just have a lower-level spec that's closer to the implementation and move complexity into pure-Gallina proofs about that spec being equivalent to a higher-level spec. In this case, though, I was trying to stress-test the loop invariant lemma, so I wanted to make sure it could handle some more complex reasoning.

PR #389, opened 2020-12-14
Remove pipeline and loop tests from monad-examples

satnam6502:
These types of tests should really be under the `tests` directory rather than an examples directory.


jadephilipoom (approved):
LGTM

PR #390, opened 2020-12-14
Rename loop to loopDelay

satnam6502:
We should rename `loop` (and `loopm`) to emphasize that we have now fused a register onto the feedback path. This PR changes the names to add the suffix `Delay`.


jadephilipoom (approved):
LGTM. I'll go ahead and merge this so I can rebase the add-with-delay proofs I'm working on on top of it, and avoid future conflicts.

PR #392, opened 2020-12-15
Proofs for add-with-delay

### LINKED ISSUE #387: Proof of add-with-delay circuit
The add-with-delay example added in #381 is our first test of a loop with a sequential subcircuit. In order to make sure that we can reason about this structure, it would be good to prove that it obeys a certain specification. Since it doesn't really correspond to any higher-level logic, the specification can be pretty mechanistic (although in the future we should try more complex and more meaningful examples).

jadephilipoom:
Resolves #387 
Related: #371

I ended up using a pretty non-mechanical specification after all, which follows the function-from-nat model of sequential semantics similar to the timed monad. It was just the most natural way (for me) to encode what this contrived circuit is doing. Nonetheless, the proof went through just fine, although with more list-manipulation reasoning. I've been gradually improving our automation for list proofs, so if we end up wanting to go this direction the proofs should get less annoying over time.

### DIFF
--- a/tests/AddWithDelay/ListProofs.v
+++ b/tests/AddWithDelay/ListProofs.v
@@ -58,7 +58,7 @@ Definition addNSpec {n} (a b : seqType (Vec Bit n)) :=
 
 (* TODO: rename typeclass arguments *)
 Lemma addNCorrect n (a b : list (Bvector n)) :
-  sequential (addN (H:=SequentialCombSemantics) (a, b)) = addNSpec a b.
+  sequential (addN (semantics:=SequentialCombSemantics) (a, b)) = addNSpec a b.
 Admitted.
 Hint Rewrite addNCorrect using solve [eauto] : seqsimpl.
 
--- a/tests/AddWithDelay/ListProofs.v
+++ b/tests/AddWithDelay/ListProofs.v
@@ -56,7 +56,6 @@ Definition addWithDelaySpecF
 Definition addNSpec {n} (a b : seqType (Vec Bit n)) :=
   map2 bvadd a b.
 
-(* TODO: rename typeclass arguments *)
 Lemma addNCorrect n (a b : list (Bvector n)) :
   sequential (addN (semantics:=SequentialCombSemantics) (a, b)) = addNSpec a b.
 Admitted.
--- a/tests/AddWithDelay/ListProofs.v
+++ b/tests/AddWithDelay/ListProofs.v
@@ -1,0 +1,124 @@
+(****************************************************************************)
+(* Copyright 2020 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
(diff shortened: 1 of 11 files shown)


tests/AddWithDelay/ListProofs.v
```diff
@@ -0,0 +1,124 @@
...
+      input 0
+    | S (S t') =>
+      (* if t > 2, out(t) = in(t-1) + out(t-2) *)
+      bvadd (input (S t')) (out t')
+    end.
+
+Definition addNSpec {n} (a b : seqType (Vec Bit n)) :=
+  map2 bvadd a b.
+
+(* TODO: rename typeclass arguments *)
```
blaxill, line 59: Minor: perhaps add an issue if not intending to fix immediately

cava/Cava/Monad/Sequential.v
```diff
@@ -282,4 +282,4 @@ Definition bufBoolList (i : list bool) : ident (list bool) :=
 (* behavioural simulation result.                                             *)
 (******************************************************************************)
 
-Definition sequential {a} (circuit : cava a) : a := unIdent circuit.
+ Definition sequential {A} (circuit : cava A) : A := unIdent circuit.
```
blaxill, line 285: Looks like an unintentional starting space?

tests/AddWithDelay/ListProofs.v
```diff
@@ -0,0 +1,124 @@
...
+Lemma addWithDelayStepCorrect :
+  forall (i : Bvector 8) (s : Bvector 8),
+    sequential ((addN >=> delay >=> fork2) ([i], [s]))
+    = ([bvzero; bvadd i s], [bvzero; bvadd i s]).
+Proof.
+  intros. seqsimpl. reflexivity.
+Qed.
+Hint Rewrite addWithDelayStepCorrect using solve [eauto] : seqsimpl.
+
+Lemma Bv2N_bvzero n : Bv2N (@bvzero n) = 0%N.
```
blaxill, line 74: Perhaps these would be useful to put into BitArithmetic/VectorUtils/new bv proof file.

blaxill (approved):
Lgtm after addressing comments

tests/AddWithDelay/ListProofs.v
jadephilipoom: I went ahead and just fixed it now.

tests/AddWithDelay/ListProofs.v
jadephilipoom: Right now, the `bvadd`/`bvzero` definitions operate on `combType (Vec Bit 8)` rather than a generic bitvector, because otherwise it's a pain to rewrite with lemmas that expect `combType`, so I can't move them to a generic file. I agree that in the long term we should have a nicer interface here, but I'm concerned that creating it for the purpose of these small experiments might be premature optimization. What do you think?

Issue #376, opened 2020-12-07
Add a sequential example & proof that features a basic protocol 


blaxill:
A simple extension to the counter would be a version that receives a valid bit to determine whether to add a value. 

This would be a precursor to reasoning about the `aes_cipher_core` https://github.com/project-oak/oak-hardware/blob/a6a179ef7660d1dbb7283fd17ec22f714ace9920/silveroak-opentitan/aes/Arrow/cipher_core.v#L55

An arrow version of this is at https://github.com/project-oak/oak-hardware/blob/main/arrow-examples/SignalingCounter.v

PR #395, opened 2020-12-17
Add signalling counter example

jadephilipoom:
Progress on #376 

This adds the implementation (no proof yet) of a signalling counter that takes in a valid bit to determine whether it accepts an input. One awkwardness I noticed is that the current loop structure locks us into only one input and one feedback signal for the loop, so in order to take both a valid bit and an 8-bit counter input for each iteration I had to concatenate them into a 9-bit vector. I think we can probably do better; I'll think about it and make an issue so we remember to discuss it once the holidays are over.

### DIFF
--- a/tests/SignallingCounter/SignallingCounter.v
+++ b/tests/SignallingCounter/SignallingCounter.v
@@ -1,0 +1,102 @@
+(****************************************************************************)
+(* Copyright 2020 The Project Oak Authors                                   *)
+(*                                                                          *)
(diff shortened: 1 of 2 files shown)


satnam6502:
Let's add back a tuple type to `SignalType`. Then you can have a tuple of a vector and a bit.

satnam6502 (approved):

satnam6502:
Also, for such "enable" circuits I plan to add another kind of primitive loop called `loopDelayEn` which has the enable control "built into the register" which results in more efficient and idiomatic hardware in general.

PR #400, opened 2020-12-21
Add pairs to SignalType

### LINKED ISSUE #397: Add tuple support to SignalType
Allow tuples to be expressed by `SignalType`. This is needed in general to describe circuits and avoid the issue faces in #395 where a composite value is needed (ideally a pair of an 8-bit value and a boolean value) but @jadephilipoom had to instead construct a 9-bit value. We previously accomplished this using the `Shape` type which let us describe internal tupling structure with port-level tuples flattened. However, this stratification was removed as an unnecessary level of complexity but we have not yet "ported" types to the `SignalType`. We should either reject circuits which have top level tuple type ports or make up fresh names for each element (recursively).

jadephilipoom:
Resolves #397 

- Adds a `Pair` constructor to `SignalType`
- Adds some constructors to `Signal` and fields to `Cava` for constructing and selecting from pairs
- Changes the `mux2` example in `acorn-examples/Multiplexers.v` to use the new pair operations
- Changes `SignallingCounter.v` example to use pairs

The port naming strategy is pretty simplistic at this point -- I just append `"_0"` to the name for the first element of the pair and `"_1"` for the second. For a circuit that exposes nested tuples at the ports, this would transform e.g. `(v1, ((v2, v3), v4)`  with name `"x`"into ports with the names `"x_0"` (v1), `"x_1_0_0"` (v2), `"x_1_0_1"` (v3), and `"x_1_1"` (v4). A nicer way to do this would probably be to make the port name field of `PortDeclaration`, which is currently just a string, into a dependent type, but that also comes with complexity tradeoffs.


cava/Cava/Netlist.v
```diff
@@ -533,6 +556,7 @@ Fixpoint toSignalExpr (t: SignalType) (v: combType t) : SignalExpr :=
   | Void, _ => NoSignal
   | Bit, v => BitVal v
   | Vec vt _, y => VecVal (map (toSignalExpr vt) (Vector.to_list y))
+  | Pair lt rt, y => VecVal [toSignalExpr lt (fst y); toSignalExpr rt (snd y)]
```
jadephilipoom, line 559: @satnam6502 I was unsure about this one; should I instead add a pair type to `SignalExpr`?

cava/Cava/Signal.v
```diff
@@ -168,11 +171,15 @@ Inductive Signal : SignalType -> Type :=
...
   | VecLit: forall {t s}, Vector.t (Signal t) s -> Signal (Vec t s)
+  | SignalPair : forall {t1 t2}, Signal t1 -> Signal t2 -> Signal (Pair t1 t2)
   (* Dynamic index *)
   | IndexAt:  forall {t sz isz}, Signal (Vec t sz) ->
               Signal (Vec Bit isz) -> Signal t
+  | SignalSel : forall {t}, Signal (Pair t t) -> Signal Bit -> Signal t
   (* Static indexing *)
   | IndexConst: forall {t sz}, Signal (Vec t sz) -> nat -> Signal t
+  | SignalFst : forall {t1 t2}, Signal (Pair t1 t2) -> Signal t1
+  | SignalSnd : forall {t1 t2}, Signal (Pair t1 t2) -> Signal t2
```
jadephilipoom, line 182: These extra constructors cause problems for the SystemVerilog printing; it looks like I need to go into `Cava2SystemVerilog` and insert some new case statements to show how to print tuples and fst/snd as SV. But I'm not sure how to do that -- I'm not finding anything online about SystemVerilog even having pairs in the first place. Do I need to remove these?

satnam6502:
So before when we had a tuple type like `Tuple (Vec 8 Bit) Bit` the corresponding type for specifying the port name would have been `string * string`. The stratification meant that elements of tuples were either of the lower-strata type or tuples, which ruled out things like a vector of tuples which meant that you could not express a vector of tuples as the type of a port argument.

Now that we can I think we should perhaps just reject programs that a vectors of tuples at the top-level because we don't (at the moment) want to allow them at the interface, but it is of course fine to have them internally in our descriptions.

satnam6502:
As for SystemVerilog printing, I can add a commit for that.

satnam6502 (approved):

PR #402, opened 2020-12-21
Proofs for signalling counter

### LINKED ISSUE #376: Add a sequential example & proof that features a basic protocol

jadephilipoom:
Resolves #376 

Adds proofs for the signalling counter using our shiny new tuple machinery.

### DIFF
--- a/tests/SignallingCounter/ListProofs.v
+++ b/tests/SignallingCounter/ListProofs.v
@@ -65,7 +65,6 @@ Lemma mux2Correct {A} (sel : seqType Bit) (f t : seqType A) :
 Proof.
   intros; cbv [mux2 sequential]. seqsimpl.
   cbv [pairSel pairSelList pairSelBool SequentialCombSemantics].
-  rewrite map2_swap.
   apply map2_ext; intros; reflexivity.
 Qed.
 Hint Rewrite @mux2Correct using solve [eauto] : seqsimpl.
--- a/tests/SignallingCounter/ListProofs.v
+++ b/tests/SignallingCounter/ListProofs.v
@@ -1,0 +1,112 @@
+(****************************************************************************)
+(* Copyright 2020 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
+(* you may not use this file except in compliance with the License.         *)
+(* You may obtain a copy of the License at                                  *)
+(*                                                                          *)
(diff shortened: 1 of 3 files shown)


tests/SignallingCounter/ListProofs.v
```diff
@@ -0,0 +1,112 @@
...
+
+Require Import Tests.SignallingCounter.SignallingCounter.
+
+Definition bvadd {n} (a b : Signal.combType (Vec Bit n)) : Signal.combType (Vec Bit n) :=
+  N2Bv_sized n (Bv2N a + Bv2N b).
+
+Definition bvzero {n} : Signal.combType (Vec Bit n) := N2Bv_sized n 0.
+
+Fixpoint signallingCounterSpec
+           (input : nat -> combType (Vec Bit 8) * combType Bit)
```
satnam6502, line 43: A general comment rather than something specifically about this function. The value of `combType` is that it tells us for a given interpretation what the corresponding type to use for the "combinational interpretation" e.g. for combinational simulation a `bool` for a single wire, and a `Signal Bit` for the netlist graph. However, for the semantics we don't have overloaded descriptions. So why write `combType (Vec Bit 8)` when you could write what it evaluates to e.g. `Vector.t Bool 8`? Or is there something about how the types are set up which prevents this?

satnam6502 (approved):

tests/SignallingCounter/ListProofs.v
jadephilipoom: There's a method to the madness; if I use `combType` everywhere, it helps `autorewrite` find the places where it can use lemmas. Otherwise, `autorewrite` doesn't see that `Bvector 8` and `combType (Vec Bit 8)` are the same type, because it's (understandably) not computing those types when trying to pattern-match. In the long term, this is a problem better automation can solve; for instance, I could simplify all instances of `combType` that I find. But it's a bit finicky, and I've been trying to minimize tactic optimization for the moment to avoid introducing confusing benavior.

PR #403, opened 2020-12-21
Generate _CoqProject files automatically

jadephilipoom:
We've occasionally had some issues with files not getting added to `_CoqProject`; this PR fixes that by populating the `_CoqProject` files automatically and adding all `.v` files tracked by git. I found a couple of files that had been excluded (not sure whether intentionally or accidentally) and which did not build; I hotfixed them where I could and commented out parts where I couldn't.


satnam6502 (approved):
Thank you very much. A very useful bit of automation.

PR #404, opened 2020-12-22
Add loopDelayEnable for clock-enable registers on feedback loops

satnam6502:
The "signallingCounter" is one way to conditionally update the state of a circuit using an enable signal. However, this leads to inefficient circuits. A much better approach is to be explicit about the clock-enable input.

This is the schematic for the netlist generated by the "signallingCounter" after synthesizing with the Xilinx Vivado tools:

image

This implementation uses 13 LUTs.

The `counterWithEnable` circuit uses the Cava primitive `delayEnable` element via the `loopDelayEnable` combinator which ensures the clock enable is appropriately wired up and recognized:

image

This implementation uses 8 LUTs.


jadephilipoom (approved):
LGTM, once the CI passes.

tests/SignallingCounter/SignallingCounter.v
```diff
@@ -99,3 +99,33 @@ Example signallingCounter_ex2:
...
+                               signal (Vec Bit 8) ->
+                               cava (signal (Vec Bit 8)) :=
+    loopDelayEnable en (addN >=> fork2).
+End WithCava.
+
+(* Note that the signalling counter with delay is slightly different than the
+   one above, in that when not enabled it *does* add the input to the current
+   state, and returns this result, but does not save it. The directly defined
+   version of the counter returns the saved (unchanged) state instead of the
+   unsaved addition result. *)
```
jadephilipoom, line 119: Just checking: is this the expected behavior?

satnam6502: Yes, that's what I expect for the behaviour as written, and as you say it is *not* the same as the "signallingCounter". For that the output of the loop should be the output of the register. Let me adapt it to be so.

satnam6502:
Merging this, but following up with various changes:
1. Add a `loopDelayS` combinator that in turn allows us to match the behaviour of `counterWithEnable` with the "signallingCounter".
2. Consider removing tuples from `SignalType` and add a typeclass that for some type `T` gives the default reset value for that type. This would allow the type of `loopDelay` to be more general and allow us to use Gallina tuples to represent structured state.

Related issue: #371

PR #406, opened 2020-12-23
Add loopDelayS and loopDelaySEnable for state output

satnam6502:
This PR is a work in progress that attempts to add new loop combinators that expose the current state value directly at the output.
See  #371 for details.

The current formulation is wrong as illustrated by the failed `Example` in `SignallingCounter.v":
```coq
Compute map Bv2N (sequential (counterWithEnableTop
                (combine
                   (# [1;1;1;1;1;1;1;1])
                   (map nat2bool [1;1;1;1;0;0;0;0])))).
```
which gives:
```coq
= [4%N]
     : list N
```
i.e. only the last value in the current state stream. I think this is because I've made a wrong guess about how to adapt `loopSeq` to also expose the state stream:
```coq
Definition loopSeqS {A B C : SignalType}
                    (f : seqType A * seqType C -> ident (seqType B * seqType C))
                    (a : seqType A) : ident (seqType B * seqType C) :=
  snd (fold_left (loopSeq' f) a (0, ret ([], [defaultCombValue C]))).
```
It's late so I am going to go to sleep! @jadephilipoom perhaps you can properly adapt `loopSeqS` to also expose the output of the register, thank you.

Once this works we should make `loopDelaySEnable` the primitive loop combinator and define `loopDelay` and `loopDelayEnable` in terms of  `loopDelaySEnable` in the `Combinators.v` file.

### DIFF
--- a/cava/Cava/Acorn/Combinators.v
+++ b/cava/Cava/Acorn/Combinators.v
@@ -295,6 +295,10 @@ Section WithCava.
 
   Definition fork2 `{Monad_m : Monad cava} {A} (a:A) := ret (a, a).
 
+  (****************************************************************************)
+  (* Operations over pairs.                                                   *)
+  (****************************************************************************)
+
   Definition first `{Monad_m : Monad cava} {A B C} (f : A -> cava C) (ab : A * B) : cava (C * B) :=
     let '(a, b) := ab in
     c <- f a ;;
@@ -304,6 +308,16 @@ Section WithCava.
     let '(a, b) := ab in
     c <- f b ;;
     ret (a, c).
+
+  (* Project out the first element of a pair. *)
+  Definition projFst `{Monad_m : Monad cava} {A B} (ab : A * B) : cava A :=
+    let '(a, _) := ab in
+    ret a.
+
+  (* Project out the second element of a pair. *)
+  Definition projSnd `{Monad_m : Monad cava} {A B} (ab : A * B) : cava B :=
+    let '(_, b) := ab in
+    ret b.
 
   (****************************************************************************)
   (* Swap                                                                     *)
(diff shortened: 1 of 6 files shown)


jadephilipoom:
Fixed the problem -- the issue was a subtlety of the `loopSeq` implementation in which the `ident (signal B * signal C)` part of the state doesn't represent both the output and feedback accumulators, as one would assume, but actually the output accumulator and the queue for _future_ feedback. That's why only one element was getting returned.

I changed the type signature of `loopSeqS` slightly -- since we're passing the full output back to the loop body, we actually only need two types, `A` and `B`, and a loop body with type `A * B -> B`. This simplified things a little, and then I was able to change the code for pulling from the feedback queue to instead get a particular value from the output accumulator and use it as feedback. It ends up being a simpler definition than the original `loopSeq`, which is a good sign.

cava/Cava/Acorn/Combinators.v
```diff
@@ -305,6 +309,16 @@ Section WithCava.
...
+  (* Project out the first element of a pair. *)
+  Definition projFst `{Monad_m : Monad cava} {A B} (ab : A * B) : cava A :=
+    let '(a, _) := ab in
+    ret a.
+
+  (* Project out the second element of a pair. *)
+  Definition projSnd `{Monad_m : Monad cava} {A B} (ab : A * B) : cava B :=
+    let '(_, b) := ab in
+    ret b.
+
```
jadephilipoom, line 321: Are these unused?

jadephilipoom (approved):

cava/Cava/Acorn/Combinators.v
satnam6502: They are now with the current formulation for count-by with enable, but the previous versionI wrote used `projSnd`. Let's just leave them in anyway as we are bound to use them sooner or later.

satnam6502:
@jadephilipoom thanks for the `loopSeqS` simplifications!

Issue #394, opened 2020-12-16
Modify Acorn AES to expand keys on-the-fly


jadephilipoom:
Currently, like the spec, it assumes the keys are precomputed. But to be a drop-in replacement for OpenTitan's AES, we would need to expand the keys on the fly as OpenTitan does. There's already proofs that link on-the-fly to precomputed keys for arrows, so we should be able to reuse some things.

PR #410, opened 2021-01-05
Groundwork for more OpenTitan-like Acorn cipher

jadephilipoom:
Progress on #394 

Our current Acorn AES cipher is based on the spec and uses precomputed keys. This PR is part 1 of 2 in changing that design to one that mirrors OpenTitan more closely. In particular, the new version will:
- expand keys on-the-fly by routing through a key expansion circuit (nonexistent, for now)
-  operate more like a state machine; more details in the next PR.

I've separated out all the general-purpose infrastructure (i.e. the changes to `cava`) required fro the change into this "part 1" PR. It's kind of a random mix, but some of the highlights include:
- new tactics - these might be useful for others to take a look at for future proofs!
  -  `subst_lets` (inlines all hypotheses bound with `:=`)
  - `compute_expr` (replaces a particular subexpression with the computed version of itself)
  -  `boolsimpl` (simplifies boolean expressions)
- new Acorn combinators
  -  `all` (produces a bit indicating whether vector of bits is all true),
  - `eqb` (compares *any* two signals of a particular type and produces a bit indicating whether they are equal)
  - `pairAssoc` (associates signal pairs)
  -  `mux4` (a generic 4-input mux)
  - `tree_all_sizes` (variant of the tree combinator which doesn't require the input vector length to be expressed as a power of 2)
- `NatUtils.v` file for generic proofs about `nat` and `N` (and conversions between the two)

Smaller changes include:
- much-improved error messages for the `factor_out_loops` tactic
- misc. bit-arithmetic and list helper lemmas

### DIFF
--- a/cava/Cava/Acorn/CombinationalProperties.v
+++ b/cava/Cava/Acorn/CombinationalProperties.v
@@ -1,0 +1,69 @@
+(****************************************************************************)
+(* Copyright 2020 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
+(* you may not use this file except in compliance with the License.         *)
+(* You may obtain a copy of the License at                                  *)
+(*                                                                          *)
+(*     http://www.apache.org/licenses/LICENSE-2.0                           *)
+(*                                                                          *)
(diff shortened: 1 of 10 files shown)


blaxill (approved):
Lgtm! Just some minor notes.

PR #412, opened 2021-01-05
More OpenTitan-like Acorn cipher

### LINKED ISSUE #394: Modify Acorn AES to expand keys on-the-fly

jadephilipoom:
Resolves #394 
Part 2 of 2 (Part 1 was #410)

There are two main changes in this PR that serve to make the Acorn AES implementation more similar to OpenTitan:
1. We now assume a key expansion circuit exists and call it on-the-fly to generate round keys.
2. The top-level cipher is now defined in terms of `cipher_step`, which given a round index and the current state performs one step of AES (including setting selectors based on the round index and then using them with muxes to choose the right data path).

Most of the changes are modifying the proofs to match the new structure. They're now a longer and more complicated-looking because of bookkeeping deal with the fact that the spec needs precomputed keys (previously, since the spec and the combinational interpretation of the circuit were basically identical, the proofs were very simple). With some additional effort I think that we can reduce the complexity  some (for instance. by making some changes to `ExpandAllKeys.v`), but this change was large enough already so that will be left to a future PR.

Currently, we loop `cipher_step` over a provided list of indices as bitvectors; as the next step, we should instead use a sequential loop that increments the index and stops when it reaches the correct number of rounds (I'll make an issue for this).

### DIFF
--- a/silveroak-opentitan/aes/Acorn/AES256Equivalence.v
+++ b/silveroak-opentitan/aes/Acorn/AES256Equivalence.v
@@ -14,9 +14,14 @@
 (* limitations under the License.                                           *)
 (****************************************************************************)
 
+Require Import Coq.Arith.PeanoNat.
 Require Import Coq.Vectors.Vector.
+Require Import Coq.Lists.List.
+Require Import coqutil.Tactics.destr.
 Require Import ExtLib.Structures.Monads.
+Require Import Cava.BitArithmetic.
 Require Import Cava.ListUtils.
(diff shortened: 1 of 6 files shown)


blaxill (approved):
Lgtm!

PR #415, opened 2021-01-06
Rename signalling counter to more accurate name

satnam6502:
Rename the "signalling counter" to reflect the fact that is is an accumulating adder with a clock-enable input.


satnam6502:
Self-merging cleanup.

Issue #416, opened 2021-01-06
Attempt to remove tuples from SignalType


satnam6502:
Tuples were recently introduced into the `SignalType` despite my desire to avoid having them in that type and attempt to just use Gallina tuples in our circuit descriptions. The need to create default values for the structured state output of `loopDelay` meant that we need some extra information somewhere to allow the default values to be determined (i.e. the initial value of the `delay` circuit after the reset signal has been asserted). Currently we use tuples to help get this information (since we know the type of the structured data in the tuple so can compute the default value for each component signal). Alternatively, we could use a typeclass like `DefaultSignalValue` and has an instance for Gallina's tuples and then `loopDelay` can have a more general type that uses Gallina tuples and we can banish tuples from `SignalType` which will lead to simplification of descriptions.

satnam6502:
Excellent. Good riddance to them.

PR #417, opened 2021-01-07
Use loopDelayS as the primary loop definition

jadephilipoom:
Resolves a TODO in `CavaClass.v`. The story here is that we initially defined `loopDelay`, a loop whose body has separate output and feedback results; the body's signature is something like like `A * C -> B * C`. We later added `loopDelayS`, which assumes the output and feedback are the same and expects a body with the form `A * B -> B`. There was a TODO to remove `loopDelay` from the typeclass and simply define it in terms of `loopDelayS`, which is what this PR does.

There is a significant proof component here because we had written invariant infrastructure for `loopDelay`, so I redid these proofs for `loopDelayS` instead and then used them to re-prove the top-level invariant proofs for `loopDelay`. Since the existing examples in `tests/` do in fact use the same thing for feedback and output, I went ahead and switched them over to `loopDelayS`.

I also removed the `defaultCombType` and `combType` definitions from `CombinationalMonad.v` because they are exact duplicates of the ones in `Signal.v` and it's confusing to have both.

### DIFF
--- a/cava/Cava/Acorn/CavaClass.v
+++ b/cava/Cava/Acorn/CavaClass.v
@@ -99,25 +99,12 @@ Class CavaSeq {signal : SignalType -> Type} (combinationalSemantics : Cava signa
   delay : forall {t: SignalType}, signal t -> cava (signal t);
   (* A unit delay with enable. *)
   delayEnable : forall {t: SignalType}, signal Bit -> signal t -> cava (signal t);
-  (* Feedback loop, with unit delay inserted into the feedback path.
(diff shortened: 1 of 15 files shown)


blaxill (approved):
Lgtm. Does this help with #416 at all? I guess not as there is still a tupled input.

jadephilipoom:
No, it unfortunately still doesn't help with #416 because, no matter how we write loops, we will sometimes want to have more than one non-tupled signal in the loop state.

Issue #413, opened 2021-01-05
Make Acorn AES use a sequential loop


jadephilipoom:
See #412 

Currently, the Acorn AES cipher is implemented by using the `foldLM` combinational loop function to iterate over a provided list of round indices, producing an unrolled loop. Instead, we should use a sequential loop that increments the round index and stops when it reaches the correct number of rounds.

PR #421, opened 2021-01-08
Increment and decrement combinators

jadephilipoom:
Progress towards #413

- `half_adder` and `half_subtractor` combinators
- Combinators `incr4` and `decr4`, which increment/decrement 4-bit vectors
- Combinators `incr` and `decr` that increment and decrement arbitrarily-sized bit vectors
- Proofs for all of the above

I originally wrote `incr4` and `decr4` as smaller tests/experiments to help me write the general versions, but I left them in because it might be interesting to see if they generate the same netlist as the generalized versions. I have no objection to removing them if people would rather not have superfluous combinators lying around.


cava/Cava/Acorn/Combinators.v
```diff
@@ -581,6 +581,40 @@ Section WithCava.
     let x := pairAssoc input in
     pairSel (indexConst sel 0) (pairSel (indexConst sel 1) x).
 
+  Definition half_adder (input : signal Bit * signal Bit)
```
satnam6502, line 584: You could use the `halfAddee` in `Cava.Lib.FullAdder`

cava/Cava/Acorn/Combinators.v
```diff
@@ -581,6 +581,40 @@ Section WithCava.
...
+    : cava (signal Bit * signal Bit) :=
+    sum <- xor2 input ;;
+    carry <- and2 input ;;
+    ret (sum, carry).
+
+  (* increment a 4-bit vector *)
+  Definition incr4 (input : signal (Vec Bit 4))
```
satnam6502, line 591: Although we can build adders etc. out of bits we can also use synthesized adders which will always get mapped to good hardware i.e. use `unsignedAdd`.

cava/Cava/Acorn/Combinators.v
```diff
@@ -581,6 +581,40 @@ Section WithCava.
...
+    : cava (signal (Vec Bit 4)) :=
+    true_ <- one ;;
+    '(sum0, carry) <- half_adder (true_, indexConst input 0) ;;
+    '(sum1, carry) <- half_adder (carry, indexConst input 1) ;;
+    '(sum2, carry) <- half_adder (carry, indexConst input 2) ;;
+    '(sum3, carry) <- half_adder (carry, indexConst input 3) ;;
+    ret (unpeel [sum0;sum1;sum2;sum3]%vector).
+
+  Definition half_subtractor (input : signal Bit * signal Bit)
```
satnam6502, line 616: Let me add a synthesized unsigned subtractor.

cava/Cava/Acorn/Combinators.v
jadephilipoom: I thought `Lib` was build on top of `Combinators.v` -- what's the difference/dependency between the two? Should we move these definitions to `Lib`, perhaps?

cava/Cava/Acorn/Combinators.v
jadephilipoom: Is there a performance difference between making this kind of half-adder incrementer, which relies on one argument being 1, and using `unsignedAdd` with 1?

cava/Cava/Acorn/Combinators.v
jadephilipoom: Based on offline discussion, definitions have now been moved to `acorn-examples`.

cava/Cava/Acorn/Combinators.v
satnam6502: No, it should work out as the same thing.

satnam6502 (approved):
Let's go with the gate-level 4-bit increment/decrement for now.
At a later stage we can think of replacing these operations with their synthesizable equivalents.

PR #424, opened 2021-01-08
Consolidate combinational semantics

jadephilipoom:


cava/Cava/Acorn/Identity.v
```diff
@@ -14,31 +14,41 @@
...
 Section Combinators.
-  Lemma zipWith_unIdent {A B C : SignalType} n f va vb :
-    unIdent (@zipWith _ _ Monad_ident A B C n f va vb)
-    = map2 (fun a b => unIdent (f (a,b))) va vb.
+  Lemma zipWith_unIdent {A B C : SignalType} n
+        (f : seqType A * seqType B -> cava (seqType C))
+        (va : combType (Vec A n)) (vb : combType (Vec B n)) :
+    unIdent (@zipWith _ _ Monad_ident A B C n f [va] [vb])
+    = Vector.map2 (fun a b => unIdent (f ([a],[b])))
+                  (peel [va]) (peel [vb]).
```
jadephilipoom, line 40: I'm currently stuck on this lemma. The current version doesn't typecheck because it's a vector of lists instead of a list of vectors. I can get it to typecheck but then I can't prove it.

satnam6502: I _think_ a related and more general issue is how we want to attribute semantics to circuits that have input sequences of different lengths (and possibly empty input sequences) as well as whether we want to be explicit in the context for how many cycles this "simulation" runs for. I will follow up further 1:1.

jadephilipoom:
I fixed this issue by ruling out the nil-vector case for `zipWIth`, but the change is fairly large so I'm separating it out and merging in PRs that don't share history with this one.

PR #428, opened 2021-01-14
Change oak-hardware to silveroak in root README.md

satnam6502:


satnam6502:
I am abandoing the rename to silveroak now because it breaks CI. Will ask @blaxill for help.

Issue #422, opened 2021-01-08
Consolidate combinational Cava instances


jadephilipoom:
At the moment, we have separate `Cava` (combinational semantics) typeclass instances for "combinational circuits" (`CombinationalSemantics` from `CombinationalMonad.v`) and "combinational components of sequential circuits" (`SequentialCombSemantics` from `Sequential.v`). This means that proofs about combinational components don't carry over once the combinational component is put into a sequential circuit unless the proof is done twice. We should either:

1. Make the `SequentialCombSemantics` version the canonical combinational semantics, since the only really interesting circuits to us have sequential components. Port all the existing proofs over to `SequentialCombSemantics`. The main potential pitfall here is increased complexity in the combinational proofs/definitions.
2. Don't require the sequential and combinational typeclasses to use the same datatype; let the sequential typeclass use a projection from the combinational datatype (e.g. `list`), and make the `CombinationalSemantics` instance the canonical combinational semantics. The main potential pitfall is that then we're dealing with differently-typed data for sequential and combinational components, which could complicate circuit descriptions.

Complicating circuit descriptions is worse than complicating proofs of combinational combinators, so I'll try 1) first, and if I run into problems I'll try 2) and see how it goes.

jadephilipoom:
This is a prerequisite for #413, because we will need to make a loop and use proofs of combinational combinators to prove properties about it.

PR #430, opened 2021-01-15
Groundwork for consolidating combinational semantics

jadephilipoom:
Part 1 of 2
Progress towards #422 

- Make `SequentialV` and `SequentialTimed` independent of `CombinationalMonad`
  - move some bit-arithmetic definitions from `CombinationalMonad` to `BitArithmetic`
  - update the timed monad to use `loopDelayS`
- New definition `extend` and a few proofs in `ListUtils`
- New definitions `InV` and `ForallV` in `VectorUtils`[0]
- Generalization of `VectorUtils` lemmas about "algebraic" loops with identity elements and group-like operations
- Generalization of `tree` lemmas to accept validity preconditions on algebraic elements
- Miscellaneous useful helper lemmas about vectors

I've now completed the task set out in #422, but the change is very large, so I've separated out some changes that are required for that change but still build when separated from it (for instance, helper lemmas in `VectorUtils.v` and `ListUtils.v`). I want to do a little more investigation into alternate strategies before merging part 2, to see if we can retain the ability to avoid list reasoning in combinational proofs. However, regardless of the outcome of that investigation, I think this part 1 should be merged, since it contains generally useful lemmas and updates.

[0] These serve the same purpose as `Vector.In` and `Vector.Forall`, but I had to redefine them because for some reason, the standard library `Vector.In` and `Vector.Forall` are defined in a way that simply makes them impossible to reason about in proofs without importing axioms. The new definitions are similar to the way the standard library's `List.In` works; they're fixpoints that compute a `Prop` based on the vector, rather than being an inductive type themselves.

### DIFF
--- a/cava/Cava/Acorn/CavaClass.v
+++ b/cava/Cava/Acorn/CavaClass.v
@@ -121,8 +121,7 @@ Class CavaSeqMonad {signal : SignalType -> Type} (combinationalSemantics : Cava 
   (* A unit delay. *)
   delaym : forall {t: SignalType}, cava (signal t) -> cava (signal t);
   (* Feeback loop, with unit delay inserted into the feedback path. *)
-  loopDelaym : forall {A B C: SignalType},
-               (signal A * signal C -> cava (signal B * signal C)) ->
-               cava (signal A) ->
-               cava (signal B);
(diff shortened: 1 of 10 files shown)


satnam6502 (approved):
Stamping LGTM.

PR #431, opened 2021-01-15
Consolidate combinational semantics

### LINKED ISSUE #422: Consolidate combinational Cava instances

jadephilipoom:
Resolves #422 

This PR redefines `CombinationalSemantics` on a list datatype (previously `SequentialCombSemantics`), and adapts all our existing combinational proofs/tests to be proofs over singleton lists. The change is fairly huge, because I had to change pretty much every circuit proof or test in the repository to get it to compile. To make reviewing easier, I separated it into two commits, one for "boilerplate" changes that are essentially just putting list expressions around things, and one for the changes to the core typeclasses and changes that required actual modification of proofs. **Definitely best reviewed commit-by-commit**.

In the future, we probably want to prove our combinational circuits for any length of input, rather than just for singleton lists; this was just the quickest, lowest code-churn way I could get the repo to build. The singleton-list proofs should be enough for AES, since the combinational part is wrapped in a top-level loop that calls the body on singleton input.

~This is a draft PR for now because, given that this change somewhat significantly increases proof complexity, I'd like to do a little more investigation into alternatives before we commit to it. But I figured it might be useful to put it up for viewing and discussion anyway.~

One lingering TODO here is that the `UnsignedAdderProofs` are currently broken and admitted -- they're not on the critical path to AES and it seemed like the proof-porting process wouldn't be straightforward, so I thought I'd save that change for a later PR.

### DIFF
--- a/cava/Cava/Acorn/Identity.v
+++ b/cava/Cava/Acorn/Identity.v
@@ -14,46 +14,151 @@
 (* limitations under the License.                                           *)
 (****************************************************************************)
 
+Require Import Coq.Lists.List.
 Require Import Coq.Vectors.Vector.
+Require Import Coq.micromega.Lia.
+Require Import coqutil.Tactics.Tactics.
 Require Import ExtLib.Structures.Monad.
 Require Import ExtLib.Structures.MonadLaws.
 Require Import Cava.Signal.
 Require Import Cava.Tactics.
+Require Import Cava.ListUtils.
 Require Import Cava.VectorUtils.
 Require Import Cava.Lib.BitVectorOps.
 Require Import Cava.Acorn.CavaClass.
 Require Import Cava.Acorn.CombinationalMonad.
 Require Import Cava.Acorn.Combinators.
+Import VectorNotations ListNotations.
+Open Scope list_scope.
 
 Existing Instance CombinationalSemantics.
 
+
+Section MapT.
+  Lemma mapT_vector_ident {A B n} (f : A -> ident B) (v : Vector.t A n) :
+    unIdent (mapT_vector f v) = map (fun a => unIdent (f a)) v.
+  Proof.
+    induction v; intros; [ reflexivity | ].
+    cbn. rewrite IHv. reflexivity.
+  Qed.
+End MapT.
+Hint Rewrite @mapT_vector_ident using solve [eauto] : simpl_ident.
+
+Section PeelUnpeel.
+  Lemma peelVecList_cons_cons {A n} (a : combType A) (v : combType (Vec A n)) :
+    peelVecList [(a :: v)%vector] = ([a]%list :: peelVecList [v])%vector.
+  Proof.
+    cbv [peelVecList indexConstBoolList]. cbn [List.map Vector.map].
+    apply to_list_inj. autorewrite with push_to_list.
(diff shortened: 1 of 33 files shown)


jadephilipoom:
There's a Haskell error I can't quite figure out:
```
Setup.hs: can't find source for EquivDec in ., dist/build/autogen,
dist/build/global-autogen
```

We're not using `EquivDec` (a `coq-ext-lib` library) anywhere, as far as I can tell, but it's listed under `other-modules` in the `Cava2HDL.cabal` file. Removing it just leads to the same error about other modules. Not sure what's going on here; @satnam6502  @blaxill could you take a look?

blaxill:
Removing EquivDec, Injection, List0 and RelDec from Cava2HDL works. Prior to this PR `List0` was required by `Sequential.hs` and the others were transitively required.

jadephilipoom:
Thanks, that worked!

cava/Cava/Acorn/Identity.v
```diff
@@ -14,46 +14,151 @@
...
+    inversion Heq; reflexivity.
+  Qed.
+End PeelUnpeel.
+
 (* Lemmas about combinators specialized to the identity monad *)
 Section Combinators.
-  Lemma zipWith_unIdent {A B C : SignalType} n f va vb :
-    unIdent (@zipWith _ _ Monad_ident A B C n f va vb)
-    = map2 (fun a b => unIdent (f (a,b))) va vb.
+  Lemma zipWith_unIdent {A B C : SignalType} n
```
satnam6502, line 101: Wow, so this got a lot more complicated!

satnam6502 (approved):
Stamping LGTM.

cava/Cava/Acorn/Identity.v
jadephilipoom: Yes; that's why I wanted to do some extra experiments before merging, because all the combinational proofs that weren't brute force got more annoying -- and some, like this one, got much more annoying. But my conclusion from the additional experiments was that it would take days or weeks more of experimentation + design and a very major overhaul to preserve the old combinational proof style, so I think it's best left for phase 2.

Issue #432, opened 2021-01-15
Rework constant values.


satnam6502:
Replace the current `one` and `zero` circuits with *expressions* based on `SignalType`.
In `CavaClass` provide a method `constant: bool -> signal Bit` which takes a `bool` value and returns its denotation for a given interpretation.
I think this leads to nicer use to constant bit values as well as nicer constant bit-vectors.

CC: @smore-lore

satnam6502:
Done in #475

PR #433, opened 2021-01-15
Define a constant member to map bool values to SignalType values

satnam6502:
Address #432 
Adds a `constant` member to Cava class to map `bool` values to values denoted via `SignalType`.

### DIFF
--- a/cava/Cava/Acorn/CavaClass.v
+++ b/cava/Cava/Acorn/CavaClass.v
@@ -27,6 +27,7 @@ Local Open Scope type_scope.
 Class Cava (signal : SignalType -> Type) := {
   cava : Type -> Type;    
   (* Constant values. *)
+  constant : bool -> signal Bit;
   zero : cava (signal Bit); (* This component always returns the value 0. *)
   one : cava (signal Bit); (* This component always returns the value 1. *)
   (* Default values. *)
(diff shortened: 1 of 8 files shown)


cava/Cava/Acorn/CavaClass.v
```diff
@@ -27,6 +27,7 @@ Local Open Scope type_scope.
 Class Cava (signal : SignalType -> Type) := {
   cava : Type -> Type;    
   (* Constant values. *)
+  constant : bool -> signal Bit;
   zero : cava (signal Bit); (* This component always returns the value 0. *)
   one : cava (signal Bit); (* This component always returns the value 1. *)
```
smore-lore, line 32: I suppose my question here is then, can these be 'zero : signal Bit' and 'one : signal Bit'?

In either case, could we de-dupe these? (either by removing zero/one, or defining them as a function)

For instance, even if these need the monad, could we do something like this:

    Definition zero : cava (signal Bit) := ret (constant false)
    
Otherwise, I feel we should require proofs that 'constant true = one' (or there is some case where this shouldn't be true?)

satnam6502: Yes, my goal is to move `one` and `zero` to a library module and have them defined as `one := constant true` and `zero := constant true`. I thought it would be good to stage these changes as different PRs.

jadephilipoom (approved):
LGTM!

PR #436, opened 2021-01-18
Experiment with stream semantics

satnam6502:


satnam6502:
Self-merging experiment.

Issue #440, opened 2021-01-19
CI checking of SystemVerilog tests broken


satnam6502:
When a SystemVerilog test fails it does not fail CI, allowing us to check in code that generated bad SystemVerilog.
@blaxill please can you investigate and fix this? Thank you kindly.
Repo, simply type `make` from the root directory and observe the `countBy` test fail.
```
clang++ -L/usr/local/opt/sqlite/lib    countBy_tb.o verilated.o verilated_vcd_c.o VcountBy_tb__ALL.a    -o VcountBy_tb -lm -lstdc++
obj_dir/VcountBy_tb
                  10: tick = 0, i = 14, o = 15
[10] %Error: countBy_tb.sv:41: Assertion failed in TOP.countBy_tb:  14 15
%Error: countBy_tb.sv:41: Verilog $stop
Aborting...
make: *** [countBy_tb.vcd] Abort trap: 6
make: *** Deleting file `countBy_tb.vcd'
```

satnam6502:
As a hint, we've had properly checked SystemVerilog tests for a long time that would fail CI when the test failed. Here I wonder if there is a different type of failure occurring that the C++ level perhaps related to an error in how the test was set up.

satnam6502:
Looking at https://pantheon.corp.google.com/cloud-build/builds/45fa50af-e452-44a3-a669-9e663efe52e5?project=oak-ci
I see the tests are generated but not run for the `tests` directory and the `tests/xilinx` directory and ` acorn-examples` and ` acorn-examples/xilinx`. Perhaps an error in the Makefile? But they do work when run locally on my laptop. Why do they not run in CI?

satnam6502:
So I think this might be an error in the implicit dependencies which manifests in CI but not when run locally. Taking over issue and digging deeper.

satnam6502:
Makefiles fixed. Now looking tests that fail which went unnoticed.

satnam6502:
I think this is now resolved.

PR #444, opened 2021-01-19
Deliberately sabotage build to check CI [DO NOT MERGE]

satnam6502:
This PR contains a commit with a sabotaged tests for `countBy` which should fail, and in turn make the CI fail.
```
clang++ -L/usr/local/opt/sqlite/lib    countBy_tb.o verilated.o verilated_vcd_c.o VcountBy_tb__ALL.a    -o VcountBy_tb -lm -lstdc++
obj_dir/VcountBy_tb
                  10: tick = 0, i = 14, o = 15
[10] %Error: countBy_tb.sv:41: Assertion failed in TOP.countBy_tb: For o expected 14 but got 15
%Error: countBy_tb.sv:41: Verilog $stop
Aborting...
make: *** [countBy_tb.vcd] Abort trap: 6
```

Investigating #440

### DIFF
--- a/cava/Cava/Acorn/NetlistGeneration.v
+++ b/cava/Cava/Acorn/NetlistGeneration.v
@@ -289,7 +289,7 @@ Definition instantiateNet (intf : CircuitInterface)
 
 Instance CavaCombinationalNet : Cava denoteSignal := {
     cava := state CavaState;
-    constant b := if b then Vcc else Gnd;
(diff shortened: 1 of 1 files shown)


satnam6502:
So this does cause the CI to fail. However, it continues to run other tests (some of which also fail due to the sabotage change). How about just stopping the CI at the first failure? @blaxill

PR #445, opened 2021-01-19
Remove tuples from SignalType and replace with SignalInterface

### LINKED ISSUE #416: Attempt to remove tuples from SignalType

jadephilipoom:
Resolves #416 

Previously, we had `Pair` as a constructor for `SignalType` so that definitions like loops, which are parameterized over `SignalType`, could have complex types. Because the type of a loop state had to have the form `signal B`, you couldn't make a loop state with a Gallina tuple. This PR fixes that problem by introducing a layer of abstraction for "a collection of signals", and modifying high-level combinators to use it. The key infrastructure is here (from `Signal.v`):

```coq
(* A collection of SignalTypes *)
Inductive SignalInterface : Type :=
| ione : forall t : SignalType, SignalInterface
| ipair : forall t1 t2 : SignalInterface, SignalInterface
.

(* The Gallina type that represents a particular signal interface, based on a
   single-signal interpretation *)
Fixpoint signals_gen (signal : SignalType -> Type)
         (i : SignalInterface) : Type :=
  match i with
  | ione t => signal t
  | ipair i1 i2 => (@signals_gen signal i1 * @signals_gen signal i2)
  end.

(* This coercion automatically converts a single SignalType written in a place
   that expects a SignalInterface into a one-element SignalInterface *)
Coercion ione : SignalType >-> SignalInterface.
```

This way, we can define something like `loopDelayS` for sequential semantics with the signature:
```coq
Definition loopSeqS {A B : SignalInterface}
           (f : signals_gen seqType A * signals_gen seqType B
                -> ident (signals_gen seqType B))
           (a : signals_gen seqType A) : ident (signals_gen seqType B) :=
```

I've also defined a shortcut definition, `signals`, in `CavaClass.v`, which automatically plugs in the class's `signal` definition to `signals_gen`, which makes definitions that come after the Cava instance declaration less verbose than the one above.

Currently, this change means we have to explicitly feed type arguments to loops and delays in more places. One resolution to this would be to define everything in terms of `SignalInterface` instead of `SignalType`. There are also a few places where proofs were broken and still need to be fixed (namely the `loopDelay` invariant proofs, currently unused, and the `CountBy` vector-semantics proofs). However, I think it's probably best to polish those things after we're done with #431 or the alternative to it that I've been exploring, because it will break the proofs all over again.


cava/Cava/Acorn/Sequential.v
```diff
@@ -292,26 +316,29 @@ Definition delayEnableBoolList (t: SignalType) (en: list bool) (i : seqType t) :
...
-                     let '(en, i) := unpair (fst en_i_state) in
-                     (second fork2 >=> (* (i, state, state) *)
-                       pairLeft >=>    (* ((i, state), state) *)
-                       first f >=>     (* (f (i, state), state) *)
-                       swap >=>        (* (state, f (i, state) *)
-                       mux2 en)        (* if en then f (i, state) else state *)
-                       (i, state)) (mkpair en input)
+       loopSeqS
+         (A:=ipair Bit A)
+         (B:=B)
```
satnam6502, line 327: Ah, so this is an example of adding more explicit type information compared to before.

cava/Cava/Acorn/SequentialV.v
```diff
@@ -42,15 +42,63 @@ Local Open Scope vector_scope.
...
+Fixpoint uncons_signals {A : SignalInterface} {ticks: nat}
+  : signals_gen (seqVType (S ticks)) A
+    -> signals_gen (seqVType 1) A * signals_gen (seqVType ticks) A :=
+  match A as A return signals_gen (seqVType (S ticks)) A
+    -> signals_gen (seqVType 1) A * signals_gen (seqVType ticks) A with
+  | ione t => fun i : seqVType (S _) t => ([Vector.hd i], Vector.tl i)
+  | ipair t1 t2 =>
+    fun i => let '(hd1, tl1) := uncons_signals (fst i) in
+          let '(hd2, tl2) := uncons_signals (snd i) in
+          ((hd1, hd2), (tl1, tl2))
```
satnam6502, line 63: Indentation.

cava/Cava/Signal.v
```diff
@@ -27,9 +27,31 @@ Inductive SignalType :=
...
   | ExternalType : string -> SignalType.            (* An uninterpreted type *)
 
+(******************************************************************************)
+(* Define the type of a collection of one or more differently-typed signals   *)
+(******************************************************************************)
+
+(* A collection of SignalTypes *)
+Inductive SignalInterface : Type :=
+| ione : forall t : SignalType, SignalInterface
+| ipair : forall t1 t2 : SignalInterface, SignalInterface
```
satnam6502, line 39: So this I think is very similar to the `Tuple` type we had before and much of the code that matches on a "one" or "many" before looks familiar i.e. the stratified tuple-type we had before. I think a key difference in this approach is the use of a coercion?

tests/Delay.v
```diff
@@ -147,9 +147,9 @@ Definition delayEnableByte_tb
 Section WithCava.
   Context `{semantics: CavaSeq} `{Monad cava}.
 
- (* A nand-gate with registers after the AND gate and the INV gate. *)
+  (* A nand-gate with registers after the AND gate and the INV gate. *)
   Definition pipelinedNAND : signal Bit * signal Bit -> cava (signal Bit) :=
-    nand2 >=> delay >=> inv >=> delay.
+    (nand2 >=> delay (t:=Bit) >=> inv >=> delay (t:=Bit)).
```
satnam6502, line 152: So this is an unfortunate consequence. Although I am happy to tolerate pain in "internals" I am loathed to go from a "user" description that goes from:
```coq
Definition pipelinedNAND : signal Bit * signal Bit -> cava (signal Bit) :=
  nand2 >=> delay >=> inv >=> delay.
```
to:
```coq
Definition pipelinedNAND : signal Bit * signal Bit -> cava (signal Bit) :=
    (nand2 >=> delay (t:=Bit) >=> inv >=> delay (t:=Bit)).
```

satnam6502 (commented):
Let's have a face to face discussion. Although I am keen to avoid code that maps between our tuples and Gallina's tuples, I am not keen on anything that makes circuit descriptions "less nice".

jadephilipoom:
There _is_ a solution to the new extra type information: if we used `signals` instead of `signal` everywhere (e.g. define `xor2` to have the type `signals (ipair Bit Bit) -> cava (signals Bit)`, then we wouldn't need that extra type information. In that case I would rename `signal` to something like `interp` that's less easy to confuse with `signals`.

jadephilipoom:
Closing this one; it seems like it's not the right solution for us, and the issue isn't blocking.

PR #448, opened 2021-01-21
Make some AES state representations little-endian

jadephilipoom:
Thanks to @smore-lore for flagging this one!

There's a conflict between Coq and FIPS in terms of endianness: Coq (and Cava, by extension) expects little-endian bit vectors, and FIPS is all big-endian. Previously, the conversions in `StateTypeConversions` handled this by taking a little-endian flat bitvector and turning it into big-endian rows and columns, which was most convenient for working with the FIPS-based subroutine specifications. However, some of the type conversions are meant for _implementations_, which will use Cava/Coq and therefore are best left little-endian.

The `cols_bits` and `cols_bitvecs` representationsare used almost exclusively in implementation rather than specification (`add_round_key` uses this representation in specification, but since it's just an xor it doesn't care about endianness). So this PR converts them to little-endian representations to make implementation more convenient.

This PR also separates the type conversions into two modules `BigEndian` and `LittleEndian` that have to be separately imported, to make it clearer what representation is being used. Now, when you import `StateTypeConversions`, you have `BigEndian.to_cols` and `LittleEndian.to_cols_bitvecs` instead of `to_cols` and `to_cols_bitvecs`. You can `Import BIgEndian` or `Import LittleEndian`, but I'd recommend importing at most one of these to avoid confusion.

### DIFF
--- a/silveroak-opentitan/aes/Spec/AES256.v
+++ b/silveroak-opentitan/aes/Spec/AES256.v
@@ -27,14 +27,16 @@ Require Import AesSpec.Cipher.
 Require Import AesSpec.CommuteProperties.
 Require Import AesSpec.StateTypeConversions.
 
+Import StateTypeConversions.BigEndian.
+
 (* TODO: these definitions need to be filled in once the mixcolumns spec is finalized *)
 Axiom inverse_mix_columns :
   forall Nb (st : Vector.t _ Nb), inv_mix_columns (mix_columns st) = st.
 
 Axiom mix_columns_add_round_key_comm :
   forall (st k : Vector.t _ 4),
-    let to_bits x := to_cols_bits (from_cols x) in
-    let from_bits x := to_cols (from_cols_bits x) in
+    let to_bits x := LittleEndian.to_cols_bits (from_cols x) in
+    let from_bits x := to_cols (LittleEndian.from_cols_bits x) in
     add_round_key 32 4
                   (to_bits (inv_mix_columns st))
                   (to_bits (inv_mix_columns k))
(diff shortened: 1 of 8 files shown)


silveroak-opentitan/aes/Spec/StateTypeConversions.v
```diff
@@ -57,33 +60,41 @@ Section Conversions.
...
+     with bit vectors instead of bytes. *)
   Definition to_cols_bits (st : state) : Vector.t (Vector.t bool bits_per_word) Nb :=
-    let cols := to_cols st in
-    (* byte conversion expects little-endian form, so reverse each column and
-       then reverse back *)
-    map (fun c => reverse (bytevec_to_bitvec _ (reverse c))) cols.
+    let cols := BigEndian.to_cols st in
+    (* byte conversion expects little-endian form, so reverse each column before
+       converting to bits *)
+    let cols_bits := map (fun c => bytevec_to_bitvec _ (reverse c)) cols in
```
smore-lore, line 80: we have 'reverse' as a nicer version of Coq.Vectors.VectorDef's 'rev'? Is there any motivation (besides standard lib version of rev being heavy/complicated)?

smore-lore (approved):
Looks great, thank you!

silveroak-opentitan/aes/Spec/StateTypeConversions.v
jadephilipoom: Allow me to answer with a demonstration:
```coq
Compute (VectorDef.rev [1;2]%vector).
     (*      = match *)
     (*     match plus_n_O 2 in (_ = H) return (H = 2) with *)
     (*     | eq_refl => eq_refl *)
     (*     end in (_ = H) return (Vector.t nat H) *)
     (*   with *)
     (*   | eq_refl => *)
     (*       match *)
     (*         match plus_tail_plus 2 0 in (_ = H) return (H = 2) with *)
     (*         | eq_refl => eq_refl *)
     (*         end in (_ = H) return (Vector.t nat H) *)
     (*       with *)
     (*       | eq_refl => [2; 1] *)
     (*       end *)
     (*   end *)
     (* : VectorDef.t nat 2 *)


Compute (reverse [1;2]).
     (*      = [2; 1] *)
     (* : Vector.t nat 2 *)
```
The standard library version unfortunately puts proof terms into the code, which messes with computation times and makes it very difficult to reason about.

silveroak-opentitan/aes/Spec/AES256.v
```diff
@@ -27,14 +27,16 @@ Require Import AesSpec.Cipher.
 Require Import AesSpec.CommuteProperties.
 Require Import AesSpec.StateTypeConversions.
```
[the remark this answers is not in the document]
jadephilipoom: The `LittleEndian` module is actually inside `StateTypeConversions`; we need to import the latter to be able to import the former.

Issue #450, opened 2021-01-21
Define a mux2 library element that operates over pairs defined in terms of indexAt


satnam6502:
An attempt to rationnalize mux2-like things in the repo so we only have once instances of it which is a special case of the application of `indexAt`

PR #451, opened 2021-01-21
Define a mux2 library element that works over pairs

satnam6502:
@jadephilipoom Why do I now need to add `(A:=Bit)` in the `MuxExamples.v` tests file below?
A stepping stone towards rationalizing the instances of `mux2`-like things in our rep. #450


blaxill:
Perhaps remove the existing mux2 ?

satnam6502:
> Perhaps remove the existing mux2 ?

I was doing it a step at a time. I had planned to assign an issue to others to remove the existing muxes, update proofs etc.

satnam6502:
@jadephilipoom another note abut the test is `MuxExamples`. There are four test vectors that are used to test `mux2` with four separate test invocations. In a subsequent PR I plan to make this just one test invocation since now the `mux2` circuit can take a stream of inputs i.e. we can just supply the four test vectors in sequence to the circuit and check the four test outputs in one test.

blaxill (approved):
Lgtm!

jadephilipoom:
> @jadephilipoom Why do I now need to add (A:=Bit) in the MuxExamples.v tests file below?

The previous `mux2_1` was already specialized to `signal Bit`, while the new one is for any `SignalType`. To the Coq type unification, the two problems look like:

with `mux2_1` : `combType Bit = bool`
with generic `mux2` : `combType ?t = bool`

The first one can be resolved by just unfolding `combType`. For the second, though, Coq would have to figure out that it can instantiate the `?t` with `Bit`, which is a much harder problem. By passing the explicit argument you plug in `t` yourself. An alternate solution is that if `true` and `false` were coerced to `combType Bit`, the problem would be `combType ?t = combType Bit`, which Coq would also figure out.

satnam6502:
> > @jadephilipoom Why do I now need to add (A:=Bit) in the MuxExamples.v tests file below?
> 
> The previous `mux2_1` was already specialized to `signal Bit`, while the new one is for any `SignalType`. To the Coq type unification, the two problems look like:
> 
> with `mux2_1` : `combType Bit = bool`
> with generic `mux2` : `combType ?t = bool`
> 
> The first one can be resolved by just unfolding `combType`. For the second, though, Coq would have to figure out that it can instantiate the `?t` with `Bit`, which is a much harder problem. By passing the explicit argument you plug in `t` yourself. An alternate solution is that if `true` and `false` were coerced to `combType Bit`, the problem would be `combType ?t = combType Bit`, which Coq would also figure out.

Perhaps the way I should write this test is to first declare a specialized version of `muxPair`, say `mux2_1`, and then use it in the tests.

PR #452, opened 2021-01-21
Add Acorn mix_columns

blaxill:
#322

### DIFF
--- a/silveroak-opentitan/aes/Acorn/MixColumns.v
+++ b/silveroak-opentitan/aes/Acorn/MixColumns.v
@@ -86,8 +86,8 @@ Section WithCava.
   Definition aes_mix_columns
     (op_i : signal Bit) (a: signal (Vec (Vec (Vec Bit 8) 4) 4))
     : cava (signal (Vec (Vec (Vec Bit 8) 4) 4)) :=
-  let transposed := aes_transpose a in
-  mixed <- mapT (aes_mix_single_column op_i) (peel transposed) ;;
-  ret (aes_transpose (unpeel mixed)).
+    let transposed := aes_transpose a in
+    mixed <- mapT (aes_mix_single_column op_i) (peel transposed) ;;
+    ret (aes_transpose (unpeel mixed)).
 
 End WithCava.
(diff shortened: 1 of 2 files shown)


silveroak-opentitan/aes/Acorn/MixColumns.v
```diff
@@ -0,0 +1,93 @@
...
+
+Section WithCava.
+  Context {signal} {semantics : Cava signal}.
+  Context {monad: Monad cava}.
+
+  Definition zero_byte : cava (signal byte) :=
+    z <- zero ;;
+    ret (unpeel (z :: z :: z :: z :: z :: z :: z :: z :: [])).
+
+  Definition aes_mix_single_column
```
jadephilipoom, line 45: Could you add a few comments here explaining what's happening in this implementation? It will be helpful when trying to match it up with the spec.

jadephilipoom (approved):
Looks great! Maybe also add a test for this implementation similar to the [one for add_round_key](https://github.com/project-oak/silveroak/blob/435e7d5c897d626ee3b68867b4c91ab6b7a74e2c/silveroak-opentitan/aes/Acorn/AddRoundKey.v#L50)?

silveroak-opentitan/aes/Acorn/MixColumns.v
blaxill: Added original verilog and verilog comments. It didn't click with me before, but the OpenTitan version uses the equations from Satoh et al., "A Compact Rijndael Hardware Architecture with S-Box Optimization". If this is an issue for proof let me know and I can do a more spec following version.

silveroak-opentitan/aes/Acorn/Pkg.v
```diff
@@ -0,0 +1,80 @@
...
+Require Import AcornAes.Common.
+Require Import AesSpec.StateTypeConversions.
+Require Import AesSpec.Tests.CipherTest.
+Require Import AesSpec.Tests.Common.
+Import Common.Notations.
+
+Import VectorNotations.
+
+Local Notation byte := (Vec Bit 8) (only parsing).
+Local Notation "v [@ n ]" := (indexConst v n) (at level 1, format "v [@ n ]").
```
satnam6502, line 34: I'm wary of using `Notation` too much. Perhaps here there is a case for using it to express static indexing? If that is the case then we should think about standardizing it across the codebase and putting it into `CavaPrelude`, and perhaps having a discussion about it with I have initiated in #456

blaxill: I'll keep this for now and I'll update to whatever is decided in #456

PR #453, opened 2021-01-21
Remove now redundant mux2 from Combinators

satnam6502:
I had to move `mux2` from `Lib.Multiplexers` to a prelude file since it is now used "in the core".
Also rename `mux2` to `muxPair` which more accurately reflects what it is doing.
 #450

### DIFF
--- a/cava/Cava/Acorn/CavaPrelude.v
+++ b/cava/Cava/Acorn/CavaPrelude.v
@@ -21,17 +21,22 @@ Import VectorNotations.
 Require Import ExtLib.Structures.Monads.
 Import MonadNotation.
 
-Require Import Cava.Acorn.Acorn.
+Require Import Cava.Acorn.CavaClass.
+Require Import Cava.Signal.
 
 Section WithCava.
   Context {signal} `{Cava signal} `{Monad cava}.
 
-(* A two to one multiplexer that takes its two arguments as a pair rather
-   than as a 2 element vector which is what indexAt works over. *)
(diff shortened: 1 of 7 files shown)


cava/Cava/Acorn/CavaPrelude.v
```diff
@@ -21,17 +21,22 @@ Import VectorNotations.
...
-                (sel : signal Bit)
-                (ab : signal A * signal A) : cava (signal A) :=
+  (* A two to one multiplexer that takes its two arguments as a pair rather
+     than as a 2 element vector which is what indexAt works over. *)
+
+  Definition muxPair {A : SignalType}
+                     (sel : signal Bit)
+                     (ab : signal A * signal A) : cava (signal A) :=
   let (a, b) := ab in
   ret (indexAt (unpeel [a; b]) (unpeel [sel])).
```
smore-lore, line 41: Compared to the definition from Combinators.v 'ret (pairSel sel (mkpair (fst i) (snd i))).'

Then, should we remove pairSel from CavaClass? Or, lift it to be based on CavaClass, so that we avoid two similar primitives?

smore-lore (approved):

cava/Cava/Acorn/CavaPrelude.v
satnam6502: No, not yet, because `pairSel` works on our own AST tuples rather than Galina/Coq tuples (which is what `muxPair` works over). I am attempting to git rid of tuples from `SignalType` and when that occurs `pairSel` and `mkpair` and `unpair` will all be removed.

Issue #474, opened 2021-01-22
Replace IncrDecr with an implementation using synthesizable adders and subractors


satnam6502:
We should not need to build incrementers and decrementers for control logic out of basic gates. Let's just use the synthesizable versions.

jadephilipoom:
I'm fine with replacing it, but I thought we decided to put it in `acorn-examples` because it was an instructive example of a very simple circuit, even though it's not useful as a library function?

PR #475, opened 2021-01-22
Rework zero and one to be signal values

satnam6502:
This PR changes `zero` and `one` to be `signal` values and moves them out of `CavaClass` and into a prelude file.
This involved changes to `mix_cols` (take not @blaxill ).

There was one proof that I could not convert in `IncrDecr.v` (although there is a higher level issue here to do with the fact that we should be using a synthsizable adder/substractor here, but that's another issue #474). In the proof I manage I get the goal to:

```
combinational (decr' [true] [input]) =[N2Bv_sized sz (if (Bv2N input =? 0)%N then (2 ^ N.of_nat sz - 1)%N else (Bv2N input - 1)%N)]
```
 which seems to be set up for the application of `rewrite decr'_correct` but that rule does not seem to match to goal.
@jadephilipoom please can you look at this?

Addresses #475


cava/Cava/Acorn/Combinators.v
```diff
@@ -87,6 +88,12 @@ Section WithCava.
                        cava (signal A * (signal B * signal C)) :=
    let '((a, b), c) := i in
    ret (a, (b, c)).
+
+  Definition mux2 {A : SignalType}
```
blaxill, line 92: I think this has been accidentally re-added ?

satnam6502: Ugh! Let me get rid of it...

PR #477, opened 2021-01-22
Remove duplicate bitwise-XOR from Pkg.v

satnam6502:
There was a definition for a bit-wise-XOR in `Pkg.v` but we have one in the library `Cava.Lib.BitVectorOps`.
Delete the duplicate definition but also provide a curried version of it `xorv`.
At the same time change one of the vector literals which is constructed by consing into a single Vector literal like `[..; ..; ..]`.
CC: @blaxill


blaxill:
THanks missed that

PR #480, opened 2021-01-26
Simplify mux test

satnam6502:
With the new combinational simulation semantics we can test combinational circuits with a single `Example` test that works over a stream of values, with each value representing a test vector. Before we had to create separate `Example`s. This PR updates `MuxTest` to remove the single-shot tests and have just one `Example` test. It defines a helper function to map a list of tuples (convenient for cycle by cycle representation of test vectors) to a tuple of lists (which is what the Coq simulation semantics expects). Note question about why I need to explicitly give the `list SignalType` argument to `fromListOfTuples` in `MuxTest.v`: I had guessed this could have been left implicit and inferred from context.

CC: @blaxill

### DIFF
--- a/tests/MuxTests.v
+++ b/tests/MuxTests.v
@@ -36,6 +36,8 @@ Section WithCava.
   Definition mux2_1: signal Bit -> signal Bit * signal Bit -> cava (signal Bit)
   := muxPair.
 
+  Definition mux2_1_top '(sel, a, b) := mux2_1 sel (a, b).
+
   (* Mux between input buses *)
 
   Definition muxBus {dsz sz isz: nat}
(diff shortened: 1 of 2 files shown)


tests/MuxTests.v
```diff
@@ -53,26 +55,22 @@ Local Close Scope vector_scope.
...
+Definition mux2_1_tb_inputs :=
+  [(false, false, true);
+   (false, true,  false);
+   (true,  false, true);
+   (true,  true,  false)].
 
-Example m4: combinational (mux2_1 [false] ([true], [false])) = [true].
-Proof. reflexivity. Qed.
+(* Transpose test vectors for use in Coq simulation. *)
+(* Satnam asks: why can't [Bit; Bit; Bit] be inferred? *)
```
jadephilipoom, line 65: It's a similar problem to why the generic version of the mux needs a SignalType argument. Coq can't look at a Gallina type that `seqType` or `combType` creates and backsolve for what `SignalType` it came from (if it did this kind of speculative search by default, unification would probably be very slow!). The type unification problem here looks like: `(bool * bool * bool) = tupleInterfaceR combType ?ts`, which can't be solved by computation because the list of `SignalType`s, `?ts`, is unknown.

If you defined `mux2_1_tb_inputs` as a `list (tupleInterfaceR combType [Bit;Bit;Bit])` by specifying its type explicitly, then you wouldn't need to give the argument explicitly because the problem would then be `tupleInterfaceR combType [Bit;Bit;Bit] = tupleInterfaceR combType ?ts`.

jadephilipoom (approved):
Nice that these tests got simpler! A datapoint for thinking about even purely combinational circuits in a sequential way.

PR #481, opened 2021-01-26
Make synthsizable arithmetic operations expressions, not circuits

satnam6502:
I think it is much better to have the synthesizable arithmetic operations as expressions rather than instantiated circuits. Their types have also been changed to work over tuples to make them easier to use in point-free style (except when specializing on one input in which case the original curried version would have been better).

### DIFF
--- a/tests/TestMultiply.v
+++ b/tests/TestMultiply.v
@@ -34,7 +34,7 @@ Section WithCava.
                         (ab: signal (Vec Bit aSize) * signal (Vec Bit bSize)):
                         cava (signal (Vec Bit (aSize + bSize))) :=
     let (a, b) := ab in
-    unsignedMult a b.
+    ret (unsignedMult (a, b)).
 
 End WithCava.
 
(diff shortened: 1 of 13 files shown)


tests/TestMultiply.v
```diff
@@ -34,7 +34,7 @@ Section WithCava.
                         (ab: signal (Vec Bit aSize) * signal (Vec Bit bSize)):
                         cava (signal (Vec Bit (aSize + bSize))) :=
     let (a, b) := ab in
-    unsignedMult a b.
+    ret (unsignedMult (a, b)).
```
jadephilipoom, line 37: nit: shouldn't this just be `ret (unsignedMult ab)`?

jadephilipoom (approved):
LGTM!

tests/TestMultiply.v
satnam6502: Ah! Goof catch! Too much late night programming!

PR #482, opened 2021-01-26
Add loops with initial states

jadephilipoom:
Needed for #413 

Our current loop formulations don't work idiomatically for AES because the initial state is always the "default values" i.e. zero. However, the AES loop wants to start with a specified initial state and iterate from there. This PR adds a new loop definition and proof that handles initial states. (It's defined in terms of `loopDelayS`, so there are no new elements in `CavaClass`.)

The circuit diagram for the new loop looks like:
```
                      _______
               ______| delay |____________
              |      |_______|            |
              |       _______             |
              |   ___| delay |_________   |
              |  |   |_______|         |  |
              |  |                     |  |
              |  |                     |  |
              |  |_                  one  |
              |_|   \       _________     |
initial_state __|mux |-----|         |    |
                |___/      |  body   |____|
        input -------------|_________|
```
In short, the loop has two pieces of state, one of which is a single bit that is true for every cycle after the very first one. This bit is used to select the initial state for the first cycle and the feedback state for every cycle thereafter. Correct behavior relies on the default for `Bit` being `false`.

### DIFF
--- a/cava/Cava/Acorn/CombinationalProperties.v
+++ b/cava/Cava/Acorn/CombinationalProperties.v
@@ -277,6 +277,70 @@ Lemma pairAssoc_mkpair {A B C} a b c :
   @pairAssoc _ _ A B C (mkpair (mkpair [a] [b]) [c]) = mkpair [a] (mkpair [b] [c]).
 Proof. reflexivity. Qed.
 
+Lemma unpair_default {A B} :
+  unpair [defaultCombValue (Pair A B)] = ([defaultCombValue A], [defaultCombValue B]).
+Proof. reflexivity. Qed.
+
+Lemma unpair_nth {A B} (p : seqType (Pair A B)) n d :
+  List.nth n p d = (List.nth n (fst (unpair p)) (fst d),
+                    List.nth n (snd (unpair p)) (snd d)).
+Proof. apply split_nth. Qed.
+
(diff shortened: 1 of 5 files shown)


blaxill (commented):
IMO in the long run I think the methods of `CavaSeq` that cause a delay should take an (optional?) reset value- That way it corresponds directly to a reset in SystemVerilog, and then the `one >=> delay` and `mux` is unneeded (loopDelayS would take a default state argument instead). This PR approach is fine but deviates from what I'd expect in verilog- not sure what @satnam6502 has in mind.

jadephilipoom:
To summarize a conversation @blaxill and I just had: this implementation won't work well because, in practice, the streams of input will include multiple rounds of AES, so treating the very first cycle as "special" will only work for the first round. Instead, we should use the round index to select either the initial key/round constant or the feedback. Thanks for the catch, Ben!

satnam6502:
I had planned to add a `delay` variant that takes an initial value specified at compile time. Perhaps I should do that right now?

blaxill:
> I don't understand "this implementation won't work well because, in practice, the streams of input will include multiple rounds of AES, so treating the very first cycle as "special" will only work for the first round".

The cipher definition itself isn't in this PR. The combinational versions we have done before trivially work over multiple cycles. The looped version needed a small tweak to work with multiple invocations to allow the definition to be used for netlist generation. (Otherwise the core would hang after the first few cycles)

PR #485, opened 2021-01-26
Make cipher implementation use a loop

jadephilipoom:
Seeking feedback earlier than I did for #482, to make sure this top-level AES loop implementation is compatible with the hardware we want to end up with. The proofs are broken at the moment; I wanted to check about the implementation before starting work on them.

### DIFF
--- a/silveroak-opentitan/aes/Acorn/CipherRound.v
+++ b/silveroak-opentitan/aes/Acorn/CipherRound.v
@@ -31,7 +31,8 @@ Import Common.Notations.
 Local Open Scope monad_scope.
 
 Section WithCava.
-  Context {signal} {semantics : Cava signal} {monad: Monad cava}.
+  Context {signal} {semantics : Cava signal} {seqsemantics : CavaSeq semantics}
+          {monad: Monad cava}.
   Context {round_index round_constant : SignalType}.
 
   Context (sub_bytes:     signal Bit -> signal state -> cava (signal state))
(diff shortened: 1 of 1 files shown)


blaxill (approved):

silveroak-opentitan/aes/Acorn/CipherRound.v
```diff
@@ -96,16 +99,24 @@ Section WithCava.
...
     : cava (signal state) :=
-    '(_, _, out) <- foldLM (cipher_step is_decrypt) round_indices
-                          (initial_key, initial_rcon, input) ;;
-    ret out.
+    let initial_state := mkpair (mkpair initial_key initial_rcon) input in
+    loop_out <- loopDelayS
+                 (fun index_and_state : signal round_index
+                                      * signal (Pair (Pair key round_constant) state) =>
+                    let '(idx, feedback_state) := index_and_state in
+                    is_first_round <- idx ==? round_0 ;;
```
blaxill, line 115: This definition makes sense to me, this notation is using a cava definition for equality check right?

jadephilipoom: Yes, the notation is defined locally earlier in the file and just makes it a little easier to read.

satnam6502:
I'll accumulate some feedback to discuss with you verbally, and then follow up here to add notes afterwards.

jadephilipoom:
Closed in favor of #536

PR #486, opened 2021-01-27
Make registers take reset value

satnam6502:


satnam6502:
@jadephilipoom it's 1:27AM and I need to sleep! I've not quite finished this PR which I think just requires a few changes in `SequentialProperties.v` (and maybe a few other places). This PR adds an explicit reset value to registers in `CavaClass` and then defines in library modules the special cases where the registers have the default value after reset.

Please can you take over the PR and make the remaining changes and then merge if you think things are fine. Thank you.

cava/Cava/Acorn/CavaClass.v
```diff
@@ -94,23 +94,25 @@ Class Cava (signal : SignalType -> Type) := {
 
 (* Sequential semantics -- assumes the sequential part of the interpretation is in [signal] *)
 Class CavaSeq {signal : SignalType -> Type} (combinationalSemantics : Cava signal) := {
-  (* A unit delay. *)
-  delay : forall {t: SignalType}, signal t -> cava (signal t);
-  (* A unit delay with enable. *)
-  delayEnable : forall {t: SignalType}, signal Bit -> signal t -> cava (signal t);
+  (* A unit delay with a reset value. *)
+  delayWith : forall {t: SignalType}, combType t -> signal t -> cava (signal t);
```
jadephilipoom, line 98: Why `combType` here instead of `signal`?

cava/Cava/Acorn/NetlistGeneration.v
```diff
@@ -205,45 +205,61 @@ Definition sliceNet {t: SignalType} {sz: nat}
...
+  match t, v with
+  | Void, _ => UndefinedSignal
+  | Bit, bv => match bv with
+               | false => Gnd
+               | true => Vcc
+               end
+  | Vec vt s, v => VecLit (Vector.map (combToSignal vt) v)
+  | Pair ta tb, (a, b) => SignalPair (combToSignal ta a) (combToSignal tb b)
+  | ExternalType typ, _ => UninterpretedSignal typ
+  end.
```
jadephilipoom, line 218: If I'm interpreting this correctly, it can only represent constant values -- combinations of Vcc and Gnd. If the reset value needs to be a constant, then this particular kind of reset won't work directly for the AES loop, which needs to "reset" by pulling its state (which is keys, input data, etc) from a different place dynamically. However, we could use a constant reset value to select the state.

jadephilipoom (approved):

cava/Cava/Acorn/NetlistGeneration.v
satnam6502: Yes, exactly, by design the aim here for the reset value to be a compile time constant.

satnam6502:
Thank you very much!

cava/Cava/Acorn/CavaClass.v
satnam6502: Because I would like this to be a compile time constant.

PR #487, opened 2021-01-27
Helper lemmas

jadephilipoom:
These are helper lemmas from #482 that I'd like to add even though that PR is not getting merged. They're generally useful, so adding them will help us avoid duplicating work in the future.

### DIFF
--- a/cava/Cava/Acorn/SequentialProperties.v
+++ b/cava/Cava/Acorn/SequentialProperties.v
@@ -127,24 +127,66 @@ Section Overlap.
     reflexivity.
   Qed.
 
+  Lemma overlap_length {A} n x y :
+    length (@overlap A n x y) = Nat.max (length x) (length y + n).
+  Proof.
+    cbv [overlap]. autorewrite with push_length. lia.
+  Qed.
+  Hint Rewrite @overlap_length using solve [eauto] : push_length.
+
+  Lemma overlap_repeat {A} n m p (x : combType A) :
+    n <= m ->
+    overlap n (repeat x m) (repeat x p) = repeat x (Nat.max m (p + n)).
+  Proof.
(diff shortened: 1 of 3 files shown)


cava/Cava/Acorn/SequentialProperties.v
```diff
@@ -127,24 +127,66 @@ Section Overlap.
...
+  Proof.
+    intros. rewrite unpair_overlap_mkpair_r.
+    rewrite !extend_le by lia. reflexivity.
   Qed.
 End Overlap.
 Hint Rewrite @overlap_cons1 @overlap_cons2 @overlap_nil_r @overlap_snoc_cons
      using solve [length_hammer] : seqsimpl.
 Hint Rewrite @overlap_0_nil @overlap_app_same using solve [eauto] : seqsimpl.
 Hint Rewrite @skipn_overlap_same using solve [eauto] : push_skipn.
+Hint Rewrite @overlap_length using solve [eauto] : push_length.
```
blaxill, line 189: Add the other lemmas into hint databases too? Or not, you know best

blaxill (approved):
Lgtm

cava/Cava/Acorn/SequentialProperties.v
jadephilipoom: I try to keep hint databases predictable by giving them a clear task; for instance, the `push_length` hint database contains only lemmas that push `length` farther into the expression (e.g. `length (combine l1 l2) = Nat.min (length l1) (length l2)` pushes `length` because now `length` is applied to `l1` and `l2` instead of `combine l1 l2`). I haven't added the other lemmas to any hint databases because so far I only use them in very particular situations and don't have a clear "mission statement" in mind for a hint database I would add them to. Does that make sense?

Issue #458, opened 2021-01-22
Redefine one and zero to be overloaded SignalType values rather than circuits


satnam6502:
Remove `zero` and `one` from `CavaClass`.
Instead, define them as specific instances of `constant` in `CavaPrelude`:
```coq
Definition zero := constant false.
Definition one := constant true.
```
Consider giving them a different name e.g. `b0` and `b1`.
CC: @blaxill

PR #493, opened 2021-01-28
Set up SystemVerilog extraction and simulation for aes_mix_columns

satnam6502:
This PR illustrates sets up SystemVerilog extraction for `aes_mix_columns` but right now it only uses an empty test bench to just check that this block is compiles and is legal SystemVerilog and that it can be processed by Verilator.
Part of addressing issue #458

### DIFF
--- a/silveroak-opentitan/aes/Acorn/.gitignore
+++ b/silveroak-opentitan/aes/Acorn/.gitignore
@@ -1,0 +1,1 @@
+!AESSV.hs
--- a/cava/Cava2HDL.cabal
+++ b/cava/Cava2HDL.cabal
@@ -44,6 +44,7 @@ library
                      BinNums
                      BinInt
                      BitArithmetic
+                     BitVectorOps
                      Bool
                      Bvector
                      CavaClass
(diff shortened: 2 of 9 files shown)


silveroak-opentitan/aes/Acorn/.gitignore
```diff
@@ -0,0 +1 @@
+!AESSV.hs
```
smore-lore, line 1: It gets kind of messy to maintain output files next to these for this, and some other, reasons. Any thought we could move all output to a dedicated output directory? (this would also simplify 'make clean').

smore-lore (approved):
Perfect, thanks!

silveroak-opentitan/aes/Acorn/.gitignore
satnam6502: Yes, agreed, I think I would definitely like to find a better way to do this.
Several times I've been bitten by having a file I worked on for ages getting deleted by the git-remove :-(

blaxill: I also agree, Coq extraction is the difficult bit at the moment- not sure how to stop it dumping output files in the source location

Issue #468, opened 2021-01-22
FPGA implementation and test of mix_columns


satnam6502:
Specific steps (create issues for them)?

- [x] Declare netlist in Cava/Coq and extract SystemVerilog for just the`mix_columns` component.
- [x] Initially produce an empty test bench and extract also.
- [x] Makes sure the SystemVerilog compiles without issue by adding `mix_columns.sv` to the `SV` list in the `Makefile` (which will cause it to get extracted and Verilator to run).
- [ ] Delare one or two test vectors in Coq and use the spec of `mix_columns` to compute the expected oututs.
- [ ] Use these values to declare a testbench for `mix_columns`. Make the test PASS at the Verilator level.
- [ ] Take a copy of the testbench and add an instantiation of the OpenTitan `mix_columns` block so now both implementations are instantiated. Drive both circuits with the same inputs and ensure they produce the same outputs.
- [ ] Run Vivado on both circuits to get synthesis reports. Identify any performance regressions, and fix until at parity.
- [ ] Update `README.md` with text that describes these steps and representative output and VCD waveforms.

Likewise for the other blocks.

blaxill:
Aggregating in #601

PR #494, opened 2021-01-28
[aes_mix_columns] Coq and SystemVerilog tests for aes_mix_columns

satnam6502:
This PR adds Coq-level and SystemVerilog level tests for the Cava `mix_columns` circuit. The following things are tested:

* A Coq test with a test input from the Wikipedia page for AES, which checks the `mix_cols` spec gives the right answer (in `MixColumnsTest.v`). This `mix_cols` works over matrices of `Bytes.bye`.
* A Coq test using the same input which uses the `AES256.mix_columns` specification which works over a flat 128-bit vector.
* A Coq test that ensure when Cava `aes_mix_columns` top level circuit which is also given the same test vector and we check to make sure the result matches what the specifications compute.
* A SystemVerilog test for exactly the same test vector, which ensure the SystemVerilog circuit computes the same result as the Cava circuit in Coq (these tests check for any errors in the axioms for our semantics for mapping to SystemVerilog).

I've defined a Coq specification which corresponds to the SystemVerilog OpenTitan `aes_mix_columns` block:
```coq
  Definition aes_mix_columns_top_spec (op : bool) (i : Vector.t (Vector.t (Vector.t bool 8) 4) 4)
                                      : Vector.t (Vector.t (Vector.t bool 8) 4) 4 
```
Now there is a lemma which needs to be proved which shows the implementation is faithful to this specification:
```coq
Lemma aes_mix_columns_correct : forall (op : bool) (i :  Vector.t (Vector.t (Vector.t bool 8) 4) 4),
      combinational (aes_mix_columns [op] [i]) = [aes_mix_columns_top_spec op i].
```
The SystemVerilog simulation passes!

```
clang++ -L/usr/local/opt/sqlite/lib    aes_mix_columns_tb.o verilated.o verilated_vcd_c.o Vaes_mix_columns_tb__ALL.a    -o Vaes_mix_columns_tb -lm -lstdc++ 
obj_dir/Vaes_mix_columns_tb
                  10: tick = 0, op_i = 0, data_i[0] = 2013301392, data_i[1] = 2013301396, data_i[2] = 2013301400, data_i[3] = 2013301404, data_o[0] = 2013301408, data_o[1] = 2013301412, data_o[2] = 2013301416, data_o[3] = 2013301420
PASSED
```
I've separated tests and the top-level proof into the file `MixColumnsTest.v` to reduce how much extraction we need to do from Coq to Haskell.
Addresses part of #468 
Still to do:
* Synthesis and FGPA implementation to get resource usage and performance information for our mix_columns.
* Also do isolated synthesis for the OpenTitan mix_columns.
* Fix any performance regressions.
* Run mix_columns on FPGA and check correct operation using ILA (logic analyzer).
*  Splice in mix_columns in the OpenTitan AES tests and check tests still pass.
* Rebuild entire OpenTitan design with our mix_columns and run end-to-end AES test to check functionality.


silveroak-opentitan/aes/Acorn/MixColumnsCircuit.v
```diff
@@ -36,6 +38,8 @@ Import Common.Notations.
 
 Import VectorNotations.
 
+Require Import AesSpec.MixColumns.
```
jadephilipoom, line 41: Try importing `AesSpec.AES256` and using mix_columns from there; it includes the necessary type conversions from a flat input, so you can convert whatever representation you have into a 128-bit big-endian bitvector and call it, which might be easier to reason about.

silveroak-opentitan/aes/Acorn/MixColumnsCircuit.v
```diff
@@ -153,23 +157,74 @@ Definition aes_mix_columns_Interface :=
...
+  := Vector.map (Vector.map (fun v => N.to_nat (Bv2N (byte_to_bitvec v)))) i.
+
+Local Open Scope list_scope.
+
+(* list of nat state *)
+Definition mixColTestInputs := [mixColTest1Input].  
+(* list of state, for running spec *)
+Definition mixColTestInputs' := map fromNatState mixColTestInputs.
+(* list of combType state for testbench *)
+Definition mixColTestInputs'' : list (combType state) := map (Vector.map (Vector.map byte_to_bitvec)) mixColTestInputs'.
```
jadephilipoom, line 192: The test vector (like FIPS) is big-endian, while Coq's bitvectors and `byte_to_bitvec` are little-endian. Instead, you may want to use the conversions from `AesSpec.StateTypeConversions`, which package these sort of transformations. Since you start with a column-based representation in bytes, you can use `BigEndian.from_cols` to get to a flat vector and then either call the `AesSpec.AES256` version of `mix_columns` directly, or use `LittleEndian.to_cols_bits` to convert to the representation this version of `mix_columns` expects.

satnam6502: Looking at `AesSpec` what I don't see is a spec that has exactly the same interface as the circuit:
```coq
  Definition aes_mix_columns
    (op_i : signal Bit) (a: signal (Vec (Vec (Vec Bit 8) 4) 4))
    : cava (signal (Vec (Vec (Vec Bit 8) 4) 4)).
```
which is designed to match the SystemVerilog component:
```verilog
module aes_mix_columns (
  input  aes_pkg::ciph_op_e    op_i,
  input  logic [3:0][3:0][7:0] data_i,
  output logic [3:0][3:0][7:0] data_o
);
```
i.e. I don't want a flat 128-bit vector, it has to be a vector (size 4) of a vector (size 4) of an vector of bits of size 8.

satnam6502: I've added:
```coq
  Definition aes_mix_columns_top_spec (op : bool) (i : Vector.t (Vector.t (Vector.t bool 8) 4) 4)
                                      : Vector.t (Vector.t (Vector.t bool 8) 4) 4 
```

satnam6502:
Ready for review please.

silveroak-opentitan/aes/Acorn/MixColumnsCircuit.v
jadephilipoom: We can standardize all of `AesSpec.AES256` on the 3-layer representation -- a flat vector just seemed like the simplest option and matches FIPS test vectors. I'd rather change the representation everywhere than add more layers on top. But taking a step back here, why does it need to be exactly the same? Why not convert to a flat vector, call the spec, and convert back? I think that's a little clearer because the explicit conversion makes it crystal-clear exactly what representation of the flat vector you're working with (e.g. column-major and little-endian). See the `mix_columns` precondition of the equivalence proof:
￼
https://github.com/project-oak/silveroak/blob/9d0c5ce3f9e08f9a0981a0fa4af977e6ee9ab480/silveroak-opentitan/aes/Acorn/AES256Equivalence.v#L92

satnam6502: The interface of the OpenTitan circuit we are trying to produce does not take a flat-vector i.e. in SystemVerilog it is `logic [3:0][3:0][7:0]` so the corresponding circuit in Cava should have that interface. I think that kind of conversion to flat vector and back is what I do for producing the specification for the top level-block?
```coq
  Definition aes_mix_columns_top_spec (op : bool) (i : Vector.t (Vector.t (Vector.t bool 8) 4) 4)
                                      : Vector.t (Vector.t (Vector.t bool 8) 4) 4 :=
    let i_bytes := Vector.map (Vector.map bitvec_to_byte) (transpose i) in
    let i_big := BigEndian.from_rows (transpose i_bytes) in
    let bytes_o := BigEndian.to_cols (mix_columns i_big) in
    Vector.map (Vector.map byte_to_bitvec) bytes_o.
```

satnam6502: Very happy to consider alternative schemes, as long as the top-level circuit and spec are "isomorphic" to the SystemVerilog interface, and as long as I can provide _exactly_ the same inputs to the circuit or the spec and get back the same results, as I do in `MuxColumnsTests.v`.

jadephilipoom: Okay, if the `top_spec` is converting to/from flat vector then I'm mostly happy with it, except for the nitpick that it might be nicer if it uses the `AES256.mix_columns` so it matches the precondition I linked above? Then it will be more obvious that it matches our proofs (and more robust in case we change the representation used by the `mix_columns` spec).

silveroak-opentitan/aes/Acorn/MixColumnsCircuit.v
```diff
@@ -153,39 +155,40 @@ Definition aes_mix_columns_Interface :=
...
+(* A function to convert a bitvec matrix to a nat matrix. *)
+Definition toNatVec (i: Vector.t (Vector.t (Vector.t bool 8) 4) 4) : Vector.t (Vector.t nat 4) 4
+  := Vector.map (Vector.map (fun v => N.to_nat (Bv2N v))) i.
+
+Local Open Scope list_scope. 
+
+(* Get the test inputs into the right format for the circuit inputs. *)
+Definition mix_cols_i1 := fromNatVec mixColTest1InputNat.
+(* Compute the expected outputs from the Coq/Cava semantics. *)
+Definition mix_cols_expected_outputs := combinational (aes_mix_columns [false] [mix_cols_i1]).
```
jadephilipoom, line 188: These seem to me like they belong more in `MixColumnsTest`?

jadephilipoom (approved):
Great to see this up and running!

silveroak-opentitan/aes/Acorn/MixColumnsCircuit.v
satnam6502: Yes, indeed, but note they are needed to produce the test vector for the SystemVerilog testbench. I tried quite carefully to tease apart as much as I could into `MixColumnsTest` but this I think has to "stay with the circuit".

satnam6502:
Jade, thanks for cleaning up the structure of the testing: that's great.

PR #496, opened 2021-01-28
Extract combinational circuits

blaxill:
- Extraction for ShiftRows SboxLut SubBytes AddRoundKey
- Small change to SubBytes to allow extraction of subcomponent aes_sbox_lut
- Spec/AddRoundKey.v change VectorDef to Vector import to solve extraction issues

### DIFF
--- a/cava/Cava2HDL.cabal
+++ b/cava/Cava2HDL.cabal
@@ -56,6 +56,7 @@ library
                      HexString
                      IdentityMonad
                      List
+                     ListUtils
                      Logic
                      Monad
                      Netlist
@@ -85,7 +86,6 @@ library
                      Hexadecimal
                      Numeral
                      Specif
-                     ListUtils
                      MonadExc
                      MonadFix
                      MonadPlus
--- a/silveroak-opentitan/aes/Acorn/AESExtraction.v
+++ b/silveroak-opentitan/aes/Acorn/AESExtraction.v
@@ -16,6 +16,9 @@
 
 Require Import AcornAes.Pkg.
 Require Import AcornAes.MixColumns.
+Require Import AcornAes.ShiftRows.
+Require Import AcornAes.SubBytes.
+Require Import AcornAes.AddRoundKey.
 Require Import Coq.extraction.Extraction.
 Require Import Coq.extraction.ExtrHaskellZInteger.
 Require Import Coq.extraction.ExtrHaskellString.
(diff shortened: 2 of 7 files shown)


jadephilipoom (approved):

satnam6502:
Why is SubBytes so large?
```console
$ wc -l  SubBytes.hs
   27887 SubBytes.hs
```

jadephilipoom:
Probably the lookup table?

blaxill:
Yup this is due to using `nats` in the lookup table. Switching to `N` encoding should be smaller.

satnam6502:
> Yup this is due to using `nats` in the lookup table. Switching to `N` encoding should be smaller.

Please can you point out exactly where the `nat`s are in these lookup tables?
I'm pretty such I can fix any `nat`-based extraction issue. Related might be a plan to change the `lut`s in `CavaClass` to take an `N` value and then have the higher order function versions defined as library functions, although that might not be so convenient for proofs. Alternatively, provide both verisons.

blaxill:
In `aes/Acorn/SubBytes.v` sbox_fwd, the literal nats could be changed to N and just mapping with `N -> nat` would provide better extraction, as currently this:
~~~
Section SboxLut.
  Definition sbox_fwd : t nat _ :=
    [ 99; 124; 119; 123; 242; 107; 111; 197;
      48; 1; 103; 43; 254; 215; 171; 118;
~~~

map to loads of `Prelude.succ` in SubBytes.hs

~~~
sbox_fwd :: Vector.Coq_t Prelude.Integer
sbox_fwd =
  Vector.Coq_cons (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ
    (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ
    (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ
    (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ
    (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ
    (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ
    (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ (Prelude.succ
~~~

Note this doesn't have an impact on the verilog (as far as I can tell).

> Note that nat should get extracted to Integer in Haskell

My understanding this is correct and it is extracted as such **but** it doesn't extract literal value, it will always extract the constructors. E.g. changing nat -> N will result in extraction with binary constructors rather than peano constructors (still an improvement but not a constant literal)

Issue #295, opened 2020-10-23
Prove that inverse-mix-columns implements inverse of mix-columns


jadephilipoom:
Related to and partially dependent on #294

Once we have a specification for `mix_columns` and `inv_mix_columns`, we need to prove the usual inverse property: `forall x, inv_mix_columns (mix_columns x) = x`. This is potentially a complicated proof, because `mix_columns` has the most mathematically advanced reasoning of all the AES subroutines.

PR #504, opened 2021-02-02
Partial proofs for mix_columns

jadephilipoom:
Progress on #295

This PR contains the proof that `mix_columns` obeys the inverse property, with two lemmas left admitted for now (`byte_mul_distr_l` and `byte_mul_assoc` to be proved next). There's a lot of code churn because I moved all the things to do with generic polynomials out of the `MixColumns.v` file. Apologies for the huge diff -- `PolynomialProperties.v` is just a whole lot of proofs about generic polynomials and can probably just be skimmed.

- Various list/vector helper lemmas
- Proofs that generic polynomials form semi-rings (doesn't apply directly to the byte field because of the modular reduction)
- Proof that, given two admitted lemmas about the modular reduction in the AES byte field arithmetic, the byte field forms a semi-ring
- Proof that `inv_mix_columns` is the inverse of `mix_columns` using the above ring

One nice feature here is that the `inv_mix_columns` and `mix_columns` proofs can use Coq's built-in `ring` tactic, which solves tedious arithmetic goals really nicely.


blaxill (approved):
LGTM. 

- Perhaps more of the lemmas added here could be added as hints?
- Are the bytefield/polynomial modules general enough that we might want to package into a separate library at some point in the future? (Definitely not a priority)

jadephilipoom:
> * Perhaps more of the lemmas added here could be added as hints?

Definitely! There could certainly be more automation here, but since this isn't a core part of Cava I've been trying to prioritize getting it done over making it nice.

> * Are the bytefield/polynomial modules general enough that we might want to package into a separate library at some point in the future? (Definitely not a priority)

They certainly could be, but I think polynomial arithmetic on its own is not especially useful without a larger algebra library around it. I think the ideal scenario (assuming time and resources to do it) would be to get the code merged to some other library.

PR #508, opened 2021-02-03
Fix mix_columns inverse error

blaxill:
There was an indexing typo in mix_columns which only surfaces in the inverse mode. Also added a `aes_test_decrypt` check for inverse mode.


satnam6502 (approved):
Great to see the OpenTitan tests working for us!

Issue #466, opened 2021-01-22
FPGA implementation and test of sub_bytes


satnam6502:

blaxill:
- [x] - Netlist #496
- [ ] - Coq tests  
- [X] - Test in tree OpenTitan simulation with Verilator (working)

blaxill:
Aggregating in #601

PR #510, opened 2021-02-03
FPGA implementation of Silver Oak aes_sub_bytes

satnam6502:
This PR contains a tcl file for driving the FPGA implementation of the Silver Oak `aes_sub_bytes` block and a rule `impl` that invokes the Xilinx Vivado tool to run the implementation to produce utilization reports (but not bitstream generation). There is no clean to clean the generated output directory because we should be careful about wiping it since it will in general contain artifacts that took a long time to generate.

The comparison with the OpenTitan `aes_sub_bytes` block shows identical utilization.
Part of issue #466

### DIFF
--- a/silveroak-opentitan/aes/Acorn/.gitignore
+++ b/silveroak-opentitan/aes/Acorn/.gitignore
@@ -1,1 +1,2 @@
 !AESSV.hs
+!aes_sub_bytes.tcl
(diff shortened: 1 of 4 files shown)


blaxill (approved):
Lgtm! I'm unsurprised that the usage here is identical as its just indexing a dense table

PR #511, opened 2021-02-03
Full proofs of inverse_mix_columns

### LINKED ISSUE #295: Prove that inverse-mix-columns implements inverse of mix-columns

jadephilipoom:
Resolves #295 

This PR fills in the last two admits that were left in #504, which stated that _modular_ multiplication in the AES byte field was associative and distributive.

The strategy here is very computational; after investigating a bit and determining that proving the properties we needed in full generality (i.e. for all polynomials) would likely be very time-consuming, I took advantage of the fact that the AES field has a fixed modulus to write explicit expressions in terms of the input for modular reduction, multiplication, and addition in terms of GF(2) (bit) operations, and then used Coq's `ring` tactic to prove these equivalent.

Because of the computational strategy (which results in pretty large expressions), the proofs are unfortunately pretty slow. Compiling `MixColumns.v` takes about a minute on my laptop. The proofs have few dependencies, though, so they should be highly parallelizable and hopefully won't do too much damage to our CI time.

CC @cipher1024 The explicit formulas for byte multiplication/modulo might be helpful for `mix_columns` proofs.

### DIFF
--- a/silveroak-opentitan/aes/Spec/MixColumns.v
+++ b/silveroak-opentitan/aes/Spec/MixColumns.v
@@ -224,15 +224,6 @@ Section MixColumnsTests.
   Proof. vm_compute. reflexivity. Qed.
 End MixColumnsTests.
 
-Ltac fequal_list :=
-  repeat match goal with
-         | |- cons _ _ = cons _ _ => f_equal
-         end.
-Ltac fequal_vector :=
-  repeat match goal with
-         | |- Vector.cons _ _ _ _ = Vector.cons _ _ _ _ => f_equal
-         end.
-
 Section ByteFieldProperties.
   Existing Instances bitops byteops.
   Local Infix "*" := fmul.
--- a/silveroak-opentitan/aes/Spec/MixColumns.v
+++ b/silveroak-opentitan/aes/Spec/MixColumns.v
@@ -224,18 +224,32 @@ Section MixColumnsTests.
   Proof. vm_compute. reflexivity. Qed.
 End MixColumnsTests.
 
+Ltac fequal_list :=
+  repeat match goal with
+         | |- cons _ _ = cons _ _ => f_equal
+         end.
+Ltac fequal_vector :=
+  repeat match goal with
+         | |- Vector.cons _ _ _ _ = Vector.cons _ _ _ _ => f_equal
+         end.
+
 Section ByteFieldProperties.
   Existing Instances bitops byteops.
+  Local Infix "*" := fmul.
+  Local Infix "+" := fadd.
+  Local Infix "-" := fsub.
+
+  (* Declare a full ring because we need subtraction for some goals *)
+  Definition bit_theory : ring_theory (R:=bool) fzero fone fadd fmul fsub fopp eq
+    := BoolTheory.
+  Add Ring bitring : bit_theory.
+
+  (* Declare semi-ring for polynomial proof preconditions *)
   Definition BitTheory : semi_ring_theory (R:=bool) fzero fone fadd fmul eq.
   Proof.
     constructor; intros; cbn [fzero fone fadd fmul bitops];
       repeat match goal with x : bool |- _ => destruct x end; reflexivity.
   Qed.
-
-  (* This odd property holds on bytes because add/sub are xors *)
-  Lemma bytes_sub_is_add (a b : byte) :
-    @fadd _ byteops a b = @fadd _ byteops a b.
-  Proof. reflexivity. Qed.
 
   Lemma poly_to_byte_to_poly p :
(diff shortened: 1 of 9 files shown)


jadephilipoom:
Looks like the CI does parallelize this well enough to avoid a major performance hit -- I ran the CI twice, and got one run on the high end of our recent CI runtimes and one on the low end.

silveroak-opentitan/aes/Spec/MixColumns.v
```diff
@@ -224,19 +224,33 @@ Section MixColumnsTests.
   Proof. vm_compute. reflexivity. Qed.
 End MixColumnsTests.
 
+Ltac fequal_list :=
```
blaxill, line 227: Perhaps move these to cava/Tactics.v?

silveroak-opentitan/aes/Spec/MixColumns.v
```diff
@@ -224,19 +224,33 @@ Section MixColumnsTests.
...
+  Local Infix "+" := fadd.
+  Local Infix "-" := fsub.
+
+  (* Declare a full ring because we need subtraction for some goals *)
+  Definition bit_theory : ring_theory (R:=bool) fzero fone fadd fmul fsub fopp eq
+    := BoolTheory.
+  Add Ring bitring : bit_theory.
+
+  (* Declare semi-ring for polynomial proof preconditions *)
   Definition BitTheory : semi_ring_theory (R:=bool) fzero fone fadd fmul eq.
```
blaxill, line 248: I see this is referenced below, but is it needed anymore since we declare a full ring theory above? Perhaps the usage of `BitTheory` can be replaced by `bit_theory`.

(After reading more of the PR) ah perhaps it is still needed as ByteTheory is a semi-ring theory and uses semi-ring BitTheory?

silveroak-opentitan/aes/Spec/MixColumns.v
```diff
@@ -246,19 +260,281 @@ Section ByteFieldProperties.
...
+  Lemma add8_correct p q :
+    length p = 8%nat -> length q = 8%nat ->
+    add_poly p q = add8 p q.
+  Proof. intros; destruct_lists_by_length; reflexivity. Qed.
+
+  Local Ltac generalize_bytes_as_polynomials :=
+    repeat lazymatch goal with
+           | |- context [byte_to_poly ?b] =>
+             pose proof (byte_to_poly_length b);
+             generalize dependent (byte_to_poly b);
```
blaxill, line 502: Not important, but for my own education I'm curious what exactly is happening here. Any notes or links would be appreciated.

blaxill (approved):
Lgtm!

silveroak-opentitan/aes/Spec/MixColumns.v
jadephilipoom: Good catch, thanks!

silveroak-opentitan/aes/Spec/MixColumns.v
jadephilipoom: Yes -- annoyingly, the Coq standard library decided not to include (as far as I can tell by searching) anything that proves `ring_theory` implies `semi_ring_theory`. But since it's just bits, it's easy to prove them separately.

silveroak-opentitan/aes/Spec/MixColumns.v
jadephilipoom: The tactic converts a proof in terms of Coq's `byte` type into a proof in terms of length-8 `list bool`s. So, if I start with a goal like:
```
x, y : byte
===
foo (byte_to_poly x) = bar (byte_to_poly y)
```
I will end up with something like:
```
l1, l2 : list bool
H : length l1 = 8
H0 : length l2 = 8
===
foo l1 = bar l2
```
The procedure looks like:
1. find a subexpression in the goal that has the form `byte_to_poly _`
2. Assert that `length (byte_to_poly b) = 8` using the `byte_to_poly_length` lemma
3. "generalize" the `(byte_to_poly b)` expression, which means Coq replaces the expression with a new variable (information about how this variable was constructed is lost, which is why it's important here to assert the length first so we don't end up having to do a proof about a `list bool` of any length)

Here's the Coq manual entry on `generalize dependent` and `generalize` (a related tactic) if you want more detail : https://coq.inria.fr/refman/proof-engine/tactics.html#coq:tacn.generalize

PR #512, opened 2021-02-03
Prove that inv_mix_columns and add_round_key commute

### LINKED ISSUE #312: Prove that add_round_key and inv_mix_columns commute (with inverse-column-mixed key)

Once we have specifications for `add_round_key` and `inv_mix_columns`, we need to prove that they have the following property:
https://github.com/project-oak/oak-hardware/blob/5a049e139962088e5e210ff5775208b87d8fe00e/silveroak-opentitan/aes/Spec/Cipher.v#L117-L119

The property is true because both operations are linear on columns.

jadephilipoom:
Resolves #312 

With this PR, we are **completely done** with the proofs on top of the AES specification (i.e. the proofs that decryption is the inverse of encryption). The only proofs now remaining for the top-level theorems are proofs that the Cava implementations match the specification.

The diff is a bit confused because I de-indented the proofs that `shift_rows` and `sub_bytes` commute -- I didn't actually change any of the code there.

### DIFF
--- a/silveroak-opentitan/aes/Spec/AES256_Assumptions.out
+++ b/silveroak-opentitan/aes/Spec/AES256_Assumptions.out
@@ -1,14 +1,1 @@
-Axioms:
-mix_columns_add_round_key_comm
-  : forall st k : Vector.t (Vector.t Byte.byte 4) 4,
-    let to_bits :=
-      fun x : Vector.t (Vector.t Byte.byte 4) 4 =>
-      LittleEndian.to_cols_bits (from_cols x) in
-    let from_bits :=
-      fun x : Vector.t (Vector.t bool (4 * 8)) 4 =>
-      to_cols (LittleEndian.from_cols_bits x) in
-    AddRoundKey.add_round_key 32 4 (to_bits (MixColumns.inv_mix_columns st))
-      (to_bits (MixColumns.inv_mix_columns k)) =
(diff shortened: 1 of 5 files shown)


satnam6502:
Excellent to see such progress. As I was saying to @blaxill yesterday, things seem to be coming together on all fronts.

satnam6502 (approved):

silveroak-opentitan/aes/Spec/AES256_Assumptions.out
```diff
@@ -1,14 +1 @@
...
-      LittleEndian.to_cols_bits (from_cols x) in
-    let from_bits :=
-      fun x : Vector.t (Vector.t bool (4 * 8)) 4 =>
-      to_cols (LittleEndian.from_cols_bits x) in
-    AddRoundKey.add_round_key 32 4 (to_bits (MixColumns.inv_mix_columns st))
-      (to_bits (MixColumns.inv_mix_columns k)) =
-    to_bits
-      (MixColumns.inv_mix_columns
-         (from_bits (AddRoundKey.add_round_key 32 4 (to_bits st) (to_bits k))))
+Closed under the global context
```
jadephilipoom, line 1: No more axioms in the `Print Assumptions` for the top-level proof about the spec! :tada:

PR #515, opened 2021-02-03
[aes_mix_columns] Updated FPGA implementation report

satnam6502:
The updated report with the fix Silver Oak `aes_mix_columns` report shows that we are 4% larger which is not bad. I could investigate to see where this 4% comes from, but I think anything less than 5% is probably not worth worrying about too much.

PR includes documentation to show how to re-generate the reports using the `Makefile`.
Part of issue #468

### DIFF
--- a/silveroak-opentitan/aes/Acorn/.gitignore
+++ b/silveroak-opentitan/aes/Acorn/.gitignore
@@ -1,2 +1,2 @@
 !AESSV.hs
-!aes_sub_bytes.tcl
+!aes_util.tcl
--- a/silveroak-opentitan/aes/Acorn/Makefile
+++ b/silveroak-opentitan/aes/Acorn/Makefile
@@ -14,7 +14,21 @@
 # limitations under the License.
 #
 
-.PHONY: all coq minimize-requires clean _CoqProject impl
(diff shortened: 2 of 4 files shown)


satnam6502:
All the currently generated modules pass FPGA implementation except `aes_add_round_key` which has too many top-level IO ports for the chip which causes a utilization error. I'll provide a workaround for this in a subsequent PR.

blaxill (approved):
Ahh a little disappointing its larger now :) LGTM

satnam6502:
I've added another commit which uses out of context mode so we don't run into problems with not enough IOs for intermediate blocks. Now all the generated SystemVerilog fils implement and have been added to the `util` target.
```console
$ ls aes_implementation
aes_add_round_key  aes_mix_columns  aes_sbox_lut  aes_shift_rows  aes_sub_bytes
```

satnam6502:
> Ahh a little disappointing its larger now :) LGTM

If we were determined I guess we have a good chance of eliminating the difference.

satnam6502:
> Ahh a little disappointing its larger now :) LGTM

Also, note that utilization is not going to be "compositional". When this block is used as part of other blocks then that might unlock further optimization opportunities.

PR #516, opened 2021-02-03
Script and notes for running OpenTitan with our components

blaxill:
I'll perhaps add more automation with future scripts.

### DIFF
--- a/silveroak-opentitan/aes/copy_to_opentitan.sh
+++ b/silveroak-opentitan/aes/copy_to_opentitan.sh
@@ -1,0 +1,25 @@
+OPENTITAN_AES_DIR=../../third_party/opentitan/hw/ip/aes/rtl
+VERILATOR_CONFIG=../../third_party/opentitan/hw/lint/tools/verilator/common.vlt
+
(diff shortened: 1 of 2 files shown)


silveroak-opentitan/aes/copy_to_opentitan.sh
```diff
@@ -0,0 +1,25 @@
+OPENTITAN_AES_DIR=../../third_party/opentitan/hw/ip/aes/rtl
```
satnam6502, line 1: License header please.

silveroak-opentitan/aes/copy_to_opentitan.sh
```diff
@@ -0,0 +1,25 @@
...
+VERILATOR_CONFIG=../../third_party/opentitan/hw/lint/tools/verilator/common.vlt
```
satnam6502, line 2: For bash scripts consider the `-e` option so they fail on error.

satnam6502 (approved):
Great! It's _fantastic_ to see the end-to-end simulation test from the AES C program working.

PR #518, opened 2021-02-04
Pack monad instance into CavaClass

jadephilipoom:
Following up on a conversation with @satnam6502 

Previously, we've had section headers like:
```coq
Context {signal} {semantics : Cava signal} {monad: Monad cava}
```
This PR removes the need to specify the `Monad` instance by including it in `CavaClass` using a special syntax `:>` that makes typeclass inference pick up on the instance. The header changes to just:
```coq
Context {signal} {semantics : Cava signal}.
```

### DIFF
--- a/cava/Cava/Acorn/CavaClass.v
+++ b/cava/Cava/Acorn/CavaClass.v
@@ -14,7 +14,7 @@
 (* limitations under the License.                                           *)
 (****************************************************************************)
 
-
+Require Import ExtLib.Structures.Monad.
 Require Import Cava.Netlist.
 Require Import Cava.Signal.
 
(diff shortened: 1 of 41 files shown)


cava/Cava/Acorn/CavaClass.v
```diff
@@ -25,7 +25,8 @@ Local Open Scope type_scope.
    us to define both circuit netlist interpretations for the Cava class
    as well as behavioural interpretations for attributing semantics. *)
 Class Cava (signal : SignalType -> Type) := {
-  cava : Type -> Type;    
+  cava : Type -> Type;
+  monad :> Monad cava;
```
jadephilipoom, line 29: This is the special syntax; it creates an implicit `Existing Instance monad` whenever a Cava instance is declared.

satnam6502 (approved):
Fantastic!

PR #519, opened 2021-02-04
Remove pairSel from CavaClass

satnam6502:
Remove pairSel from CavaClass because it does not need an axiomatic definition because it can be defined in terms of vectors and vector indexing i.e. `indexAt`.
Some proofs will beed to be updated. @jadephilipoom please can you take care of that and if everything looks good then go ahead and merge? Thank you kindly.


jadephilipoom (approved):
Proofs should be fixed now!

satnam6502:
Thanks for fixing my vandalism.

PR #522, opened 2021-02-09
Merge sequential class into Cava

satnam6502:
@jadephilipoom  I'm trying to merge the combinational and sequential elements into a single class.

I run into trouble with `Example accumulatingAdderEnable2_ex1` in `AccumulatingAdderEnable.v`:
```coq
File "./AccumulatingAdderEnable/AccumulatingAdderEnable.v", line 136, characters 7-19:
Error: Unable to unify
 "map (fun i : nat => N2Bv_sized 8 (N.of_nat i)) [1; 2; 3; 4; 4; 4; 4; 4]"
with
 "semantics
    (accumulatingAdderEnableTop
       (combine
          (map (fun i : nat => N2Bv_sized 8 (N.of_nat i))
             [1; 1; 1; 1; 1; 1; 1; 1])
          (map nat2bool [1; 1; 1; 1; 0; 0; 0; 0])))".

make[2]: *** [AccumulatingAdderEnable/AccumulatingAdderEnable.vo] Error 1
```
presumably because I got the semantics of `loopDelaySEnableR` wrong? I tried to re-write the semantics of this as a regular Gallina function rather than literally translating into a circuit and taking its interpretation i.e.
```coq
Definition encodeLoopEnable
  {A B : SignalType}
  (en: seqType Bit)
  (f : seqType A * seqType B -> ident (seqType B))
  (i_state : seqType A * seqType B) :
  ident (seqType B) :=
    let state := snd i_state in
    let i := fst i_state in 
    let newState := unIdent (f (i, state)) in
    if en then ret newState else ret state.
```


cava/Cava/Acorn/CombinationalMonad.v
```diff
@@ -186,11 +280,18 @@ Instance CombinationalSemantics : Cava seqType :=
...
     instantiate _ circuit := circuit;
     blackBox intf _ := ret (tupleInterfaceDefaultS (map port_type (circuitOutputs intf)));
-}.
+    delayWith t d i := ret (d :: i);
+    delayEnableWith t d en i := delayEnableBoolList t d en i;
+    loopDelaySR _ _ := loopSeqS;
+    (* The semantics of loopDelaySEnable is defined in terms of loopDelayS and
+       the circuitry required to model a clock enable with a multiplexor. *)
+    loopDelaySEnableR _ _ resetValue en f input :=
+       loopSeqS resetValue (encodeLoopEnable en f) input;
```
jadephilipoom, line 289: @satnam6502 I think the problem you're running into stems from plugging in `en` here instead of including it as part of the loop state. Instead of peeling off one input from the `en` list each loop step, this will instead recycle the original `en` list and pull off the first element every time. Unfortunately I don't think there's a way around pairing the `en` and `input` here, because both need to be in the loop state.

satnam6502:
I am regretting separating out `en` like this!

cava/Cava/Acorn/CombinationalMonad.v
satnam6502: > @satnam6502 I think the problem you're running into stems from plugging in `en` here instead of including it as part of the loop state. Instead of peeling off one input from the `en` list each loop step, this will instead recycle the original `en` list and pull off the first element every time. Unfortunately I don't think there's a way around pairing the `en` and `input` here, because both need to be in the loop state.

I've decide just to transliterate what was there before (not that I am awake again!):
```coq
Definition encodeLoopEnable
  {A B : SignalType}
  (f : seqType A * seqType B -> ident (seqType B))
  (en_i_state : seqType (Pair Bit A) * seqType B) :
  ident (seqType B) :=
    let state := snd en_i_state in
    let (en, i) := split (fst en_i_state) in 
    let newState := unIdent (f (i, state)) in
    if en then ret newState else ret state.
```

PR #523, opened 2021-02-09
Correct and clear specifications for OT AES subcircuits

jadephilipoom:
In #494, we added a section to `Spec/AES256.v` with a conversion to and from the representation used by OpenTitan (which is actually row-major, not column-major as in FIPS) for use with tests. A few of our circuit-equivalence axioms (`shift_rows_equiv` and `mix_columns_equiv`) were therefore incorrect as stated; they assumed a column-major order. I said I'd follow up with something that made us use this conversion everywhere and clarify our circuit specs a bit, and this is that follow-up!

Changes:
- Added specifications for all subcircuits using the new `to_flat` and `from_flat` definitions to convert back and forth between OT's representation to the flat vector used by the specs
- Change axioms in `AESEquivalence` and `add_round_key` lemma statement to use the new specifications
- Repaired `add_round_key_equiv` proof (now that the implementation is row-major, it was slightly more complicated because it had a `transpose` in it, but I added some automation so the proof actually got shorter)
- Helper vector lemmas needed for `add_round_key_equiv`

This change reduces confusion about state representation and makes our equivalence proof statements a little cleaner; the axiom for `shift_rows` equivalence, for instance, went from:
```coq
Axiom shift_rows_equiv :
  forall (is_decrypt : bool) (st : combType state),
    unIdent (shift_rows [is_decrypt] [st])
    = [if is_decrypt
       then to_cols_bitvecs (AES256.inv_shift_rows (from_cols_bitvecs st))
       else to_cols_bitvecs (AES256.shift_rows (from_cols_bitvecs st))].
```
to
```coq
Axiom shift_rows_equiv :
  forall (is_decrypt : bool) (st : combType state),
    unIdent (shift_rows [is_decrypt] [st])
    = [AES256.aes_shift_rows_circuit_spec is_decrypt st].
```

@smore-lore @blaxill If you've already written lemma statements for the `sub_bytes` and `shift_rows` equivalence proofs based on the old axioms, you should update to the new ones. Sorry for the conflict!

### DIFF
--- a/cava/Cava/VectorUtils.v
+++ b/cava/Cava/VectorUtils.v
@@ -701,6 +701,14 @@ Section VectorFacts.
                    | reflexivity ].
   Qed.
 
+  Lemma reverse_map {A B} n (f : A -> B) v :
+    reverse (map (n:=n) f v) = map f (reverse v).
+  Proof.
+    induction n; intros; [ autorewrite with vsimpl; reflexivity | ].
+    rewrite (eta v). cbn [map reverse]. autorewrite with vsimpl.
+    rewrite IHn. rewrite snoc_map. reflexivity.
+  Qed.
+
   Lemma fold_left_andb_true (n : nat) :
     Vector.fold_left andb true (Vector.const true n) = true.
   Proof.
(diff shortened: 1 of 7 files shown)


blaxill (approved):
Lgtm!

Issue #462, opened 2021-01-22
Prove that mix_columns implementation matches spec


jadephilipoom:
After #461 is complete, we should prove that `mix_columns` matches its spec and fill in the axioms:

https://github.com/project-oak/silveroak/blob/633cabbaf673e5af73966c7ea2accc779f2b955a/silveroak-opentitan/aes/Acorn/AES256Equivalence.v#L51-L52
https://github.com/project-oak/silveroak/blob/633cabbaf673e5af73966c7ea2accc779f2b955a/silveroak-opentitan/aes/Acorn/AES256Equivalence.v#L90-L95

jadephilipoom:
Resolved by #535!

PR #525, opened 2021-02-09
Skeleton for mix_columns equivalence proof

jadephilipoom:
Progress on #462 

Prove `aes_mix_columns` circuit equivalent to spec in terms of two admitted lemmas about `transpose` and `aes_mix_single_column`.

@cipher1024 The `aes_transpose_correct` admit is the lemma we've talked about as a good entry point -- I'll leave it to you!


blaxill (approved):
Lgtm thanks Jade

Issue #526, opened 2021-02-09
Fix naming and organization in silveroak-opentitan/aes/Acorn


jadephilipoom:
Currently the naming and organization for the Acorn AES circuits is a bit chaotic. In particular:

- Circuit names are `add_round_key`, `sub_bytes`, `aes_mix_columns`, and `aes_shift_rows`; we should decide whether we want `aes_` prefix or not
- The circuit and tests are in separate files for `mix_columns` (`MixColumnsCircuit.v` and `MixColumnsTests.v`) but not for any other subcircuit
- The file containing the top-level cipher is called `CipherRound.v` but actually contains the full cipher definition
- We're missing a test for the inverse direction of `shift_rows`
- We only use the `AES256.from_flat` and `AES256.to_flat` conversions in some of the tests but ad-hoc transformations in others

For most of these, I think it matters more that we have a convention than what the convention is. However, this is how I would resolve them if I were to do it right now:
- Add `aes_` prefix to `sub_bytes` and `add_round_key`
- Create new `*Tests.v` files for all the subcircuits and move tests there + rename circuit files to `*Circuit.v`
- Rename `CipherRound.v` to `Cipher.v`
- Add inverse test for `shift_rows`
- Use  `AES256.from_flat` and `AES256.to_flat` conversions everywhere

Any thoughts/opinions?

jadephilipoom:
#527 fixed the issues with tests, leaving the `aes_` prefix and `CipherRound.v` naming still to be resolved.

blaxill:
#532

PR #527, opened 2021-02-09
Separate tests from circuits and make tests more uniform

jadephilipoom:
Progress towards #526 

This addresses a couple of points raised in #526, namely that `mix_columns` circuit implementation and proofs are separated out into `MixColumnsCircuit.v` and `MixColumnsTests.v` but the other AES subcomponents just have one file with both the circuit and tests, and the issue that `shift_rows` was missing an inverse-direction test. It also makes the test structure more uniform throughout the Acorn directory.

In this PR, each subcomponent has three files:
- `XCircuit.v` : Cava circuit definition
- `XNetlist.v` : netlist definition and test bench (this is the file that gets extracted)
- `XTests.v` : tests (including FIPS test and any additional ones)


blaxill (approved):
CI is currently failing on Haskell imports, I believe the `*Circuit` modules need to be added to AESExtraction.v and then it LGTM

silveroak-opentitan/aes/Acorn/SubBytesTests.v
```diff
@@ -0,0 +1,54 @@
...
+(* Test against FIPS test vectors *)
+Section FIPSTests.
+  (* Create a version of AES with the sub_bytes circuit plugged in *)
+  Let impl : AESStep -> Vector.t bool 128 -> Vector.t bool 128 -> Vector.t bool 128 :=
+    (fun step key =>
+       match step with
+       | SubBytes =>
+         fun st =>
+           let input := from_flat st in
+           let output := unIdent (sub_bytes [false]%list [input]%list) in
```
satnam6502, line 37: `Local Open Scope list_scope` to get rid of the annotations?

silveroak-opentitan/aes/Acorn/SubBytesTests.v
```diff
@@ -0,0 +1,54 @@
...
+           to_flat (List.hd (defaultCombValue _) output)
```
satnam6502, line 38: Perhaps `defaultCombValue` should have an implicit argument for the signal type. [No action required.]

satnam6502 (approved):

satnam6502:
Thanks. It's good to have such systematic naming here.

silveroak-opentitan/aes/Acorn/SubBytesTests.v
jadephilipoom: Agreed, I think in most places the type can be inferred.

PR #529, opened 2021-02-10
aes_sbox_lut performance comparison

satnam6502:
Issue #466


blaxill (approved):
Lgtm! Hopefully this should pass LEC too

satnam6502:
> Lgtm! Hopefully this should pass LEC too

It's a "P2" but I expect the "data-path" components to pass LEC (easily) although the top-level control will be another kettle of fish.

PR #530, opened 2021-02-10
Plug in mix-columns and shift-rows implementations

jadephilipoom:
Equivalence proofs are still admitted, but since we now have implementations for `mix_columns` and `shift_rows` I removed the axioms in `AESEquivalence.v`. The equivalence proofs are still admitted, but at least we have some assurance that the types match and there are no more proof-slowness problems around unfolding the implementations like we had with `sub_bytes`.


blaxill (approved):
Lgtm!

PR #531, opened 2021-02-10
Prove shift_rows circuit equivalent to spec

### LINKED ISSUE #460: Prove shift_rows implementation matches spec
Now that we have a `shift_rows` implementation, we should prove it matches the spec and fill in the axioms here:

https://github.com/project-oak/silveroak/blob/633cabbaf673e5af73966c7ea2accc779f2b955a/silveroak-opentitan/aes/Acorn/AES256Equivalence.v#L49-L50
https://github.com/project-oak/silveroak/blob/633cabbaf673e5af73966c7ea2accc779f2b955a/silveroak-opentitan/aes/Acorn/AES256Equivalence.v#L84-L89

jadephilipoom:
Resolves #460

### DIFF
--- a/cava/Cava/Acorn/CombinationalProperties.v
+++ b/cava/Cava/Acorn/CombinationalProperties.v
@@ -661,3 +661,7 @@ Proof.
     | _ => reflexivity
     end.
 Qed.
+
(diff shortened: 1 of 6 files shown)


blaxill (approved):
Lgtm thanks Jade

PR #532, opened 2021-02-10
Additional cleanup for #526

blaxill:
@jadephilipoom let me know if I've missed something from #526


jadephilipoom (approved):
Lgtm, thanks!

Issue #463, opened 2021-01-22
Prove that sub_bytes implementation matches spec


jadephilipoom:
Once #320 is complete, we should prove that `sub_bytes` matches the spec and fill in the axioms:

https://github.com/project-oak/silveroak/blob/633cabbaf673e5af73966c7ea2accc779f2b955a/silveroak-opentitan/aes/Acorn/AES256Equivalence.v#L47-L48
https://github.com/project-oak/silveroak/blob/633cabbaf673e5af73966c7ea2accc779f2b955a/silveroak-opentitan/aes/Acorn/AES256Equivalence.v#L78-L83

PR #534, opened 2021-02-11
SubBytes Equivalence

smore-lore:
This replaces an axiom in AES256Equivalence for the equivalence of
SubBytes with a proof.

Closes: #463

### DIFF
--- a/silveroak-opentitan/aes/Acorn/SubBytesEquivalence.v
+++ b/silveroak-opentitan/aes/Acorn/SubBytesEquivalence.v
@@ -1,0 +1,182 @@
+(****************************************************************************)
+(* Copyright 2021 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
+(* you may not use this file except in compliance with the License.         *)
+(* You may obtain a copy of the License at                                  *)
+(*                                                                          *)
+(*     http://www.apache.org/licenses/LICENSE-2.0                           *)
+(*                                                                          *)
+(* Unless required by applicable law or agreed to in writing, software      *)
+(* distributed under the License is distributed on an "AS IS" BASIS,        *)
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+Require Import Cava.Acorn.Acorn.
+Require Import Cava.Acorn.CombinationalProperties.
+Require Import Cava.Acorn.Identity.
+Require Import Cava.Acorn.MonadFacts.
+Require Import Cava.BitArithmetic.
+Require Import Cava.Lib.BitVectorOps.
+Require Import Cava.ListUtils.
+Require Import Cava.Tactics.
+Require Import Cava.VectorUtils.
+Require Import Coq.Lists.List.
+Require Import Coq.Vectors.Vector.
+Require Import ExtLib.Structures.Monads.
+Import VectorNotations.
+Close Scope vector_scope.
+Import ListNotations.
+
+Require Import AesSpec.AES256.
+Require Import AesSpec.StateTypeConversions.
+Require Import AcornAes.SubBytesCircuit.
+Import StateTypeConversions.LittleEndian.
+
+Existing Instance CombinationalSemantics.
+
+Section Equivalence.
+  Local Notation byte := (Vector.t bool 8).
+  Local Notation state := (Vector.t (Vector.t byte 4) 4) (only parsing).
+  Local Notation key := (Vector.t (Vector.t byte 4) 4) (only parsing).
+
+  Lemma sub_bytes_fwd_bytewise:
+    forall (b : byte),
+    unIdent (aes_sbox_lut [false] [b]) = [byte_to_bitvec (Sbox.forward_sbox (bitvec_to_byte b))].
+  Proof.
+    intros.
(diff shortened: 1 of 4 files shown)


silveroak-opentitan/aes/Acorn/SubBytesEquivalence.v
```diff
@@ -0,0 +1,196 @@
...
+    cbv [from_flat
+         to_flat
+         BigEndian.from_list_rows
+         BigEndian.to_list_rows
+         BigEndian.from_rows
+         BigEndian.to_rows
+         BigEndian.to_cols
+         BigEndian.from_cols
+         BigEndian.to_big_endian_bytes
+         BigEndian.from_big_endian_bytes].
```
jadephilipoom, line 148: Without knowing what exactly gets unfolded here, I think it's possible that you could unfold only a few of these and then use `autorewrite with conversions` to get rid of ones that cancel out (the `conversions` database contains rewrites like `to_rows (from_rows x) = x`).

silveroak-opentitan/aes/Acorn/SubBytesEquivalence.v
```diff
@@ -0,0 +1,196 @@
...
+    rewrite ! reverse_reverse.
+    rewrite ! reshape_flatten.
+    rewrite ! transpose_involutive.
+
+    constant_vector_simpl st.
+    constant_vector_simpl x.
+    constant_vector_simpl x0.
+    constant_vector_simpl x1.
+    constant_vector_simpl x2.
+    clear.
```
jadephilipoom, line 159: The `shift_rows` equivalence proof has my stab at eliminating the autogenerated names, in case it's useful:

https://github.com/project-oak/silveroak/blob/0aee628787e3c667d554490d31b6b7640e3538a8/silveroak-opentitan/aes/Acorn/ShiftRowsEquivalence.v#L52-L59

jadephilipoom (approved):
Thanks for the proof! I'll check it out and see if I have any ideas to fix the rewriting slowness.

jadephilipoom:
I think the slowness was coming from computing the s-boxes just because they're huge tables. I was able to fix it by using `map_ext` rewrites to apply the `_bytewise` lemmas earlier on and change `aes_sbox_lut` to `Sbox.forward_sbox` and `Sbox.inverse_sbox`, and then the performance issues disappear:

https://github.com/jadephilipoom/oak-hardware/commit/7e4b9ad5d4b7e2ea626fd6e1b17f670d86a8ab8b

Let me know if you want me to push this commit to your branch, or just feel free to copy over the strategy!

silveroak-opentitan/aes/Acorn/SubBytesEquivalence.v
smore-lore: Thanks! That's much better.

silveroak-opentitan/aes/Acorn/SubBytesEquivalence.v
smore-lore: Woah cool. I made a slight change to this - lemino what you think (I'm not sure about performance, and mine will be doing more work in certain cases).

smore-lore:
> I think the slowness was coming from computing the s-boxes just because they're huge tables. I was able to fix it by using `map_ext` rewrites to apply the `_bytewise` lemmas earlier on and change `aes_sbox_lut` to `Sbox.forward_sbox` and `Sbox.inverse_sbox`, and then the performance issues disappear:
> 
> [jadephilipoom@7e4b9ad](https://github.com/jadephilipoom/oak-hardware/commit/7e4b9ad5d4b7e2ea626fd6e1b17f670d86a8ab8b)
> 
> Let me know if you want me to push this commit to your branch, or just feel free to copy over the strategy!

Jade, thank you so much! I will definitely be adding this match trick, to apply lemmas deeply, to my toolbox. Though in this case,  with your other improvements here, I was able to reorder things to avoid that statement. Still - nice!

silveroak-opentitan/aes/Acorn/SubBytesEquivalence.v
```diff
@@ -0,0 +1,182 @@
...
+         ((if is_decrypt then Sbox.inverse_sbox else Sbox.forward_sbox)
+            (bitvec_to_byte b))].
+  Proof.
+    destruct is_decrypt; auto using sub_bytes_fwd_bytewise, sub_bytes_inv_bytewise.
+  Qed.
+
+  Lemma opaque_byte : byte.
+  Proof.
+      exact (byte_to_bitvec Byte.x00).
+  Qed.
```
smore-lore, line 83: I'm pretty happy with this. Of course, one other alternative is do make this section parameterized on a byte, and then instantiate it on an opaque/non-opaque byte when we actually use it.

Still - very happy with this "you must not care about the values of don't cares". Of course, this prevents certain proof strategies, but they aren't needed here.

silveroak-opentitan/aes/Acorn/SubBytesEquivalence.v
jadephilipoom: Nice! I think yours is better :slightly_smiling_face:

silveroak-opentitan/aes/Acorn/SubBytesEquivalence.v
jadephilipoom: It's a cool way to avoid introducing a new variable (and then later having to pass a dummy value to `rewrite`) while still enforcing that the value doesn't matter! I think the tradeoff here is that if you ever need to use the lemma in a place with a specific _other_ value plugged in as the default for `List.hd`, then the lemma won't apply.

PR #535, opened 2021-02-11
Mix single column correct

cipher1024:


jadephilipoom:
Hmm, I think if you're getting down to the nuts and bolts of `xor` here we've probably lost some of the  algebraic structure that would make this proof easier. I think the ideal way I imagine this proof could break down is:
- get a formula for each element of the vector in terms of `aes_mul2` etc. on the LHS and byte-field `fmul` on the RHS
- use lemmas about `aes_mul2` and its cousins to transform it into byte-field `fmul` and `fadd` expressions
- use the `fequal_vector` tactic to get a separate goal for each element of the vector
- Use Coq's `ring` tactic with the byte-field ring to solve the goals

If you'd like I could take a stab at this, and factor out the lemmas we need for `aes_mul2` etc?

cipher1024:
Thanks for the pointers! I'll give it another try. I did look for lemmas about `aes_mul2` but I didn't find them. Can you point them to me?

jadephilipoom:
There are currently no lemmas about `aes_mul2` since only `mix_columns` uses it. So figuring out what shape those lemmas need to take will be part of the proof!

jadephilipoom (approved):
LGTM, thanks!

silveroak-opentitan/aes/Acorn/FullCipherEquiv_Assumptions.out
```diff
@@ -1,11 +1 @@
...
-  : forall (is_decrypt : bool) (col : t (t bool 8) 4),
-    unIdent (aes_mix_single_column is_decrypt col) =
-    (if is_decrypt
-     then
-      Vector.map byte_to_bitvec
-        (MixColumns.inv_mix_single_column (Vector.map bitvec_to_byte col))
-     else
-      Vector.map byte_to_bitvec
-        (MixColumns.mix_single_column (Vector.map bitvec_to_byte col)))
+Closed under the global context
```
jadephilipoom, line 1: This is awesome to see! :tada:

PR #536, opened 2021-02-11
Add cipher based on loopDelayS

blaxill:
An updated version of #485

I had to add more mkpair/unpair'ing, `key'` etc are used to avoid `key` etc notation.

### DIFF
--- a/silveroak-opentitan/aes/Acorn/Cipher.v
+++ b/silveroak-opentitan/aes/Acorn/Cipher.v
@@ -108,4 +108,29 @@ Section WithCava.
                           (initial_key, initial_rcon, input) ;;
     ret out.
 
+  Definition cipher_inner_loop
+             (is_decrypt : signal Bit) (* called op_i in OpenTitan *)
+             (initial_state : signal (Pair (Pair key round_constant) state))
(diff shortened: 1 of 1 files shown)


silveroak-opentitan/aes/Acorn/Cipher.v
```diff
@@ -108,4 +108,29 @@ Section WithCava.
...
+  Definition cipher_inner_loop
+             (is_decrypt : signal Bit) (* called op_i in OpenTitan *)
+             (initial_state : signal (Pair (Pair key round_constant) state))
+             (index_and_state : signal round_index * signal (Pair (Pair key round_constant) state))
+    : cava (signal (Pair (Pair key round_constant) state)) :=
+    let '(idx, feedback_state) := index_and_state in
+    is_first_round <- idx ==? round_0 ;;
+    cipher_state <- muxPair is_first_round (initial_state, feedback_state) ;;
+    let '(key_round, state') := unpair cipher_state in
+    let '(key', round) := unpair key_round in
```
jadephilipoom, line 120: The convention I've been using to avoid `key`/`state`  collisions is to call keys `k` or `round_key` and states `st` or `data`.

jadephilipoom (approved):
Lgtm!

Issue #469, opened 2021-01-22
Final top level AES end-to-end execution on FPGA of Silver Oak AES


satnam6502:
The final FPGA implementation test where the top-level AES is shown to be running correctly on the FPGA hardware. After which, we're done with FPGA implementation work for Phase 1!

blaxill:
Closed in #624

PR #538, opened 2021-02-11
Performance comparison of OT top-level block vs. Silver Oak

satnam6502:
Issue #469

### DIFF
--- a/silveroak-opentitan/aes/Acorn/README.md
+++ b/silveroak-opentitan/aes/Acorn/README.md
@@ -309,3 +309,121 @@ which performs a bit-wise XOR between 4x4 bytes. We're confident this is directl
   xor inst_6 (net[122], key_i[3][3][2], data_i[3][3][2]);
 ...  
 ```
+
+## Top level Earl Grey circuit
+### Open Titan baseline Earl Grey block
+The complete FPGA Earl Grey FPGA design for the Nexys Video board is 26651 LUTs in size. Specificaslly, after synthesis the utilization is:
+```
+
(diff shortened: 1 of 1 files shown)


blaxill (approved):
LGTM

> For the final analysis we need to complete the implementation fo the AES control block.

FWIW, at least initially we won't support 128/192 AES (only 256) which should mean we will use less resources.

satnam6502:
> LGTM
> 
> > For the final analysis we need to complete the implementation fo the AES control block.
> 
> FWIW, at least initially we won't support 128/192 AES (only 256) which should mean we will use less resources.

Noted. I keep forgetting about this.

satnam6502:
W

> > LGTM
> > > For the final analysis we need to complete the implementation fo the AES control block.
> > 
> > 
> > FWIW, at least initially we won't support 128/192 AES (only 256) which should mean we will use less resources.
> 
> Noted. I keep forgetting about this.

Perhaps we can easily tweak the OT design to remove the 128/192 behviour.

blaxill:
> W
> 
> > > LGTM
> > > > For the final analysis we need to complete the implementation fo the AES control block.
> > > 
> > > 
> > > FWIW, at least initially we won't support 128/192 AES (only 256) which should mean we will use less resources.
> > 
> > 
> > Noted. I keep forgetting about this.
> 
> Perhaps we can easily tweak the OT design to remove the 128/192 behviour.

The 192 can be disabled through a SystemVerilog parameter (at least the parameter is there, I haven;t tested), 128 would require modification. I'll also try to estimate how much additional work for us to support 128/192

PR #539, opened 2021-02-11
Add assumption-tracking files back to version control

jadephilipoom:
I completely missed in reviewing #532 that these files were removed. They are autogenerated files, but my argument for keeping them version-controlled is that they're a good way to track exactly what assumptions our top-level theorems depend on. And if some change introduces a new axiom (for example, one of the axioms provided by the standard library in certain imports like `Classical`), it will automatically show up here and raise the proper red flags!


blaxill (approved):
My bad sorry, LGTM

Issue #489, opened 2021-01-27
Create top level Acorn aes_cipher_core


blaxill:
- [ ] - `aes_cipher_core` without status signals possibly required by OpenTitan, no decryption - #540 
- [ ] - Some form of test
- [ ] - `aes_cipher_core` with all OpenTitan required signals tested with `aes_test.c`, no decryption
- [ ] - Decryption tested with `aes_test.c`

PR #540, opened 2021-02-12
Add `aes_cipher_core` without additional status signals

blaxill:
- Additional status signals might be required for OpenTitan
- Inverse/decryption mode currently not operational

Part of #489

### DIFF
--- a/silveroak-opentitan/aes/Acorn/Cipher.v
+++ b/silveroak-opentitan/aes/Acorn/Cipher.v
@@ -162,17 +162,20 @@ Section WithCava.
              (input_and_state : signal cipher_signals * signal cipher_state)
     : cava (signal cipher_state) :=
     let '(input, feedback_state) := input_and_state in
-    let '(is_decrypt, num_rounds_regular, round_0, initial_state, idx) := unpacked_signal input in
+    let '(input, idx) := unpair input in
+    let '(input, initial_state) := unpair input in
+    let '(input, round_0) := unpair input in
+    let '(is_decrypt, num_rounds_regular) := unpair input in
     is_first_round <- idx ==? round_0 ;;
-    cipher_state' <- muxPair is_first_round (initial_state, feedback_state) ;;
-    let '(k, round, st) := unpacked_signal cipher_state' in
-    '(k, round, st) <- cipher_step num_rounds_regular round_0 is_decrypt
-                                   (k, round, st) idx ;;
-    ret (packed_signal cipher_state (k, round, st)).
+    cipher_state <- muxPair is_first_round (initial_state, feedback_state) ;;
+    let '(key_round, st) := unpair cipher_state in
+    let '(k, round) := unpair key_round in
+    '(key_round, st) <- cipher_step num_rounds_regular round_0 is_decrypt
+                                    (k, round, st) idx ;;
+    ret (mkpair (mkpair (fst key_round) (snd key_round)) st).
 
   Context {seqsemantics : CavaSeq semantics}.
-  Context (initial_rcon_forward: signal round_constant). (* #1 *)
-  Context (initial_rcon_backward: signal round_constant). (* #64 *)
+  Context (initial_rcon: signal Bit -> cava (signal round_constant)).
 
   Definition cipher_new
              (num_rounds_regular round_0 : signal (round_index))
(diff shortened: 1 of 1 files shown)


silveroak-opentitan/aes/Acorn/Cipher.v
```diff
@@ -49,12 +49,44 @@ Section WithCava.
...
+      | _ => fun x => x
+      end.
+
+    Fixpoint packed_signal A: unpacked_left_type A -> signal A :=
+      match A with
+      | Pair a b => fun '(x,y) =>
+        mkpair (packed_signal _ x) y
+      | _ => fun x => x
+      end.
+  End Packing.
```
jadephilipoom, line 95: Why are these needed? Since all the types are concrete here,  we should be able to just use Gallina tuples and mkpair/unpair. There are extra `unpair` and `mkpair` lines but imo it's easier to understand what's going on at a glance, and it will be simpler in proofs if we don't have to reason about more layers of type definitions. We could also potentially make a notation to get the best of both worlds.

I'd like these to be removed from `cipher_new` and the definitions below it before merging to avoid complicating the proofs, unless there's a specific reason to keep them.

silveroak-opentitan/aes/Acorn/Cipher.v
```diff
@@ -123,33 +162,106 @@ Section WithCava.
...
+        (* Accept input if we are not busy next cycle *)
+        accepted_input <- and2 (becoming_idle, in_valid_i) ;;
+        (* We are only truly idle if there was no incoming input *)
+        inv_in_valid_i <- inv in_valid_i ;;
+        idle <- and2 (becoming_idle, inv_in_valid_i) ;;
+
+        (* Are we generating decryption key?
+        * Decryption requires a full pass to generate decryption key
+        * Currently not implemented/supported *)
+        let generating_decryption_key := constant false in
```
jadephilipoom, line 236: We can support this pretty easily, I think; `cipher_new` runs the loop, which generates both keys and ciphertext, and then selects only the ciphertext to return. We could just return both, and select only the ciphertext in the lemma statements.

silveroak-opentitan/aes/Acorn/Cipher.v
```diff
@@ -123,33 +162,106 @@ Section WithCava.
...
-        mkpair (mkpair (mkpair (mkpair num_rounds_regular round_0)
-                               is_decrypt) initial_state) round_i in
+    let loop_input := packed_signal cipher_signals
+      (is_decrypt, num_rounds_regular, round_0, initial_state, round_i) in
     loop_out <- loopDelayS cipher_inner_loop loop_input ;;
     let '(_, final_data) := unpair loop_out in
     ret final_data.
 
+  Context (num_rounds_regular round_0 round_1 round_final: signal (round_index)).
+  Context (inc_round: signal round_index -> cava (signal round_index)).
```
jadephilipoom, line 193: Since these are only used by one definition, maybe just have them as arguments instead?

jadephilipoom (changes requested):

silveroak-opentitan/aes/Acorn/Cipher.v
blaxill: I find the nested `mkpair (mkpair (mkpair (mkpair` less readable and the multiline `unpairs` verbose, but I'll remove this change from `cipher_new`. For the `aes_cipher_core` I will keep the packing as the state representation is somewhat fluid and it makes it easier to modify. I'd also be fine with notation as it would be no different for me. I wasn't think about proof- I realise now that these packing definitions add unnecessary proof work.

silveroak-opentitan/aes/Acorn/Cipher.v
blaxill: Yeah, I want to get the forward mode working correctly in verilog first so held off on changing `cipher_new` but actually if this will impact proof work its worth making the change sooner rather than later.

silveroak-opentitan/aes/Acorn/Cipher.v
blaxill: Ill address this in a follow up along with netlist generation.

silveroak-opentitan/aes/Acorn/Cipher.v
jadephilipoom: I'll see if I can make a notation -- that would be the ideal middle ground between concise definitions and automatable proofs!

jadephilipoom (approved):

satnam6502 (approved):

silveroak-opentitan/aes/Acorn/Cipher.v
```diff
@@ -49,12 +49,51 @@ Section WithCava.
...
+    Bit (* op_i/is_decrypt : true for decryption, false for encryption *)
+      ** round_index (* num_rounds_regular : round index of final round *)
       ** round_index (* round_0 : round index of first round *)
-      ** Bit (* op_i/is_decrypt : true for decryption, false for encryption *)
       ** cipher_state (* initial state, ignored for all rounds except first *)
       ** round_index (* current round_index *).
 
+  Let cipher_control_state : SignalType :=
+    Bit            (* idle *)
+    ** Bit         (* generating decryption key *)
```
satnam6502, line 60: As I always keep saying, I'm not a fan of notation but perhaps it is tolerable here.

satnam6502 (approved):

PR #541, opened 2021-02-15
Fix handling of loop input signals

jadephilipoom:
As discussed on Friday -- AES loop with all the signals that need to be accessed inside the loop body packaged up and treated as loop input so that their timing is correctly handled.


blaxill (approved):
Lgtm

PR #542, opened 2021-02-15
Separate cipher control core

blaxill:
@jadephilipoom I'll wait for your upcoming PR and will resolve any conflicts in this PR


jadephilipoom (approved):
Thanks for offering to wait and resolve conflicts!

One nitpick: since our recent naming convention has `XCircuit` for implementations and `XEquivalence` for the  corresponding proofs that the implementation is equivalent to the spec, the name `CipherCircuit.v` implies that the new file contains the implementation that `CipherEquivalence.v` has proofs about, which isn't the case. Maybe `CipherControl.v` or `CipherControlCircuit.v` instead?

PR #543, opened 2021-02-16
Proofs for AES with sequential loop

### LINKED ISSUE #413: Make Acorn AES use a sequential loop

jadephilipoom:
Resolves #413 

The top-level proof (`full_cipher_equiv`) as we discussed on chat, now states that the output value *at timestep 14* in the output sequence holds the expected value. For intermediate states, see the very slightly lower-level `cipher_loop_equiv`, which uses the "interleaved cipher" spec to provide the exact full trace of the loop. The proof that decryption is the inverse of encryption runs the circuit twice and pulls out the right value using `nth` each time, which is not quite as elegant as the monad composition we were able to do before.

Currently, there are a few admitted lemmas about `muxPair`, `map`, and `combine`, which have shown up in our assumptions-tracking files. This is because when I tried to merge with main I found that the type signature of `cipher` had changed and now required a circuit to select the round constant as an argument. Adding the admitted lemmas was the quickest way to resolve the merge conflicts so we can unblock #542, but we should discuss whether to prove them or just put the selector circuit into the cipher control level instead.

Also:
- New lemmas about `fold_left_accumulate` invariants
- New lemmas to simplify reasoning about loops with purely combinational bodies
- Some improvements to tactics (`instantiate_app_by_reflexivity` and `factor_out_loops`)

### DIFF
--- a/silveroak-opentitan/aes/Acorn/Cipher.v
+++ b/silveroak-opentitan/aes/Acorn/Cipher.v
@@ -112,7 +112,7 @@ Section WithCava.
                            * signal (Pair (Pair key round_constant) state) =>
          let '(idx, feedback_state) := index_and_state in
          is_first_round <- idx ==? round_0 ;;
-         cipher_state <- muxPair is_first_round (initial_state, feedback_state) ;;
+         cipher_state <- muxPair is_first_round (feedback_state, initial_state) ;;
          cipher_step is_decrypt cipher_state idx)
       round_i.
 
--- a/silveroak-opentitan/aes/Acorn/Cipher.v
+++ b/silveroak-opentitan/aes/Acorn/Cipher.v
@@ -49,11 +49,50 @@ Section WithCava.
   Let cipher_state : SignalType := key ** round_constant ** state.
   (* The non-state signals that each round of the cipher loop needs access to *)
   Let cipher_signals : SignalType :=
-    round_index (* num_rounds_regular : round index of final round *)
+    Bit (* op_i/is_decrypt : true for decryption, false for encryption *)
+      ** round_index (* num_rounds_regular : round index of final round *)
       ** round_index (* round_0 : round index of first round *)
-      ** Bit (* op_i/is_decrypt : true for decryption, false for encryption *)
       ** cipher_state (* initial state, ignored for all rounds except first *)
       ** round_index (* current round_index *).
+
+  Let cipher_control_state : SignalType :=
+    Bit            (* idle *)
+    ** Bit         (* generating decryption key *)
+    ** round_index (* current round *)
+    ** Bit         (* in_ready_o *)
+    ** Bit         (* out_valid_o *)
+    ** state       (* state_o *)
+    ** Bit         (* out latch when not accepted *).
+
+  Let cipher_control_signals :=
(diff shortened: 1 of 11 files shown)


silveroak-opentitan/aes/Acorn/Cipher.v
```diff
@@ -94,110 +94,96 @@ Section WithCava.
...
-                                    (k, round, st) idx ;;
-    ret (mkpair (mkpair (fst key_round) (snd key_round)) st).
-
-  Context {seqsemantics : CavaSeq semantics}.
-  Context (initial_rcon: signal Bit -> cava (signal round_constant)).
-
-  Definition cipher_new
-             (num_rounds_regular round_0 : signal (round_index))
+  Definition cipher
+             (initial_rcon_selector : signal Bit -> cava (signal round_constant))
```
blaxill, line 163: Feel free to move `initial_rcon_selector` as you see fit. The intention is for it to be absorbed into key_expand and there to be no mention of rcon outside of key expansion.

blaxill (approved):
>Currently, there are a few admitted lemmas about muxPair, map, and combine, which have shown up in our assumptions-tracking files. This is because when I tried to merge with main I found that the type signature of cipher had changed and now required a circuit to select the round constant as an argument. 

Are the `map` and `combine` proofs because of lifting rcon too? Its not clear to me why they are additionally needed. Will noving `rcon` into `key_expand` (like the opentitan version) will remove all these proof requirements?

silveroak-opentitan/aes/Acorn/CipherEquivalence.v
```diff
@@ -81,150 +82,384 @@ Section WithSubroutines.
...
+    (* simplify round-constant selection *)
+    rewrite muxPair_correct_strong.
+    rewrite (pad_combine_eq init_rcon_fwd) by length_hammer.
+    rewrite (@pad_combine_eq
+               Bit (Pair (Vec Bit 8) (Vec Bit 8))
+               _ (combine init_rcon_fwd _))
+      by (cbn [combType] in *; length_hammer).
+    rewrite combine_repeat_l, map_map
+      by (cbn [combType] in *; length_hammer).
+    cbn [combType] in *. rewrite map_combine_drop_r by length_hammer.
```
jadephilipoom, line 443: @blaxill Here's where the `map_combine` lemmas get used; this block is for simplifying the round_constant selection and was added to the proof for both cipher directions. Basically, because `mkpair` is implemented by padding the lists to be the same length if necessary, there's a lot of length/list-combining reasoning required to simplify the expression (in this case, to use the constant selector to write away the mux).

jadephilipoom:
> Will noving `rcon` into `key_expand` (like the opentitan version) will remove all these proof requirements?

Hard to say at this stage, but I think yes. I think it could work like the current `initial_state` and `initial_key` for `cipher`; it could take in some constant inputs for the forward/inverse round constants, and then based on the round index select either that value or the round constant from the feedback wire.

PR #544, opened 2021-02-16
Factor out mul2 and mul4

jadephilipoom:
Follow-up to #535 

Just enough work on the `mix_single_column` equivalence lemma to factor out the right lemmas for `aes_mul2` and `aes_mul4`!

### DIFF
--- a/silveroak-opentitan/aes/Acorn/MixColumnsEquivalence.v
+++ b/silveroak-opentitan/aes/Acorn/MixColumnsEquivalence.v
@@ -27,7 +27,8 @@ Require Import Cava.Acorn.MonadFacts.
 Require Import Cava.Acorn.Identity.
 Require Import Cava.Acorn.Acorn.
 Require Import Cava.Lib.BitVectorOps.
-Import ListNotations.
+Import ListNotations VectorNotations.
+Local Open Scope list_scope.
 
 Require Import AesSpec.AES256.
 Require Import AesSpec.StateTypeConversions.
@@ -36,6 +37,7 @@ Require Import AcornAes.MixColumnsCircuit.
 Import StateTypeConversions.LittleEndian.
 
 Existing Instance CombinationalSemantics.
+Existing Instance MixColumns.byteops.
 
 Section Equivalence.
   Local Notation byte := (Vector.t bool 8).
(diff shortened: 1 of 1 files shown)


silveroak-opentitan/aes/Acorn/MixColumnsEquivalence.v
```diff
@@ -67,13 +69,89 @@ Section Equivalence.
...
     = [if is_decrypt
        then map byte_to_bitvec
                 (MixColumns.inv_mix_single_column (map bitvec_to_byte col))
        else map byte_to_bitvec
                 (MixColumns.mix_single_column (map bitvec_to_byte col))].
+  Proof.
+    (* simplify LHS *)
+    cbv [aes_mix_single_column]. simpl_ident.
+    constant_vector_simpl col.
+    repeat lazymatch goal with
```
cipher1024, line 131: Would it be more concise instead of this `repeat lazymatch ... rewrite` to use `autorewrite` and build a list with all those lemmas? You would still need the `change` statements but the list of lemmas might be reusable.

jadephilipoom: It would be more concise, but much slower. `lazymatch` is fairly fast because it only tries the first matching arm, while `autorewrite` will try to match every hint. It's already a big expression and a little slow, so here I opted for the faster strategy.

It's also helpful to have a `match` here because some of the lemmas, like `muxPair_correct` don't successfully rewrite unless you explicitly pass a type argument (which you can dynamically pull out of the goal expression in a match) -- the lemmas are written about `combType ?t` and Coq doesn't unify `combType ?t` with `Vector.t bool 8` unless you explicitly plug in `t:=Vec Bit 8`.

cipher1024: Thanks for checking!

cipher1024 (approved):

silveroak-opentitan/aes/Acorn/MixColumnsEquivalence.v
jadephilipoom: For archival purposes/Coq performance tidbit: Simon and I talked about this further over chat and ended up doing a performance comparison. In case anyone else is interested in the numbers: replacing the `xorV_is_add`, `xorv_is_add`, `aes_mul2_correct`, `aes_mul4_correct`, `byte_to_bitvec_to_byte`, and `bitvec_to_byte_to_bitvec` clauses with an `autorewrite` using a new hint database with those lemmas takes 7.36s as opposed to 4.49s for `lazymatch`.

PR #546, opened 2021-02-17
Make loop combinator use a Gallina tuple for feedback

satnam6502:
As part of the process of eliminating the tuple type from `SignalType` and doing away with `mkpair` and `unpair` this PR proposes an alternative formulation for a loop combinator that uses a Gallina flat-tuple to structure feedback information. The PR needs proofs to be done for the new combinator, which I am leaving to @jadephilipoom to add as further commits to this PR. Please delete the old combinators e.g. `loopDelaySR` and then rename `loopDelaySRAlt` to `loopDelaySR` etc. Thank you kindly.

The PR demonstrates how to use the existing `tupleInterface` infrastructure we have to define a tuple feedback tuple for `loopDelaySR`:

```coq
  loopDelaySRAlt : forall {A : SignalType} {B : list SignalType},
                   tupleInterface combType B ->
                   (signal A * tupleInterface signal B -> cava (tupleInterface signal B)) ->
                   signal A ->
                   cava (tupleInterface signal B); 
```

For example, bit `B = [Bit]` then a specific loop type would be:
```coq
forall {A : SignalType} 
                   Bool ->
                   (signal A * signal Bit -> cava (signal Bit)) ->
                   signal A ->
                   cava (signal Bit); 
```
If `B = [Bit; Vec Bit 8]` then we would have:
```coq
forall {A : SignalType} 
                   (Bool, Vector.t Bool 8) ->
                   (signal A * (signal Bit, signal (Vec Bit 8)) -> cava (signal Bit * signal (Vec Bit 8))) ->
                   signal A ->
                   cava (signal Bit * signal (Vec Bit 8)); 
```

This might be a good time to make `loopDelaySEnableR` the sole primitive loop combinator in `CavaSeq` and then define all the others as specialized versions in a library module (or `Combinators.v`). Another thing that could be considered at this stage is to merge back `CavaSeq` back into `CavaClass` (or this can be left for a later PR).

Note how we have separate signals for each component of the composite state, and the netlist layer generates separate registers for each of the elements of the composite state. So now we can have multiple streams of values flowing into the core of the loop body which can be taken apart and merged with combinators. The current formulation has been restricting us to a single feedback signal (which can contain values built using our `SignalType` tuples), which I think is not what we really want.

Once the loop combinators are ported over we should be able to delete the `SignalType` tuples, `mkpair`, `unpair` etc. assuming these definitions do not add more complexity for proofs than they remove.

A concrete test example added in the PR, along with a Coq-level `Example` test and netlist generation for a SystemVerilog test:
```coq
  Definition countByInvAlt : signal (Vec Bit 8) -> cava (signal (Vec Bit 8) * signal Bit)
    := loopDelaySRAlt (tupleInterfaceDefault [Vec Bit 8; Bit])
       (pairLeft >=> first addN >=> second inv).
```
The Coq test:
```coq
(* Check the toggling output version. *)
Example countByInvAlt_ex1: sequential (countByInvAlt [b14; b7; b3; b250])
  = ([b14; b21; b24; b18], [true; false; true; false]).
Proof. reflexivity. Qed.
```
The SystemVerilog test:
```
clang++ -L/usr/local/opt/sqlite/lib    countByInvAlt_tb.o verilated.o verilated_vcd_c.o VcountByInvAlt_tb__ALL.a    -o VcountByInvAlt_tb -lm -lstdc++ 
obj_dir/VcountByInvAlt_tb
                  10: tick = 0,  i = 14,  o = 0,  t = 0
                  20: tick = 1,  i = 7,  o = 14,  t = 1
                  30: tick = 2,  i = 3,  o = 21,  t = 0
                  40: tick = 3,  i = 250,  o = 24,  t = 1
```
**QUESTION** : Why does the Coq `Example` test omit the first output i.e. `(0, false)`?
The generated SystemVerilog does produce this first output value, hence the different expected output given for the SysemVerilog test bench for this case vs. the `Example` expected output.

CC: @blaxill 

Addresses part of #416

### DIFF
--- a/cava/Cava/Acorn/CavaClass.v
+++ b/cava/Cava/Acorn/CavaClass.v
@@ -104,6 +104,13 @@ Class CavaSeq {signal : SignalType -> Type} (combinationalSemantics : Cava signa
                 (signal A * signal B -> cava (signal B)) ->
                 signal A ->
                 cava (signal B);
+  (* A version of loopDelaySRAlt that has a composite state value of a tuple-type computed
+     from the list B, or a scalar type when B is a singleton list. *)
+  loopDelaySRAlt : forall {A : SignalType} {B : list SignalType},
+                   tupleInterface combType B ->
+                   (signal A * tupleInterface signal B -> cava (tupleInterface signal B)) ->
+                   signal A ->
+                   cava (tupleInterface signal B); 
   (* A version of loopDelayEnable with a clock enable and current state at
      the output. *)
   loopDelaySEnableR : forall {A B: SignalType},
--- a/cava/Cava/Acorn/NetlistGeneration.v
+++ b/cava/Cava/Acorn/NetlistGeneration.v
@@ -252,6 +252,61 @@ Definition loopNetS (A B : SignalType)
   assignSignal o oDelay ;;
   ret out.
 
+Fixpoint zipTupleMR {m} `{Monad m}
+                    (signal : SignalType -> Type)
+                    (ts : list SignalType)
+                    (f : forall (A : SignalType), signal A -> combType A -> m (signal A)) :
+                    tupleInterfaceR signal ts ->
+                    tupleInterfaceR combType ts ->
+                    m (tupleInterfaceR signal ts) :=
+  match ts  with
+  | [] => fun _ _ => ret tt
+  | x::xs => fun a b => v <- f x (fst a) (fst b) ;;
+                        vx <- zipTupleMR signal xs f (snd a) (snd b) ;;
+                        ret (v, vx)
+  end.
+
+Definition zipTupleM {m} `{Monad m}
+                     (signal : SignalType -> Type)
+                     (ts : list SignalType)
+                     (f : forall (A : SignalType), signal A -> combType A -> m (signal A)) 
+                     (a : tupleInterface signal ts)
+                     (b : tupleInterface combType ts)
+                     : m (tupleInterface signal ts) :=
+  v <- zipTupleMR signal ts f (unbalance signal ts a) (unbalance combType ts b) ;;
+  ret (rebalance signal ts v).
+
+Fixpoint zipTupleMR_ (signal : SignalType -> Type)
+                     (ts : list SignalType)
+                     (f : forall (A : SignalType), signal A -> signal A -> state CavaState unit) :
+                     tupleInterfaceR signal ts ->
+                     tupleInterfaceR signal ts ->
+                     state CavaState unit :=
+  match ts  with
+  | [] => fun _ _ => ret tt
+  | x::xs => fun a b => f x (fst a) (fst b) ;;
+                        zipTupleMR_ signal xs f (snd a) (snd b) ;;
+                        ret tt
+  end.
+
+Definition zipTupleM_ (signal : SignalType -> Type)
+                      (ts : list SignalType)
+                      (f : forall (A : SignalType), signal A -> signal A -> state CavaState unit)
+                      (a b : tupleInterface signal ts)
+                      : state CavaState unit :=
+  zipTupleMR_ signal ts f (unbalance signal ts a) (unbalance signal ts b).
+
+Definition loopNetSAlt (A : SignalType) (ts : list SignalType)
+                       (resetValue : tupleInterface combType ts)
+                       (f : Signal A * tupleInterface Signal ts -> state CavaState (tupleInterface Signal ts))
+                       (a : Signal A)
+                       : state CavaState (tupleInterface Signal ts) :=
+  o <- mapTuple_ Signal ts newSignal ;;
+  out <- f (a, o) ;;
+  oDelay <- zipTupleM Signal ts (fun A x r => delayNet A r x) out resetValue ;;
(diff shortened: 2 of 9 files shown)


cava/Cava/Acorn/NetlistGeneration.v
```diff
@@ -252,6 +252,61 @@ Definition loopNetS (A B : SignalType)
   assignSignal o oDelay ;;
   ret out.
 
+Fixpoint zipTupleMR {m} `{Monad m}
```
jadephilipoom, line 255: nit: maybe `map2` instead of `zip`, since a function is being applied here?

jadephilipoom (commented):
Why do you expect a `(0, false)` output from this function? Is that from the delay? The loops' output is measured before the delay, so if the SV testbench checks the values after the delay it makes sense that they would disagree.

cava/Cava/Acorn/CavaClass.v
```diff
@@ -104,6 +104,13 @@ Class CavaSeq {signal : SignalType -> Type} (combinationalSemantics : Cava signa
                 (signal A * signal B -> cava (signal B)) ->
                 signal A ->
                 cava (signal B);
+  (* A version of loopDelaySRAlt that has a composite state value of a tuple-type computed
+     from the list B, or a scalar type when B is a singleton list. *)
+  loopDelaySRAlt : forall {A : SignalType} {B : list SignalType},
+                   tupleInterface combType B ->
+                   (signal A * tupleInterface signal B -> cava (tupleInterface signal B)) ->
+                   signal A ->
+                   cava (tupleInterface signal B);
```
jadephilipoom, line 113: Unfortunately we often need many input values, so we'll need a list for A as well.

blaxill:
> **QUESTION** : Why does the Coq `Example` test omit the first output i.e. `(0, false)`?
> The generated SystemVerilog does produce this first output value, hence the different expected output given for the SysemVerilog test bench for this case vs. the `Example` expected output.

FWIW this seems to correspond to registering state but not registering output which is what I'd expect:

~~~
I -> S -> updated S (delayed)
I -> S -> O (not delayed)
~~~

This corresponds to what i'd expect for handshaking- when both valid & ready on the same clock cycle we transition to accept token state at next time step. With a delayed `O` its still possible, but our output valid/ready would be registered which would be extra resources and the state transition logic would have to account for it.

jadephilipoom:
After some thought, I have three primary concerns here:

1. This requires a lot of layers of definitions. In particular, because every function on tuples is essentially defined twice (once for `tupleInterfaceR` and once for `tupleInterface`, every lemma about `tupleInterface` or any function defined over it will also require a corresponding proof about `tupleInterfaceR`.
2. To allow the input to be a tuple, we need to define `toListOfTuples`, the inverse of `fromListOfTuples` (so that `fold_left` can work over the list). However, this exposes us to the same problem that we have currently with `unpeel` -- that is, if the list of types is empty, `toListOfTuples` has to convert an *empty* tuple of lists into a list of empty tuples. But, because there are no lists to look at, it can't know how long to make the list of empty tuples. This would force us to put a `ts <> nil` precondition on all loop lemmas, which would complicate the already tricky inductive logic in those proofs.
3.  The `tupleInterface` setup has the exact same problem for which we rejected https://github.com/project-oak/silveroak/pull/445; it will require explicit type annotations in circuit descriptions for loops. The reasons `countByInvAlt` doesn't currently require annotations is because a) the loop input isn't a tuple, so no annotation is needed for that and b) it's a loop with a reset value, so it's pulling the state type from the reset value which is also defined in terms of tuples. If the explicit type there is removed `(tupleInterfaceDefault [Vec Bit 8; Bit])` => `(tupleInterfaceDefault _)`, then the definition fails to typecheck without `(B:=[Vec Bit 8; Bit])`. Since `loopDelayS` doesn't have the default type, any definition with `loopDelayS` will require the type annotation.

I'll try to do some fail-fast experiments with different ways of handling the tuple problem to see if we can resolve these issues somehow, and report back soon.

satnam6502:
OK! All great points. Let's suspend this PR for now and I shall propose an alternative PR.

cava/Cava/Acorn/NetlistGeneration.v
satnam6502: Yes, I was worrying about the name, which is really a `zipWith`.

jadephilipoom:
I've put my experiments up at https://gist.github.com/jadephilipoom/7d2ccba2b8a7849c6f93efaf22b94696

For each experiment, I did the minimal setup to define the `countByInvAlt` example as a comparison point. I tried the existing setup (pairs in `SignalType`), the `tupleInterface` setup, the setup from #445 (SignalInterface), and two new ideas: a) making lists work over completely abstract types (didn't work) and b) creating an inductive structure to represent the sequential structure of a circuit (did work).

I actually think the last one (in the experiments under `SequentialInductive`) is promising. It doesn't require type annotations for the circuit, and actually has the property that the loop "disappears" when the circuit is interpreted. It would allow us to keep the shallow embedding for combinational subcomponents, but leverage the power of a known inductive structure when reasoning about sequential ones.

satnam6502:
Abandoning in preference to a better idea from @jadephilipoom for how to do loop.

PR #547, opened 2021-02-17
Generate aes_cipher_core netlist

blaxill:
@satnam6502 trying to generate the netlist for `aes_cipher_core_simplified_Netlist` causes CI to run out of memory (and be killed), same thing happens locally for me. Any ideas whats wrong here? 

This currently compiles all our aes modules into one module. A backup plan would be to create all the netlist separately and then instantiate them explicitly (so separate systemverilog files for all).


blaxill:
Superseded by #573

PR #548, opened 2021-02-17
_CoqProject generation w/ find.

### LINKED ISSUE #472: Reconsider git ls-files | grep "\.v$" >> _CoqProject
I am not sure if we want to have the Coq source files discovered in this way:
```
git ls-files | grep "\.v$" >> _CoqProject
```
I often trip up when added a new file which has not had `git add` and commit applied to it.
When we have some stability perhaps it is not too bad just to explicitly list the source files in `_CoqProject`.

I also deleted a file `Types.v` but this still gets filtered to `_CoqProject`.
```
$ git status
On branch notypes
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   Cava/Cava.v
	modified:   Cava/Extraction.v
	deleted:    Cava/Types.v
	modified:   Cava2HDL.cabal
$ git ls-files | grep "\.v$"
...
Cava/Types.v
...
$ make
git ls-files | grep "\.v$" >> _CoqProject
coq_makefile -f _CoqProject -o Makefile.coq
/Library/Developer/CommandLineTools/usr/bin/make -f Makefile.coq
COQC Cava/Acorn/Acorn.v
File "./Cava/Acorn/Acorn.v", line 17, characters 15-35:
Error: Unable to locate library Cava.Acorn.CavaClass.

make[2]: *** [Cava/Acorn/Acorn.vo] Error 1
make[1]: *** [all] Error 2
make: *** [coq] Error 2
```

If I add the changes then things work fine.

CC: @jadephilipoom

smore-lore:
Closes #472 

This change:
- declares '.v' files as prerequisites for generating _CoqProject
  (avoids running rule, and thereby touching _CoqProject, if no .v
  files are changed, for faster incremental builds).
- uses 'find' over 'git ls-files'. Making code git-agnostic,
  specifically when creating a new file, don't need to `git add` in
  order to build it.
- adds assumption that file names don't have spaces

Future considerations:
- avoid 'xilinx' subdir excludes, for consistent style
- consider factoring out file lookup logic into a define or include
  which can use args or variables to specify source globs + dependency
  graph (this would also make it easier to order sources - which we
  now rely on being ordered alphabetically)

### DIFF
--- a/acorn-examples/Makefile
+++ b/acorn-examples/Makefile
@@ -35,13 +35,13 @@ xilinx:		adder8.vcd
 VERILATOR = verilator +1800-2012ext+sv verilator.vlt
 VLINT = $(VERILATOR) --lint-only
 
-_CoqProject:
+_CoqProject:    $(shell find . -name '*.v' -not -path "*/xilinx/*" | sort)
 		@echo "INSTALLDEFAULTROOT = AcornExamples" > _CoqProject
 		@echo "-R . AcornExamples" >> _CoqProject
 		@echo "-R ../cava/Cava Cava" >> _CoqProject
 		@echo "-R ../third_party/coq-ext-lib/theories ExtLib" >> _CoqProject
 		@echo "-R ../third_party/bedrock2/deps/coqutil/src/coqutil coqutil" >> _CoqProject
-		git ls-files | grep "\.v$$" | grep -v "^xilinx/" >> _CoqProject
+		@printf '%s\n' $^ >> _CoqProject
 
 Makefile.coq:	_CoqProject
 		coq_makefile -f _CoqProject -o Makefile.coq
--- a/acorn-examples/xilinx/Makefile
+++ b/acorn-examples/xilinx/Makefile
@@ -43,13 +43,13 @@ extraction:	$(SV)
 VERILATOR = verilator +1800-2012ext+sv verilator.vlt -y $(XILINX)/data/verilog/src/unisims
 VLINT = $(VERILATOR) --lint-only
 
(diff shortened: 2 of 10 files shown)


satnam6502:
One of the reasons for treating the `xilinx` directories differently is that they have a dependency on the Xilinx tools for synthesis and implementation, and even for Verilator simulation (because Xilinx simulation libraries are needed for the primitive elements). Perhaps that is not a good enough reason to treat them differently.
I'll defer to the others for review, especially @jadephilipoom since this proposes to replace her approach.

smore-lore:
Yeah, the directory structure is perfectly reasonable (and I'm not changing it - carrying forward the exceptions here). What I mean in 'future considerations' is that we could construct an abstraction like this:

```
LOCAL_MODULE_NAME := Cava
...
include $(BUILD_COQ_LIBRARY)
...
LOCAL_MODULE_NAME := foobarbazbiz
LOCAL_COQ_DEPS := Cava # to lookup include path for '-R' option
LOCAL_SRC_FILES := *.v foobar/*.v # explicit source declaration
include $(BUILD_COQ_LIBRARY)
```

That is - we shouldn't assume all files in subdirectories are part of a given unit/module. However, I'm really not sure if there is better make-fu for this (I'm heavily influenced by Android's make style, but it _really_ stretches the limits of make and it really doesn't scale). I also hesitate to suggest a different build system, because velocity :)

jadephilipoom (approved):
LGTM, thanks!

Issue #552, opened 2021-02-18
muxPair/pairSel potentially creates Pairs in Vectors


blaxill:
Pairs in Vectors become a problem for our SystemVerilog generation and shouldn't be emitted in our netlists. Unfortunately since muxPair/pairSel now put their arguments into a vector to be indexed, nested Pairs will create Pairs in Vectors when passed to muxPair.

satnam6502:
And just to be clear, we _could_ define SystemVerilog record types to represent each instance of pairs, but I'd rather we did not and avoided all issues related to nominal vs. structural typing. Also, although we emit SystemVerilog, I'd like to keep open the door to also emitting Verilog, especially for processing by Lutsig, which means no records.

jadephilipoom:
Resolved by #595, where pairs were removed entirely.

PR #553, opened 2021-02-18
Avoid potentially creating Pairs inside Vecs in muxPair/pairSel

blaxill:
Unfortunately this solution would break some proofs. I've admitted broken proofs. The proof automation no longer works for the examples, I presume as I changed the type of pairSel to `cava _`  #552

cc @satnam6502  @jadephilipoom

### DIFF
--- a/cava/Cava/Acorn/CavaPrelude.v
+++ b/cava/Cava/Acorn/CavaPrelude.v
@@ -24,6 +24,8 @@ Import MonadNotation.
 Require Import Cava.Acorn.CavaClass.
 Require Import Cava.Signal.
 
+Local Open Scope monad_scope.
+
 Section WithCava.
   Context {signal} `{Cava signal} `{CavaSeq signal}.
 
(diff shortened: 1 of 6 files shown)


satnam6502:
A pair is still being created somewhere.
```
Generating cipher.sv...AESSV: (v35 + 8'h1)
CallStack (from HasCallStack):
  error, called at Cava2SystemVerilog.hs:178:29 in Cava2HDL-0.1.0.0-A0nwmqFM9TwLGofj2vO0AK:Cava2SystemVerilog
CallStack (from -prof):
  Cava2SystemVerilog.showSignal (Cava2SystemVerilog.hs:(159,1)-(181,88))
  Cava2SystemVerilog.generateInstance (Cava2SystemVerilog.hs:(208,1)-(258,65))
  Cava2SystemVerilog.cava2SystemVerilog (Cava2SystemVerilog.hs:(70,1)-(100,46))
  Cava2SystemVerilog.writeSystemVerilog (Cava2SystemVerilog.hs:(46,1)-(54,73))
  Main.main (AESSV.hs:(26,1)-(37,57))
make: *** [AESSV] Error 1
```

I changed `showSignal` to now generate an error when we find a pair.
```haskell
      SignalPair _ _ a b -> error ("SignalPair found: (" ++ showSignal a ++ ", " ++ showSignal b ++ ")")
```

satnam6502:
It's late so I am going to turn in, but I think we need to find out where a pair-type from the feedback path on the loop combinator is not being immediately converted to a Gallina tuple. I guess some part of the feedback structure is remaining in a pair-type which then ends up as a value in a vector (which is not legal for our SystemVerilog generation).

blaxill:
> I changed `showSignal` to now generate an error when we find a pair.

I think our debugging may have diverged a bit, can you push a branch with your changes?

Won't that `showSignal` change error on all `Pair`s, even those not inside `Vec`s?

blaxill:
Ahhh I think https://github.com/project-oak/silveroak/blob/0adc80b5194a54afff21b7648dea4d3993dc2ea2/cava/Cava2SystemVerilog.hs#L175-L176 should be `showSignal (simplifySignal signal)`

satnam6502:
> Ahhh I think
> 
> https://github.com/project-oak/silveroak/blob/0adc80b5194a54afff21b7648dea4d3993dc2ea2/cava/Cava2SystemVerilog.hs#L175-L176
> 
> should be `showSignal (simplifySignal signal)`
> #555

Ah! Thanks for the fix: applied locally.

cava/Cava/Acorn/CavaPrelude.v
```diff
@@ -42,18 +44,25 @@ Section WithCava.
...
+    match A with
+    | Pair _ _ => fun '(a,b) =>
+      let '(x1, y1) := unpair a in
+      let '(x2, y2) := unpair b in
+      x <- muxPair sel (x1, x2) ;;
+      y <- muxPair sel (y1, y2) ;;
+      ret (mkpair x y)
+    | _ => fun '(a, b) =>
+      ret (indexAt (unpeel [a; b]) (unpeel [sel]))
+    end.
```
satnam6502, line 59: I've proposed a different change to `muxPair` to avoid expressions blowing up in size: #562
Let's get to the bottom of why there are pairs getting through the the netlist layer: this should not be happening except for the single case of Fst and Snf projecting from a pair literal.

blaxill:
Pairs will no longer be required with the approach in #561

PR #556, opened 2021-02-18
Move rcon_selector into cipher control

### LINKED ISSUE #545: Fix map/combine and muxPair admits
See discussion in #543

We should move `initial_rcon_selector` into the cipher core logic for now to stop depending on the admitted lemmas, but also explore making rcon completely internal to key expansion; it shouldn't really be visible to `cipher`.

jadephilipoom:
Resolves #545


blaxill (approved):
Lgtm! For #545 it would be good if we could go all the way to pushing it into key_expansion

jadephilipoom:
> Lgtm! For #545 it would be good if we could go all the way to pushing it into key_expansion

I made a new issue for this: #560

PR #561, opened 2021-02-18
Inductive structure for sequential circuits

jadephilipoom:
This PR adds some parallel files with the new infrastructure I'd like to use for sequential circuits. At a glance:
- The Cava instance is only for combinational circuits and no longer uses lists
- An inductive structure `Circuit` represents loops, delays, and chaining of these with each other and with combinational parts (which are still just monadic Gallina functions)
- Added a test `double_count_by` for a loop-within-a-loop : a counter with a smaller counter inside it that increments when the smaller counter overflows
- Duplicate of `Cipher.v` that defines the AES loop in terms of the new loop structure and removes `Pair`s

This is sort of like a "Goldilocks" solution between the arrows approach, with inductive structures all the way down, and the current approach. I think it can help us capture enough structural information to keep track of internal state while keeping enough of the structure (i.e. the combinational stuff) in native Gallina to avoid performance problems.


satnam6502 (approved):
This looks like a promising approach!

cava/Cava/Acorn/Circuit.v
```diff
@@ -0,0 +1,89 @@
...
+  | Comb : forall {i o}, (i -> cava o) -> Circuit i o
+  | Compose : forall {i t o}, Circuit i t -> Circuit t o -> Circuit i o
+  | First : forall {i o t}, Circuit i o -> Circuit (i * t) (o * t)
+  | Second : forall {i o t}, Circuit i o -> Circuit (t * i) (t * o)
+  | Loop :
+      forall {i s o : Type},
+        Circuit (i * s) (s * o) ->
+        Circuit i o
+  | Delay : forall {t}, Circuit t t
+  .
```
blaxill, line 45: I could be wrong but I think the definition as-is won't work for NetlistGeneration as omitting the `(signal : SignalType -> Type)` doesn't allow instance specific representations. i.e. you will be able to build `Circuit` with an arbitrary `Cava`, but instances like NetlistGeneration need to be able to inspect their inputs/outputs generically. 

Specifically, the netlist generation equivalent of `interp` needs to be able to create new `Signal`s of its port types by calling `newSignal` with an argument of `SignalType`. We can restrict the `i` `o` types to `Signal i` `Signal o`, but because Loop has `s` as an existential type not exposed in its constructor type, I don't think we can write a recursive function to wire up the netlist:

```
Fixpoint netlist_generation {i o} (c : Circuit (Signal i) (Signal o)) :=
match c with
| @Loop i s o f => 
   o <- @newSignal _ ;; 
                   ^ this hole is SignalType and should correspond to 's'
                      but we don't have anything in scope, and its not "match"able as 
                      its of type Type
```

The reason for `newSignal` is its needed to wire up the circuit, e.g. the existing `loopNetS`: 

```
Definition loopNetS (A B : SignalType)
                    (resetValue : combType B)
                    (f : Signal A * Signal B -> state CavaState (Signal B))
                    (a : Signal A)
                    : state CavaState (Signal B) :=
  o <- @newSignal B ;;
  out <- f (a, o) ;;
  oDelay <- delayNet B resetValue out ;;
  assignSignal o oDelay ;;
  ret out.
```

I'm not sure how to resolve this without losing the ability to use Gallina tuples.

blaxill (commented):
Im pro single step semantics, so I'm happy to see it here.

I think the current Circuit definition might be problematic for other instances such as NetlistGeneration who expect their circuit types to be inspectable, see inline comment. I would be pleasantly surprised if I'm wrong though.

If the inductive structure is preferable for proof, it could be used with the existing monad structure by making the sequential instance construct a `Circuit` and running `interp` on it like you have shown here, but without requiring e.g. the netlist instance to use `Circuit`.

cava/Cava/Acorn/Circuit.v
jadephilipoom: I see what you mean. Let me think on this one and see if I can write the netlist generation function that wires up the circuit.  I think we'll have to have some more information passed in by the caller to netlist generation to maintain the type generality, since in this setup it *is* possible to construct a loop with invalid types. I have a couple initial ideas:

- Require the caller of netlist generation to pass in the correct new signals, which we can determine based on `circuit_state`.
- The existing `interp`works for all monads, and creates a monadic expression for the circuit's step function. Netlist generation could therefore use `interp` and then just wire up corresponding parts of the state (i.e. connect the output state to the input state with delays). We could write a tactic that matches on the type of `s` and creates the right expression for this (it would also have to be passed in by the caller).
- Change `Circuit` to construct loops by having two constructors; one creates an identity-function loop by adding signals to the state one at a time (i.e. constructs a delay with the right state type one signal at a time), and the other hooks it up to a loop body.

jadephilipoom:
It turns out that I can just change the `s` parameter of `Loop` to be `signal` without losing anything -- multiple signals are represented as multiple loops. I can still express the AES loop just fine, using three `Loop`s to correspond to the three signals. This makes sense to me -- if you think of the circuit diagram for a three-signal loop, it would be three delays with wires looped around to the same internal circuit, which is exactly what the three loops represent.

silveroak-opentitan/aes/Acorn/CipherNewLoop.v
```diff
@@ -0,0 +1,132 @@
...
+                        round_0, initial_state, idx) := input in
+                  let '(ik, ir, iv) := initial_state in
+                  is_first_round <- idx ==? round_0 ;;
+                  k <- muxPair is_first_round (fk, ik) ;;
+                  r <- muxPair is_first_round (fr, ir) ;;
+                  v <- muxPair is_first_round (fv, iv) ;;
+                  '(k',r',v') <- cipher_step is_decrypt is_first_round
+                                            num_rounds_regular (k,r,v) idx ;;
+                  let out := (k',r',v') in
+                  ret (out,k',r',v'))))).
```
jadephilipoom, line 126: New `cipher_loop` with state separated into three `Loop`s -- outermost has `(s:=key)`, middle has `(s:=round_constant)`, and innermost has `(s:=state)`. The associativity of the tuples ends up grouping slightly differently (e.g. associating the key with input/output instead of other state information), but imo that's a reasonable price to pay here.

cava/Cava/Acorn/NetlistGeneration.v
```diff
@@ -322,3 +322,132 @@ Instance CavaSequentialNet : CavaSeq CavaCombinationalNet :=
...
+    : state CavaState o :=
+  in_state <- newCircuitStateSignals c ;; (* x : circuit_state c *)
+  '(out, out_state) <- interp c in_state input ;;
+  linkCircuitStateSignals c in_state out_state ;;
+  ret out.
+
+Definition makeCircuitNetlist (intf : CircuitInterface)
+           (c : Circuit (tupleNetInterface (circuitInputs intf))
+                        (tupleNetInterface (circuitOutputs intf))) : CavaState :=
+  makeNetlist intf (interpCircuit c).
```
jadephilipoom, line 377: I think now I've managed to define netlist generation! Below here are a couple of tests with delays and loops to see if it works and it looks fine to me -- @satnam6502 @blaxill could you take a look? I'll delete the tests before merging.

cava/Cava/Acorn/NetlistGeneration.v
```diff
@@ -322,3 +322,132 @@ Instance CavaSequentialNet : CavaSeq CavaCombinationalNet :=
...
+  (*             moduleName := "test"; *)
+  (*             netlist := [AssignSignal (NamedWire "o") (Wire 0); *)
+  (*                        Netlist.Delay Bit Gnd (Wire 0) (NamedWire "i1")]; *)
+  (*             inputs := [{| port_name := "i1"; port_type := Bit |}; *)
+  (*                       {| port_name := "i0"; port_type := Vec Bit 2 |}; *)
+  (*                       {| port_name := "rst"; port_type := Bit |}; *)
+  (*                       {| port_name := "clk"; port_type := Bit |}]; *)
+  (*             outputs := [{| port_name := "o"; port_type := Bit |}] |}; *)
+  (*   libraryModules := [] |} *)
+  (* : CavaState *)
```
jadephilipoom, line 410: Test with delay

cava/Cava/Acorn/NetlistGeneration.v
```diff
@@ -322,3 +322,132 @@ Instance CavaSequentialNet : CavaSeq CavaCombinationalNet :=
...
+  (*   module := {| *)
+  (*             moduleName := "loop_test"; *)
+  (*             netlist := [AssignSignal (NamedWire "o") (Wire 1); Netlist.Delay Bit Gnd (Wire 0) (Wire 1); *)
+  (*                        Xor (NamedWire "i") (Wire 0) (Wire 1)]; *)
+  (*             inputs := [{| port_name := "i"; port_type := Bit |}; *)
+  (*                       {| port_name := "rst"; port_type := Bit |}; *)
+  (*                       {| port_name := "clk"; port_type := Bit |}]; *)
+  (*             outputs := [{| port_name := "o"; port_type := Bit |}] |}; *)
+  (*   libraryModules := [] |} *)
+  (* : CavaState *)
```
jadephilipoom, line 452: Test with loop

silveroak-opentitan/aes/Acorn/CipherNewLoop.v
blaxill: Interesting. I think this solution might also work for the existing `loopDelayS` by doing the same trick - having input not specialized to `signal _`?

blaxill: This seems to work quite nicely, I like this idea.

blaxill (approved):
Lgtm!

silveroak-opentitan/aes/Acorn/CipherNewLoop.v
jadephilipoom: I agree, it could probably work for `loopDelayS` also, although it wouldn't solve the fundamental problem with the current loop formulation where it produces incorrect behavior for loops within loops.

PR #564, opened 2021-02-19
Parallel Identity.v and CombinationalProperties.v

jadephilipoom:
Recreate new versions of `Identity.v` and `CombinationalProperties.v` that use the new purely-combinational `Cava` instance, similar to what we had before switching over to using `seqType` for everything. I also moved the `zipWith` and `xorV` lemmas to `CombinationalProperties` instead of `Identity` because I think they are better categorized there. The new files are much smaller than the old ones because of the simpler proofs and the lack of need for any proofs about `peel`, `unpeel`, `mkpair`, or `unpair`: down about 75% from a combined 764 LOC to 192.


blaxill (approved):
Lgtm! Great to see that its smaller/simpler

PR #573, opened 2021-02-22
Move control circuit to new Circuit

blaxill:
- Uses instantiated versions of `aes_shift_rows`,`aes_mix_columns` etc, this is necessary to prevent excessive netlist generation (and non termination when generating SystemVerilog)

### DIFF
--- a/silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
+++ b/silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
@@ -25,6 +25,7 @@ Require Import ExtLib.Structures.Monads.
 Import MonadNotation.
 
 Require Import Cava.Cava.
+Require Import Cava.Acorn.Circuit.
 Require Import Cava.Acorn.Acorn.
 Require Import Cava.Acorn.Combinators.
 Require Import AcornAes.Pkg.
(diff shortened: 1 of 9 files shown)


silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
```diff
@@ -170,29 +156,31 @@ Section WithCava.
         initial_rcon <- initial_rcon_selector is_decrypt ;;
 
         (* we only need to grab the state at the last round *)
-        st <- cipher round_final round_0 is_decrypt initial_rcon initial_key
-                    initial_state round ;;
+        (* st <- cipher round_final round_0 is_decrypt initial_rcon initial_key *)
+        (*             initial_state round ;; *)
+        let st := initial_state in
```
blaxill, line 162: @jadephilipoom The new Circuit type separates combinational/sequential parts- I'd like to just instantiate the cipher here like before. Any thoughts on how to restructure this without rewriting completely?

jadephilipoom: You should be able to do something like
```coq
(Comb (<code before cipher> ... ret (cipher_input_data, other_data))
>==> First cipher
>==> Comb (fun '(st, other_data) => <code after cipher>)
```
...where `>==>` is `Circuit.Compose` (you can get it by importing `Circuit.Notations`). Because there's a loop, `cipher` can't be inside `Comb`, but you can do all the combinational processing on either side with `Comb`.

jadephilipoom (approved):
LGTM modulo the change to `Comb >==> cipher >==> Comb`.

Issue #566, opened 2021-02-19
Port AES proofs to new sequential setup


jadephilipoom:
See #561 -- the AES proofs should be changed so that combinational proofs use the `Cava` instance in `Combinational.v`, which has `combType` for signals instead of `seqType`. For combinational proofs (i.e. everything under the top-level cipher loop), the changes should be pretty minimal; will mostly involve replacing singleton lists `[x]` with just `x`. `Cipher.v` should be replaced with the parallel file `CipherNewLoop.v`, which contains the same definitions in the new setup. The two main tasks are
- [x] Port the cipher core implementation
- [x] Port proofs about  the top-level loop  (`CipherEquivalence.v` and `AES256Equivalence.v`)

PR #574, opened 2021-02-22
Port CipherEquivalence.v

jadephilipoom:
Progress on #566

New, parallel version of `CipherEquivalence.v` with the `Circuit` structure. Along the way, I separated out `multistep` and a couple of proofs about it into its own file, added a new tactic `destruct_inner_pair_let`, and added some helper lemmas about lists.

### DIFF
--- a/cava/Cava/Acorn/Combinational.v
+++ b/cava/Cava/Acorn/Combinational.v
@@ -75,16 +75,3 @@ Instance CombinationalSemantics : Cava combType :=
     instantiate _ circuit := circuit;
     blackBox intf _ := ret (tupleInterfaceDefault (map port_type (circuitOutputs intf)));
 }.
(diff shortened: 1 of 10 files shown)


blaxill (approved):
Lgtm! Did the `CipherEquivalence` proofs copy over to the `CipherEquivalenceNew` or did you have to completely rewrite?

jadephilipoom:
> Did the `CipherEquivalence` proofs copy over to the `CipherEquivalenceNew` or did you have to completely rewrite?

Mostly they did, except the `cipher_loop_equiv` proof now has a helper proof that describes the single-step behavior (it wasn't strictly necessary, but I think it's nice to have). So the previously super long and complicated proof became two medium-length and simpler proofs.

Issue #577, opened 2021-02-23
Reduce expression duplication from index expression copying


satnam6502:
Find places in the code base that causes expressions to become large because indexing resulting in copying, and fix them with the use of `localSignal` for now, with a view to applying a better fix as described in #575

jadephilipoom:
Resolved by #596 -- indexing is now always wrapped in `localSignal` by construction.

Issue #575, opened 2021-02-23
Revert indexAt and indexConst to being circuits


satnam6502:
They used to be circuits which made resource usage clear. I changed them to expressions but they've ended up being used in a way that does not make resource usage clear, which is a shame because that is definitely something we want to have a firm control of. Revert to previous definitions. Hopefully minimal impact on proof.
```coq
(* Dynamic indexing *)
  indexAt : forall {t : SignalType} {sz isz: nat},
            signal (Vec t sz) ->              (* A vector of n elements of type signal t *)
            signal (Vec Bit isz) ->           (* A bit-vector index of size isz bits *)
            cava (signal t);                  (* The indexed value of type signal t *)
  (* Static indexing *)
  indexConst : forall {t: SignalType} {sz: nat},
               signal (Vec t sz) ->           (* A vector of n elements of type signal t *)
               nat ->                         (* Static index *)
               cava (signal t);               (* The indexed bit of type signal bit *)
```

CC: @blaxill @jadephilipoom

blaxill:
Additionally `slice`, `unsignedAdd`, `unsignedMult`, `greaterThanOrEqual` are not assigned wires and could result in duplication? (Also pairing but paring is to be removed)

satnam6502:
> Additionally `slice`, `unsignedAdd`, `unsignedMult`, `greaterThanOrEqual` are not assigned wires and could result in duplication? (Also pairing but paring is to be removed)

Yes, indeed!

jadephilipoom:
Resolved by #596

PR #578, opened 2021-02-23
Reduce size of aes_mix_columns.sv from 25307 to 9939 lines

satnam6502:
Part of addressing #577 

The fix is simple: find places where an indexing expression would get copied, and pin it to a name with the use of `localSignal`, which should have identity function semantics for the proofs (in the `ident` monad).

@blaxill ~ please can you continue in other PRs to track down other places where duplication is occurring? I can tell from the generated SystemVerilog that we are duplicating vector literals, and in any case even though the size of generated file has been more than halved, I still feel it should be a lot shorter than 9939 lines. 

CC: @jadephilipoom ~ it might be worth adding a bit of proof automation for `localSignal`?
I am thinking of changing `indexAt` and `indexConst` to be circuits, which would then not require `localSignal` (see #575 )

### DIFF
--- a/silveroak-opentitan/aes/Acorn/MixColumnsCircuit.v
+++ b/silveroak-opentitan/aes/Acorn/MixColumnsCircuit.v
@@ -52,14 +52,15 @@ Section WithCava.
        (data_i[@3], data_i[@2]);
        (data_i[@2], data_i[@1]);
        (data_i[@1], data_i[@0])] in
-    x <- mapT xorV x ;;
+    x'' <- mapT xorV x ;;
+
 
     (* // Mul2(x) *)
     (*   for (genvar i = 0; i < 4; i++) begin : gen_x_mul2 *)
     (*     assign x_mul2[i] = aes_mul2(x[i]); *)
     (*   end *)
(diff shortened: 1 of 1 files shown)


jadephilipoom:
Since `localSignal` is just an identity, we could add it to the `cbn` list in the `simpl_ident` tactic, which will just inline it. Since `simpl_ident` is used in most proofs, I think this would end up requiring very little change to proofs.

blaxill:
I'm keen to fix this in the primitives as described in #575

blaxill (approved):
Perhaps `(un)peel` should also be circuits? If `localSignal` is always wanted after peel/unpeel.

satnam6502:
> Perhaps `(un)peel` should also be circuits? If `localSignal` is always wanted after peel/unpeel.

Yes, I wondered about that too.

satnam6502:
Another thing I did here was to rename a few signals to prevent name reuse e.g.
Chaging:
```coq
  x <- wombat x ;;
```
to:
```coq
  x' <- wombat x ;;
```
I propose that we never "reuse" names.

PR #579, opened 2021-02-23
Port AESEquivalence.v to Circuit

jadephilipoom:
Progress on #566 

Adds a duplicate of `AESEquivalence.v` with the new `Circuit` construction. The proofs changed pretty minimally, but some of the preconditions got simpler. I ended up changing the order of arguments to `cipher` slightly to make the arguments nicer to construct, and tweaking the `cipher_equiv` proof so that we ignore the "initial state" input values after the first cycle.


blaxill (approved):
Lgtm!

PR #581, opened 2021-02-23
Key expansion loop

### LINKED ISSUE #414: Simplify precomputed-vs-interleaved key preconditions
See #412 

In both the arrow and acorn top-level cipher proofs, there are lengthy and awkward preconditions that relate the on-the-fly key expansion function used by the implementation to the precomputed keys required by the spec. These can probably be made less painful with some changes to `ExpandAllKeys.v`.

### LINKED ISSUE #560: Make round constants internal to key expansion
See #545 

Round constants shouldn't be visible to the cipher or cipher control; they should be internal to key expansion. Since we currently are leaving key expansion out of scope, we should just be able to remove round constants from everything else, including the key expansion type signature , and everything should work.

jadephilipoom:
Resolves #560 
Resolves #414 

Makes `key_expand` a sequential `Circuit` instead of a loop, with the round constant as part of its internal state. This actually simplified the top-level lemma statements quite a bit and means we no longer need to use the "interleaved" version of the spec as an intermediate step.

### DIFF
--- a/silveroak-opentitan/aes/Acorn/AES256EquivalenceNew.v
+++ b/silveroak-opentitan/aes/Acorn/AES256EquivalenceNew.v
@@ -102,9 +102,10 @@ Hint Resolve add_round_key_equiv sub_bytes_equiv shift_rows_equiv
      mix_columns_equiv : subroutines_equiv.
 
 Definition full_cipher {signal} {semantics : Cava signal}
+           key_expand
   : Circuit
       (signal Bit * signal round_index * signal round_index * signal round_index
-       * (signal key * signal round_constant * signal state))
+       * signal key * signal state)
       (signal state) :=
   cipher
     aes_sub_bytes aes_shift_rows aes_mix_columns aes_add_round_key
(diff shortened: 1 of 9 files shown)


silveroak-opentitan/aes/Acorn/AES256EquivalenceNew.v
```diff
@@ -134,179 +135,122 @@ Local Ltac solve_side_conditions :=
...
-                          nat_to_bitvec_sized _ 0, nat_to_bitvec_sized _ i))
-                      (seq 0 (S Nr)))
-                   (init_cipher_state :: init_cipher_state_ignored) ->
+                   (combine
+                      (map
+                         (fun i =>
+                            (is_decrypt, nat_to_bitvec_sized _ Nr,
+                             nat_to_bitvec_sized _ 0, nat_to_bitvec_sized _ i))
+                         (seq 0 (S Nr)))
+                      init_key_input)
```
satnam6502, line 157: It's a bit hard to see what is being `combine`d here, perhaps a named local binding for the `map` expression? I appreciate the indentation perhaps helps enough here to identify `init_key_input` as the second argument to `combine`. No strong feeling so feel free to disrgard.

satnam6502 (approved):
The proofs are beyond me ~ my comments pertain to the circuits!

silveroak-opentitan/aes/Acorn/AES256EquivalenceNew.v
jadephilipoom: Since this pattern reappeared a lot, I added a definition `make_cipher_signals` to do the packaging of the input signals, which made a lot of the lemma statements more concise. Thanks for pointing it out!

silveroak-opentitan/aes/Acorn/CipherEquivalenceNew.v
```diff
@@ -348,159 +316,104 @@ Section WithSubroutines.
...
-                        (fun i =>
-                           (true, nat_to_bitvec_sized _ Nr,
-                            nat_to_bitvec_sized _ 0, nat_to_bitvec_sized _ i))
-                        (seq 0 (S Nr)))
-                     (init_cipher_state :: init_cipher_state_ignored) ->
+    length init_key_input = S Nr ->
+    length init_state_ignored = Nr ->
+    length middle_keys = Nr - 1 ->
+    cipher_input = make_cipher_signals Nr is_decrypt init_key_input
+                                       (init_state :: init_state_ignored) ->
```
jadephilipoom, line 379: This is what the new preconditions on `cipher_input` look like.

PR #582, opened 2021-02-23
Change CavaClass methods to always return circuits

blaxill:
Just a quick attempt at converting (un)peel, indexConst, indexAt etc to return circuits to see the impact. Unfortunately it breaks a lot of proofs.
@satnam6502 @jadephilipoom  any thoughts on what direction to take this? 

Related #575


satnam6502:
Thanks for doing this. I think this results in "less nice" descriptions, but makes resource usage clear and helps to avoid expression duplication.

cava/Cava/Acorn/NetlistGeneration.v
```diff
@@ -313,12 +316,12 @@ Instance CavaCombinationalNet : Cava denoteSignal := {
     unpair := @unpairNet;
     peel := @peelNet;
     unpeel := @unpeelNet;
-    indexAt k sz isz := IndexAt;
-    indexConst k sz := IndexConst;
+    indexAt k sz isz v i := localSignalNet (IndexAt v i);
+    indexConst k sz v i := localSignalNet (IndexConst v i);
```
satnam6502, line 320: An alternative is to revert `IndexAt` and `IndexConst` to being constructors of `Instance`. But first let's see if this works for avoiding duplication.

satnam6502 (approved):

jadephilipoom:
The proofs should really not be hard to fix, since the added `localSignal` will unfold to the identity. If you add `localSignal` to the `cbn` list in `simpl_ident` tactic, I think many of the proofs will work again, and the ones that don't will require only minor modification (e.g. adding a call to `simpl_ident`).

`simpl_ident` currently exists in two locations because of the ongoing port to `Circuit`, so here are the two places to add `localSignal`:
https://github.com/project-oak/silveroak/blob/main/cava/Cava/Acorn/Identity.v#L190
https://github.com/project-oak/silveroak/blob/8f7d6f5a3e83c1b3da83766999b14f0eac62866b/cava/Cava/Acorn/IdentityNew.v#L69-L71

blaxill:
Note unfortunately a number of definitions need to be changed such as `peel (a::as) = a :: peel as`

satnam6502:
> Note unfortunately a number of definitions need to be changed such as `peel (a::as) = a :: peel as`

I suspect we don't _need_ to make `peel` and `unpeel `circuits.

satnam6502:
I suspect this PR is now defunct.

PR #588, opened 2021-02-24
Recreate OpenTitan aes_cipher_control state machine

blaxill:
I'd hoped we could use a simpler `aes_cipher_control` with simpler signalling but after running some tests with Verilator and inspecting how the `aes_control` uses `aes_cipher_control` by observing the waveforms between the two its clear we need the full signalling. This PR removes the simpler `aes_cipher_control` and models the transition logic of the OpenTitan directly. A follow up PR will finish up connecting the pieces together again. 

What does "full signalling" mean and what does "simpler siognalling" mean? Is it's not too much bother can you share the waveform screen grabs or other output? I'm keen to understand what is going on here. Don't hold the PR up for this.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "third_party/coq-tools"]
 	path = third_party/coq-tools
 	url = https://github.com/JasonGross/coq-tools.git
+[submodule "third_party/coq-record-update"]
+	path = third_party/coq-record-update
+	url = https://github.com/tchajed/coq-record-update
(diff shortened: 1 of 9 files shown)


blaxill:
A couple of things could be cleaned up, commit now so as to not drift to far from `main`

satnam6502 (approved):

blaxill:
>What does "full signalling" mean and what does "simpler signalling" mean? 

Since we are focused on AES256 we have some hardwired scheduling of resources that is more dynamic in OT. Additionally things not covered by the FIPS spec but is important to physical security I avoided for a first implementation such as clearing registers, and filling with pseudo random numbers.

Issue #567, opened 2021-02-19
Clean up old sequential files and remove Pair from SignalType


jadephilipoom:
Once #565 and #566 are complete:
- Remove files that should be no longer needed: `CombinationalMonad.v`, `Sequential.v` `CombinationalProperties.v`, and `SequentialProperties.v`
- Remove `Pair` from SignalType
- Remove `mkpair` and `unpair` from CavaClass
- Do any renaming that make sense (e.g. `CombinationalPropertiesNew.v` can be renamed to `CombinationalProperties.v` once the latter is removed).

Also resolves #416

PR #589, opened 2021-02-24
Port Acorn proofs to new sequential circuit setup.

smore-lore:
Main notes:
- not sure why 'CipherNewLoop' has this name. Should we drop the 'New' here as well?
- CombinationalPropertiesNew/IdentityNew packages aren't renamed since there are still a few tests which use these
- MixColumns proof still has an 'Admitted' Lemma, though I've replaced the Axiom here


jadephilipoom:
> not sure why 'CipherNewLoop' has this name. Should we drop the 'New' here as well?
> CombinationalPropertiesNew/IdentityNew packages aren't renamed since there are still a few tests which use these

The "New" names are temporary; these files are intentionally duplicated right now because the port to `Circuit` isn't finished. If all the Acorn code is ported it's now okay to replace the old `CipherEquivalence` and `AESEquivalence`, but `IdentityNew` and `CombinationalPropertiesNew` will be renamed later as part of #567. I'll go ahead and merge this as-is and do the `CipherNewLoop.v` => `Cipher.v` replacement separately.

silveroak-opentitan/aes/Acorn/AddRoundKeyEquivalence.v
```diff
@@ -31,7 +31,63 @@ Require Import AesSpec.StateTypeConversions.
 Require Import AcornAes.AddRoundKeyCircuit.
 Import StateTypeConversions.LittleEndian.
 
+(*
```
jadephilipoom, line 34: Remove this comment?

silveroak-opentitan/aes/Acorn/AES256EquivalenceNew.v
```diff
@@ -1,246 +0,0 @@
-(****************************************************************************)
```
jadephilipoom, line 1: The diff right now makes it very hard to tell if the file was moved and changed or just moved; I had to check out your branch and do a diff of the files from main and here manually to see the true diff. Did you use `git mv` or just `mv`? Usually `git mv` is nicer because the diff shows a change to the file _name_ instead of a deleted file and an added file, and preserves the file history -- I'm not 100% certain about the case where the destination already exists, though, so forgive me if you did use it here.

jadephilipoom (approved):
LGTM once the commented-out lemma is removed

silveroak-opentitan/aes/Acorn/AES256EquivalenceNew.v
smore-lore: 'git mv' won't actually change how git interprets the file. In this case, it would still show a removed file + a diff file. However, I thought that github would show the overall diff, so that having this specific diff for this specific commit would be distracting, since it would have the differences reflected in two places. That seems to be my misunderstanding.

Anyway for git blame - the squashed commit here should do the right thing (let's see!).

PR #592, opened 2021-02-25
Port remaining tests and examples

### LINKED ISSUE #565: Port tests to new sequential setup
See #561 -- code under `tests` should be ported to use the new `Circuit` inductive so that the old `seqType`-based `Cava` instance can be eventually removed.

jadephilipoom:
Resolves #565 

There might still be some small cleanup to do, but all the proofs and large examples are now ported to `Circuit`. Once #589 is merged, we're clear to start #567.


blaxill (approved):
Lgtm

Issue #590, opened 2021-02-24
Prove AES256Equivalence Axiom key_expand_equiv


smore-lore:
I noticed we don't have this, while trying to port proofs over to know sequential circuit setup.

Are there any other axioms we need to finalize phase I?

jadephilipoom:
Phase 1 intentionally excludes key expansion (see the description on the project: "minus key expansion"); the axiom is expected to remain. We decided at the start of phase 1 to narrow the scope as much as possible, and since key expansion is relatively separate from the rest of the cipher (and also an "optional" part of a hardware AES block --  see discussion [here](https://docs.opentitan.org/hw/ip/aes/doc/) under "Datapath Architecture and Operation"). It may seem counter-intuitive to leave out key expansion, but the goal here isn't to produce a production-ready system; we just want a case study for our methodology.

PR #593, opened 2021-02-25
Remove old sequential code

jadephilipoom:
Progress on #567 

This PR rips out all the old sequential code and renames the duplicate files to take the place of old ones. Still to do on #567 is to remove `Pair`.

Interestingly, the new construct for key expansion from #581 means we _don't_ depend on the key expansion axioms any more -- I stand corrected in #590. Instead, the final lemmas take in an argument that is the `key_expand` implementation, and require the precondition that (for `full_cipher_equiv`) it matches the spec's precomputed keys, or (for `full_cipher_inverse`) the decryption keys are the reverse of the encryption keys. So key expansion is still out of scope, but we don't have `Axiom`s about it -- and don't need to know very much about it to prove that our cipher works.

Best reviewed commit-by-commit; a lot of code was deleted here!


acorn-examples/xilinx/NandLUT.v
```diff
@@ -41,10 +42,7 @@ Definition lutNANDNetlist := makeNetlist lutNANDInterface lutNAND.
  [(false, false); (false, true); (true, false); (true, true)].
 
  Definition lutNAND_tb_expected_outputs : list bool :=
-   map (fun '(i0,i1) =>
-          List.hd (defaultCombValue _)
-                  (combinational (lutNAND ([i0],[i1])%list)))
-       lutNAND_tb_inputs.
+   multistep (Comb lutNAND) lutNAND_tb_inputs.
```
satnam6502, line 45: Much nicer!

acorn-examples/xilinx/XilinxAdderExamples.v
```diff
@@ -42,24 +43,24 @@ Definition v44  := N2Bv_sized 8 44.
 
 (* Perform a few basic checks to make sure the adder works. *)
 
-Example xadd_17_52_0 : combinational (xilinxAdderWithCarry ([false], ([v17], [v52]))) =
-                                     ([v69], [false]).
+Example xadd_17_52_0 : unIdent (xilinxAdderWithCarry (false, (v17, v52))) =
+                       (v69, false).
```
satnam6502, line 47: Great!

satnam6502 (commented):
A great improvement. Thank you.

satnam6502 (approved):

PR #594, opened 2021-02-26
Finish control loop

blaxill:
Follow up PR has the logic that ties the cipher and cipher_control together


jadephilipoom (approved):
LGTM!

PR #595, opened 2021-02-26
Remove pairs from SignalType

### LINKED ISSUE #416: Attempt to remove tuples from SignalType

### LINKED ISSUE #567: Clean up old sequential files and remove Pair from SignalType

jadephilipoom:
Resolves #416 
Resolves #567 

With the new `Loop` infrastructure, no pairs are needed in `SignalType`! This PR removes them.

### DIFF
--- a/cava/Cava/Signal.v
+++ b/cava/Cava/Signal.v
@@ -27,7 +27,6 @@ Inductive SignalType :=
   | Void : SignalType                              (* An empty type *)
   | Bit : SignalType                               (* A single wire *)
   | Vec : SignalType -> nat -> SignalType              (* Vectors, possibly nested *)
-  | Pair : SignalType -> SignalType -> SignalType    (* A pair of signals *)
   | ExternalType : string -> SignalType.            (* An uninterpreted type *)
 
 (******************************************************************************)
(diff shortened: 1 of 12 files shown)


blaxill (approved):

cava/Cava/Signal.v
```diff
@@ -27,7 +27,6 @@ Inductive SignalType :=
   | Void : SignalType                              (* An empty type *)
   | Bit : SignalType                               (* A single wire *)
   | Vec : SignalType -> nat -> SignalType              (* Vectors, possibly nested *)
-  | Pair : SignalType -> SignalType -> SignalType    (* A pair of signals *)
```
satnam6502, line 30: Ah, the deletion I've been waiting for so long.

PR #596, opened 2021-02-26
Update CavaClass to use circuits for indexing and arithmetic

jadephilipoom:
Updated version of #482 

Doesn't change `peel`/`unpeel`, since in discussion we weren't sure if they would need to change. @blaxill @satnam6502 , does this solve the duplicated-code problem you were seeing? Should we modify `peel`/`unpeel` as well?

### DIFF
--- a/acorn-examples/IncrDecr.v
+++ b/acorn-examples/IncrDecr.v
@@ -51,19 +51,27 @@ Section WithCava.
   (* increment a 4-bit vector *)
   Definition incr4 (input : signal (Vec Bit 4))
     : cava (signal (Vec Bit 4)) :=
-    '(sum0, carry) <- half_adder (one, indexConst input 0) ;;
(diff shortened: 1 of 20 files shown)


blaxill:
Peel & unpeel will be generating unnecessary items as e.g. peel generates `n` netlist constructors for an `n` element vector as: `peel := Vector.map (IndexConst v) (vseq 0 s)`. This will be duplicated for every usage of an expression defined as the result of `peel`, whereas a monadic version can generate an `assignSignal` that will share a single definition. 

**However** I haven't benchmarked the impact of this and it can be mitigated by immediately passing through `localSignal`.

blaxill (approved):
LGTM!

jadephilipoom:
Okay, I'll prepare another PR with the `peel`/`unpeel` as circuits too!

PR #597, opened 2021-02-26
Make peel and unpeel into circuits

jadephilipoom:
Follow up for #582, #596

Make `peel` and `unpeel` circuits so that they can be essentially wrapped in `localSignal` always.

@blaxill I was kind of guessing with the naming in `CipherControlCircuit`, feel free to change anything that doesn't match your conventions!


blaxill (approved):
Lgtm thanks Jade!

PR #598, opened 2021-02-26
Finish untested aes_cipher_core

blaxill:
Generates a reasonable looking netlist for `aes_cipher_core` but a bit more in-tree OpenTitan testing required. Specifically, the control signals should be correct now, but how the final state is held and some timings might need to be tweaked.

### DIFF
--- a/silveroak-opentitan/aes/Acorn/CipherCircuit.v
+++ b/silveroak-opentitan/aes/Acorn/CipherCircuit.v
@@ -49,7 +49,6 @@ Section WithCava.
       * signal round_index (* num_rounds_regular : round index of final round *)
       * signal round_index (* round_0 : round index of first round *)
       * signal round_index (* current round_index *)
-      * signal key (* initial key, ignored for all rounds except first *)
       * signal state (* initial state, ignored for all rounds except first *).
 
   Definition cipher_round
@@ -104,7 +103,7 @@ Section WithCava.
             let '(input, k, feedback_state) := input_and_state in
             (* extract signals from the input tuple *)
             let '(is_decrypt, num_rounds_regular,
-                  round_0, idx, _, initial_state) := input in
+                  round_0, idx, initial_state) := input in
             is_first_round <- idx ==? round_0 ;;
             st <- muxPair is_first_round (feedback_state, initial_state) ;;
             st' <- cipher_step is_decrypt is_first_round
--- a/silveroak-opentitan/aes/Acorn/CipherCircuit.v
+++ b/silveroak-opentitan/aes/Acorn/CipherCircuit.v
@@ -49,6 +49,7 @@ Section WithCava.
       * signal round_index (* num_rounds_regular : round index of final round *)
       * signal round_index (* round_0 : round index of first round *)
       * signal round_index (* current round_index *)
+      * signal key (* initial key, ignored for all rounds except first *)
       * signal state (* initial state, ignored for all rounds except first *).
 
   Definition cipher_round
@@ -103,7 +104,7 @@ Section WithCava.
             let '(input, k, feedback_state) := input_and_state in
             (* extract signals from the input tuple *)
             let '(is_decrypt, num_rounds_regular,
-                  round_0, idx, initial_state) := input in
+                  round_0, idx, _, initial_state) := input in
             is_first_round <- idx ==? round_0 ;;
             st <- muxPair is_first_round (feedback_state, initial_state) ;;
             st' <- cipher_step is_decrypt is_first_round
--- a/silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
+++ b/silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
@@ -663,6 +663,7 @@ Section WithCava.
       * signal Bit (* crypt_o *)
       * signal Bit (* dec_key_gen_o *)
       * signal Bit (* key_clear_o *)
+      * signal Bit (* data_out_clear_o *)
       * signal state (* state_o *)
       ) :=
     Loop (
@@ -703,14 +704,14 @@ Section WithCava.
              , crypt
              , dec_key_gen
              , key_clear
-             , _
+             , data_out_clear
              , current_round
         ) := control_signals in
 
         r13 <- round_13 ;;
         r0 <- round_0 ;;
         ret ( (op, r13, r0, current_round, st, k, k)
-            , (in_ready, out_valid, crypt, dec_key_gen, key_clear))
+            , (in_ready, out_valid, crypt, dec_key_gen, key_clear, data_out_clear))
 
       ) >==>
       First cipher_loop >==>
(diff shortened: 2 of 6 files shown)


silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
```diff
@@ -655,45 +628,102 @@ Section WithCava.
...
+    Circuit (signal Bit * signal Bit * signal Bit * signal round_index * signal (Vec Bit 3) * signal key)
+            (signal key)).
+  Context (cipher_loop:
+    Circuit (
+    signal Bit (* op_i/is_decrypt : true for decryption, false for encryption *)
+      * signal round_index (* num_rounds_regular : round index of final round *)
+      * signal round_index (* round_0 : round index of first round *)
+      * signal round_index (* current round_index *)
+      * signal state (* initial state, ignored for all rounds except first *)
+      * signal key ) (signal state)).
```
jadephilipoom, line 652: Comment here explaining what the `signal key` is for? Is it the initial key?

silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
```diff
@@ -655,45 +628,102 @@ Section WithCava.
...
+             , (op, key_len, prng, st, k, feedback_state)) := inputs in
+
+        let '( _ , _ , _ , _ , _
+             , key_expand_step
+             , key_expand_clear
+             , _ , _ , _ , _ , _ , _
+             , key_expand_op
+             , key_expand_round
+             , _ , _ , _ , _ , _ ) := control_signals in
+        (*FIXME(blaxill): Fix k *)
```
jadephilipoom, line 688: What does this mean?

silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
```diff
@@ -655,45 +628,102 @@ Section WithCava.
...
+        ret ( ( key_expand_op, key_expand_step, key_expand_clear, key_expand_round, key_len, k)
+            , (control_signals, op, st, feedback_state) )
+      ) >==>
+      First key_expand >==>
+
+      (* 2. Run cipher_loop: Place key_expand signals into the right place *)
```
jadephilipoom, line 694:
```suggestion
      (* 2. Run cipher_loop: Place cipher signals into the right place *)
```

silveroak-opentitan/aes/Acorn/CipherCircuit.v
```diff
@@ -49,7 +49,6 @@ Section WithCava.
       * signal round_index (* num_rounds_regular : round index of final round *)
       * signal round_index (* round_0 : round index of first round *)
       * signal round_index (* current round_index *)
-      * signal key (* initial key, ignored for all rounds except first *)
```
jadephilipoom, line 52: If the goal is to use `cipher` instead of `cipher_loop` at some point, it might not be a good idea to remove this. I left the `signal key` in the cipher signals because I figured key expansion would need an initial key signal just like the cipher needs an initial state signal, and I think I see that reflected in the current control circuit as well. Does it affect netlist generation to have this piece of data available to both key expansion and the cipher loop even though only the latter uses it? I wrote it this way for the convenience of using the same datatype for both loops; figured that just making the wire accessible inside a loop would be "free" from a netlist standpoint. It would be good for me to know if that assumption is wrong, so I can avoid the pattern in the future!

jadephilipoom (commented):
I'd like to clarify whether we want to remove the key signal or not before merging.

silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
blaxill: `k` does not follow the exact signalling of OT and might need to be fixed

silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
blaxill: Yeah the comment got lost but the definition is the same as CipherCircuit, its just introduced via context so I don't have to plug in the definitions for submodules (e.g. mixcolumns) or abstract over them as in CipherCircuit. (Currently I want to blackbox the submodules and this is only available working with Signal vs signal)

silveroak-opentitan/aes/Acorn/CipherCircuit.v
blaxill: In terms of resource usage looped items generate registers. Since we weren't using the key here there is extra resources generated.

The choice here is really where the feedback path is (rather than initial), the key expansion does need to be fed last iteration results which is in the control logic here. The more "control" logic from OT that we move away from the cipher control the harder getting the timings of the implementation right are. 

E.g. moving the `cipher_loop` stateful feedback and mux `is_first_round` to control logic would make the control logic simpler but I presume would make the proof more complex as you'd have to reason over the ugly state machine logic in this PR.

jadephilipoom: I can see why an element added to the loop _state_ would create a register,  but loop _input_ should not generate registers. These signals are loop input, so they should be just "things the loop body may access"; they are not on the feedback wire, and netlist generation shouldn't create a register for them. If we're getting registers for them then something is going wrong.

blaxill: Yeah sorry misread. 

We have `key` twice, once implicitly in `cipher_signals `, once explicitly in input of `cipher_loop`, I dont think we need both? The change could be the other way round removing from `cipher_loop` and leaving in `cipher_signals`. Removing it as a proof hypothesis seemed a benefit to me but happy for either way round.

jadephilipoom: It's true that the `key` signal in `cipher_signals` is the initial key, which is not strictly needed for the cipher. However, it is presumably needed for key expansion -- I think if you just remove the initial key from `cipher_signals`, then it becomes impossible to write a `key_expand` that can have the type signature required by `cipher`.

The alternative option is to essentially define two versions of `cipher_signals` : one for the `cipher_loop`, which is the same as currently but without the initial key, and one for `key_expand`, which would be the same except without the initial state signal (since `key_expand` doesn't need it). However, as long as carrying around the extra signal doesn't cause problems in the netlist, it's convenient to not define two different versions of `cipher_signals`. Having the extra parameter in proofs doesn't really weaken them, since there are no preconditions for it; all you need to say is that the initial key signal exists.

blaxill: I've reverted the change, I'm not sure I buy the reasoning though- whats the benefit of the key_expand in cipher right now? The OT version includes additional signalling- I think if we do want to model the whole key_expand in Coq then the number of differing signals between ciper_loop/key_expand would grow.

jadephilipoom: The benefit of putting the two together is that then we can do proofs about the composition. If they should be composed in a different way or if `key_expand` should be part of the control logic, then we should discuss changing the proofs to be about `cipher_loop` instead of `cipher` (which I think is doable).

PR #599, opened 2021-03-01
Create test benches based on FIPS for subroutine netlists

jadephilipoom:
Currently, we're testing that the Coq semantics for our circuits match the test vectors from FIPS, but the actual test benches for the netlists only test one or two handwritten examples for each subroutine. This PR adds much more extensive test benches based on the same test data as the semantics tests.


satnam6502:
I've had a quite skim but I'll defer to @blaxill for the actual review. Thanks.

satnam6502 (approved):

PR #600, opened 2021-03-01
Reorganize/rename multiplexers

### LINKED ISSUE #450: Define a mux2 library element that operates over pairs defined in terms of indexAt

jadephilipoom:
A bit of cleanup. Now that we don't need `muxPair` to define loops, we can move it back to `Lib` (and move `mux4` there too). And since we don't have pairs any more, I renamed `muxPair` to `mux2`.

Resolves #450


blaxill (approved):
Apart from CI failing, lgtm

PR #607, opened 2021-03-03
Vec library

jadephilipoom:
While trying to write a tutorial for Cava, I noticed myself writing some boilterplate `peel`s and `unpeel`s that followed similar patterns -- convert to `Vector.t`, do a vector operation, and then convert back to `Vec`. I've had the thought before that it would be nice if we had a clean library with all the `Vector.t` operations we need that does the `peel` and `unpeel` work in the background, so here it is. Now, instead of `peel >=> mapT f >=> unpeel`, you can write `Vec.map f`, and instead of `v' <- peel v ;; unpeel (x :: v')` you can write `Vec.cons x v`, and so on. There's also `Vec.bitvec_literal` for quick literal bit-vectors. The `simpl_ident` tactic will convert these definitions back to `Vector.t` automatically in proofs.

- Defines circuits for all of the `Vector.t` operations (there aren't that many) on `Vec` (in `Cava/Lib/Vec.v`)
- Proves the circuits equivalent to the standard library implementations
- Adjusts tests, examples, and AES circuits to use the `Vec` implementations

### DIFF
--- a/acorn-examples/IncrDecr.v
+++ b/acorn-examples/IncrDecr.v
@@ -27,6 +27,8 @@ Require Import Cava.BitArithmetic.
 Require Import Cava.NatUtils.
 Require Import Cava.Tactics.
 Require Import Cava.VectorUtils.
+Require Import Cava.Lib.VecProperties.
+Require Cava.Lib.Vec.
 Import VectorNotations ListNotations MonadNotation.
 Open Scope monad_scope.
 Open Scope list_scope.
@@ -80,13 +82,11 @@ Section WithCava.
           signal (Vec Bit sz0) -> cava (signal (Vec Bit sz0)) with
     | 0 => fun input => ret input
     | S sz' => fun input : signal (Vec Bit (S sz')) =>
-                i <- peel input ;;
-                let i0 := Vector.hd i in
+                i0 <- Vec.hd input ;;
+                rem <- Vec.tl input ;;
                 '(sum0, carry) <- half_adder (carry, i0) ;;
-                rem <- unpeel (Vector.tl i) ;;
-                sum' <- incr' carry rem ;;
-                sum <- peel sum' ;;
-                unpeel (sum0 :: sum)%vector
+                sum <- incr' carry rem ;;
+                Vec.cons sum0 sum
     end.
 
   (* increments a bit vector of any length *)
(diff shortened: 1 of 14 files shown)


blaxill (approved):
It might be useful to provide a generic

~~~
Definition withPeeled {a b n m}
  (f : Vector.t (signal a) n -> cava (Vector.t (signal b) m))
  : signal (Vec a n) -> cava (signal (Vec b m)) :=
  peel >=> f >=> unpeel.
~~~

If Coq had monadic folds (e.g. like https://hackage.haskell.org/package/base-4.14.1.0/docs/Control-Monad.html#v:foldM ) then `fold_left` etc could be defined as `withPeeled (fold_left f v) i`.

Thinking even more wishfully, if coq-ext-lib provided functors/monads on categories other than Coq types/functions, `withPeeled` would just become `sequence` from a more flexible `Traversable`.

jadephilipoom:
I did consider making something like `withPeeled`, but I found that it just wouldn't be possible to use it in very many places. For instance, `fold_left` couldn't actually be defined in the way you describe, since the state type is not necessarily a vector. The type signature is:
```coq
forall {A B} (f : B * signal A -> cava B)  {n}, signal (Vec A n) -> B -> cava B 
```
For `withPeeled` to work here, `B` would need to be a vector, which is not always the case. We would essentially need different versions of `withPeeled` for cases when we want to return a vector versus return a more abstract type, and given that there weren't all that many vector definitions in the first place we could only use each version once or twice. It  just didn't seem worth the tradeoff to add an extra layer of definition, which inherently makes things just a little more complicated to simplify/understand/prove. 

As for `foldM`, we actually do have monadic folds in our own combinator library (for lists, though):
https://github.com/project-oak/silveroak/blob/885f88a4cb1878e7839ed4727ee538176d5b6001/cava/Cava/Acorn/Combinators.v#L82-L86

blaxill:
> For instance, fold_left couldn't actually be defined in the way you describe ... For withPeeled to work here, B would need to be a vector, which is not always the case. We would essentially need different versions of withPeeled for cases when we want to return a vector versus return a more abstract type

There are ways of abstracting this, either collecting in a monoid or making a peelable type class that gives a reasonable interpretation to non-vecs (likely a no-op for other types). I'm not suggesting its necessarily worth the effort though.

> As for foldM, we actually do have monadic folds in our own combinator library (for lists, though):

I see that as another point for moving to lists vs vectors :)

satnam6502:
This looks great. I am not keen on the names `peel` and `unpeel` because it is hard to tell _what_ is being peeling (i.e. the direction of the peeling). What is really going on here is some kind of "packing" and "unpacking" so some names like `pack` and `unpack` (but a bit more specialized, since `pack` and `unpack` are common names so likely to clash) might be better?

Issue #610, opened 2021-03-03
Renaming pass


jadephilipoom:
A few renames would be nice to do as part of our code cleanup:

- [x] `peel`/`unpeel` -> `packv`/`unpackv`
- [ ] `multistep` -> `simulate`

Any others on someone's mind, add it here!

PR #612, opened 2021-03-03
Rename peel/unpeel to packv and unpackv

jadephilipoom:
Progress on #610

I have put `packv` for what used to be `unpeel` -- packaging a Coq vector into a signal -- and `unpackv` for its inverse. Does that match others' intuitions?


satnam6502:
How do you feel about `packV` and `unpackV`?

satnam6502 (approved):

jadephilipoom:
> How do you feel about `packV` and `unpackV`?

No objection. I'll change it.

PR #623, opened 2021-03-04
aes_cipher_core: Forward mode working / odd bug with eqb

blaxill:
Note, this change is out of date but I'm leaving this draft here at the moment in case someone wants to investigate. I'm not sure if its a bug or a misunderstanding on my part.

Related : #647 appears to have been verilator bug

### DIFF
--- a/silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
+++ b/silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
@@ -59,6 +59,8 @@ Section WithCava.
     bitvec_to_signal (nat_to_bitvec_sized _ 1).
   Definition round_2: cava (signal round_index) :=
     bitvec_to_signal (nat_to_bitvec_sized _ 2).
+  Definition round_10: cava (signal round_index) :=
+    bitvec_to_signal (nat_to_bitvec_sized _ 10).
   Definition round_13: cava (signal round_index) :=
     bitvec_to_signal (nat_to_bitvec_sized _ 13).
   Definition round_14: cava (signal round_index) :=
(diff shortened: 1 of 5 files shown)


silveroak-opentitan/aes/Acorn/CipherControlCircuit.v
```diff
@@ -651,79 +692,102 @@ Section WithCava.
...
+        k0 <- indexConst last_kp 0 ;;
+        k1 <- indexConst last_kp 1 ;;
         r0 <- round_0 ;;
-        ret ( (op, r13, r0, current_round, st, k, k)
-            , (in_ready, out_valid, crypt, dec_key_gen, key_clear, data_out_clear))
+        r14 <- round_14 ;;
+
+        (* is_r0 <- eqb r0 round ;; (1* bugged? *1) *)
+        is_r0 <- eqb_bb  (r0, round) ;;
+        k <- mux2 is_r0 (k1, k0);;
```
blaxill, line 773: There is and odd bug here, if `Acorn.Combinators.eqb` is used _here_ directly the value is delayed by one cycle, there are other uses of `eqb` such as in `cipher_loop, which is part of this function, that get the correct value. Instantiating `eqb` as a blackbox produces the correct behaviour. 

Are we doing any deduplication that could be affecting this? I ask because using the commented out version, i.e. `eqb` directly causes the `eqb` in `cipher_loop` to stop working. I have observed the inputs and checked they are correct, its very odd behaviour. 

@satnam6502

blaxill: Hmm I might be wrong about the interaction between two instances, but its still true that using eqb directly at that line is delayed whereas using the same function in a blackbox is not delayed.

blaxill:
As per #647 appears to have been verilator bug

Issue #609, opened 2021-03-03
Tutorial for Cava


jadephilipoom:
Write a tutorial that walks through example Cava circuits and their proofs, explaining core concepts.

PR #625, opened 2021-03-05
Inverter tutorial

jadephilipoom:
Progress on #609

Example 1 in the tutorial: a simple inverter circuit that I use to demonstrate some foundational Cava concepts.

As before, the HTML is auto-generated, so the best way to review this is to review the Coq source file, which has the prose descriptions as special comments. When you see a `coq:: none` tag at the end of a comment, it means that the Coq code between this comment and the next one will not show up in the resulting webpage.

### DIFF
--- a/demos/tutorial.v
+++ b/demos/tutorial.v
@@ -22,6 +22,181 @@
 ========
 Tutorial
 ========
+
+Welcome! This is a quick primer for designing circuits with Cava. We'll walk
+through a few small examples end-to-end.
+
+.. coq:: none
 |*)
 
-Check nat.
+Require Import ExtLib.Structures.Monads.
+Import MonadNotation.
+Open Scope monad_scope.
+
+Require Import Cava.Acorn.Acorn.
+
+(*|
+Example 1 : Inverter
+====================
+
+To start, let's define a 1-bit inverter.
+|*)
+
+Definition inverter
+           {signal : SignalType -> Type}
+           {semantics : Cava signal}
+  : Circuit (signal Bit) (signal Bit) :=
+  Comb inv.
+
+(*|
+A few things to notice here:
+
+* ``SignalType`` is Cava's type system. The inverter is parameterized over
+  ``signal``, which converts ``SignalType``\ s to Coq types. ``Bit`` is one
+  example of a ``SignalType``; we'll see more examples later on.
+* ``Comb`` is short for "combinational"; our inverter has no loops, registers,
+  or timing requirements, so it is a purely combinational circuit.
+* The inverter is also paramterized over ``semantics``, an instance of the
+  typeclass ``Cava``. This instance provides implementations of circuit
+  primitives, such as 1-bit logic gates. One primitive is a 1-bit inverter
+  ``inv``, so our inverter is just a simple invocation of the primitive.
+
+Normally, we'd write circuit definitions a little more concisely by writing them
+inside a ``Section`` that contains ``signal`` and ``semantics`` as context
+variables, like this:
+|*)
+
+Section WithCava.
+  Context {signal} {semantics : Cava signal}.
+
+  Definition inverter_concise : Circuit (signal Bit) (signal Bit)
+    := Comb inv.
+
+(*|
+For the rest of the circuit definitions in this tutorial, you can assume that
+we're inside the section and that every definition is parameterized over the
+``signal`` and ``semantics`` context variables.
+
+Back to our inverter. Let's take a closer look at the ``inv`` primitive.
+|*)
+
+  About inv.
+
+(*|
+You can see in the type signature ``signal Bit -> cava (signal Bit)`` that
+``inv`` is defined as a pure Coq function in terms of a monad called
+``cava``. The ``cava`` monad, like ``inv``, is provided by ``semantics``. The
+monad is used to preserve sharing; it's semantically different in Cava to write::
+
+  x <- inv zero ;;
+  y <- inv zero ;;
+  xor2 x y
+
+than it is to write::
+
+  x <- inv zero ;;
+  xor2 x x
+
+Both expressions have the same meaning, and if we were using Gallina ``let``
+binders there would be no difference. But the generated circuit will use two
+wires for the first definition, and fork the same wire in the second. As circuit
+diagrams, this is the difference between::
+
+         +-----+      +-----+
+  1 ---- | inv |------|     |
+         +-----+      | xor | ---- out
+         +-----+      |     |
+  1 ---- | inv | ---- |     |
+         +-----+      +-----+
+
+and::
+
+                            +-----+
+                     +----- |     |
+                     |      | xor | ---- out
+         +-----+     |      |     |
+  1 ---- | inv | ----+----- |     |
+         +-----+            +-----+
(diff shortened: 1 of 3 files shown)


demos/tutorial.v
```diff
@@ -22,6 +22,228 @@
 ========
 Tutorial
 ========
+
+Welcome! This is a quick primer for designing circuits with Cava. We'll walk
+through a few small examples end-to-end. This tutorial assumes some familiarity
+with Coq syntax. Use Ctrl+down and Ctrl+up to step through the Coq code along with the tutorial.
```
satnam6502, line 28: I wonder if this will work on a Mac? I'll try!

demos/tutorial.v
```diff
@@ -22,6 +22,228 @@
...
+Definition inverter
+           {signal : SignalType -> Type}
+           {semantics : Cava signal}
+  : Circuit (signal Bit) (signal Bit) :=
+  Comb inv.
+
+(*|
+A few things to notice here:
+
+* ``SignalType`` is Cava's type system. The inverter is parameterized over
```
satnam6502, line 63: I wonder if it makes sense to at first elide the type of `signal` and then explain it later? So for the moment just say:
```coq
Definition inverter {signal} {semantics : Cava signal}
  : Circuit (signal Bit) (signal Bit) 
```
Or to start off with these as `Context` definitions with a "trust us for now" note? You do immediately afterwards introduce the `Context` form I suppose...
Also very happy with what's written here.

demos/tutorial.v
```diff
@@ -22,6 +22,228 @@
...
+  ``signal``, which converts ``SignalType``\ s to Coq types. ``Bit`` is one
+  example of a ``SignalType``; we'll see more examples later on.
+* ``Comb`` is short for "combinational"; our inverter has no loops, registers,
+  or timing requirements, so it is a purely combinational circuit.
+* The inverter is also paramterized over ``semantics``, an instance of the
+  typeclass ``Cava``. This instance provides implementations of circuit
+  primitives, such as 1-bit logic gates. One primitive is a 1-bit inverter
+  ``inv``, so our inverter is just a simple invocation of the primitive.
```
satnam6502, line 71: "primitive gate"?

demos/tutorial.v
```diff
@@ -22,6 +22,228 @@
...
+Back to our inverter. Let's take a closer look at the ``inv`` primitive.
+|*)
+
+  About inv.
+
+(*|
+You can see in the type signature ``signal Bit -> cava (signal Bit)`` that
+``inv`` is defined as a pure Coq function in terms of a monad called
+``cava``. The ``cava`` monad, like ``inv``, is provided by ``semantics``. The
+monad is used to preserve sharing; it's semantically different in Cava to write::
```
satnam6502, line 98: capture sharing

demos/tutorial.v
```diff
@@ -22,6 +22,228 @@
...
+  y <- inv zero ;;
+  xor2 x y
+
+than it is to write::
+
+  x <- inv zero ;;
+  xor2 x x
+
+Both expressions have the same meaning, and if we were using Gallina ``let``
+binders there would be no difference. But the generated circuit will use two
```
satnam6502, line 110: "might use"
The two representations are semantically equivalent and interchangeable, which is the root of the problem.

demos/tutorial.v
```diff
@@ -22,6 +22,228 @@
...
+                    +---|     |
+                    |   | xor |---- out
+         +-----+    |   |     |
+  0 -----| inv |----+---|     |
+         +-----+        +-----+
+
+This difference isn't significant in determining what the value of ``out`` will
+be, but it can be very useful when trying to exercise fine-grained control over
+circuit layout and area! At a first approximation, you can think of a monadic
+bind (``_ <- _ ;; ...``) as *naming a wire* in the circuit graph.
```
satnam6502, line 133: Perhaps some kind of note saying:

We could have represented sharing by describing circuit graphs with a list of nodes and edges. However, this is essential the "machine code" of structural hardware descriptions which is far too tedious a representation for humans to work with. Cava provides abstractions that allows human engineers to think at a more convenient level of abstraction and be more productive about how circuits are described and composed.

demos/tutorial.v
```diff
@@ -22,6 +22,228 @@
...
+
+Parameterizing over the ``cava`` monad and primitive implementations allows us
+to use different instances of ``Cava`` to interpret the same circuit definition
+in different ways. One ``Cava`` instance generates netlists by adding and
+connecting wires in the background using a state monad. For circuit simulations
+and proofs of functional correctness, on the other hand, we don't care about
+sharing at all; these use no-op identity monad that acts the same as a ``let``
+binder.
```
satnam6502, line 141: Perhaps an analogy with a class in Java/C++ and overloading? Perhaps that's too much of a distraction here.

demos/tutorial.v
```diff
@@ -22,6 +22,228 @@
...
+circuit's input and output ports and behavior relative to the (global) clock and
+reset signals. Then we can compute a netlist (type ``CavaState``), which
+describes the full layout of the circuit in a way that can be easily translated
+to SystemVerilog.
+|*)
+
+(* netlist-generating semantics *)
+Existing Instance CavaCombinationalNet.
+
+Definition inverter_interface
```
satnam6502, line 162: A note to say that although this is a combinational circuit and does not need a clock and reset input, we want to _drive_ this circuit in a context where we provide test stimuli in a sequence that is synchronized with a clock. For that reason we instantiate it as a sequential circuit so we can "step through it".

demos/tutorial.v
```diff
@@ -22,6 +22,228 @@
...
+``signal`` for this ``Cava`` instance is ``combType``, which interprets a
+``Bit`` simply as a Coq ``bool``. If we provide the three inputs
+``[true; false; true]`` to the circuit simulation function ``multistep``, we'll
+get ``[false; true; false]``:
+|*)
+
+(* identity-monad semantics *)
+Existing Instance CombinationalSemantics.
+
+Compute (multistep inverter [true; false; true]).
```
satnam6502, line 186: I assume this will change to `simulate` at some point.

satnam6502 (approved):
Great!

demos/tutorial.v
jadephilipoom: Yes, I'm going to do a little more cleanup and then merge main into the gh-pages branch to pick up those updates.

demos/tutorial.v
satnam6502: Indeed, Ctrl+down and Ctrl+up don't work with the default key bindings on a Mac keyboard.

jadephilipoom: Do Cmd+up amd Cmd+down work?

satnam6502: No.

jadephilipoom: I shall do some investigation.

satnam6502:
Well until then I can just rebind the keys on my Mac to make the UI use a
modifier different from ctrl.


Book a Guru coaching session via go/satnam-guru
[image: Google Logo]
Satnam Singh
Software Engineer, Google Research
***@***.***

PR #627, opened 2021-03-05
Remove unIdent

### LINKED ISSUE #591: Remove the need for unIdent
Possibly controversial, but I think it's a bit awkward to have to write `unIdent` everywhere. The reason is that `coq-ext-lib` defines the identity monad with:
```coq
Inductive ident A := mkIdent { unIdent : A }.

Global Instance Monad_ident : Monad ident :=
{ ret  := fun _ v => mkIdent v
; bind := fun _ _ c f => f (unIdent c)
}.
```
However, there's no need to define it with a record type this way. I would have written:
```coq
Definition ident A := A.

Global Instance Monad_ident : Monad ident :=
{ ret  := fun _ v => v
; bind := fun _ _ c f => f c
}.
```
This would simplify how we talk about the correctness of combinational circuits; `unIdent (foo x y) = foo_spec` would become just `foo x y = foo_spec`.

We also wouldn't lose any helpful lemmas from `coq-ext-lib`, because there just aren't any -- the only file that seems to contain identity monad lemmas is completely commented out. I think it's just in general a slightly neglected piece of code, and we wouldn't lose anything by defining a nicer one ourselves.

jadephilipoom:
Resolves #591 

Cleans up our lemma statements and single-step simulations of combinational circuits by using our own identity monad that does not have a wrapper record type.

### DIFF
--- a/acorn-examples/AdderTree.v
+++ b/acorn-examples/AdderTree.v
@@ -79,15 +79,15 @@ Local Open Scope string_scope.
 Local Open Scope vector_scope.
 
 Definition v0_v1 := [v0; v1].
-Definition v0_plus_v1 : Bvector 8 := unIdent (adderTree 0 v0_v1).
+Definition v0_plus_v1 : Bvector 8 := adderTree 0 v0_v1.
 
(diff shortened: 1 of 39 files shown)


blaxill (approved):
Lgtm, i really like this change!

PR #637, opened 2021-03-11
BitonicSorter in acorn-examples

smore-lore:
Definition + tests (no proof yet)


acorn-examples/BitonicSorter.v
```diff
@@ -0,0 +1,209 @@
...
+(*                                                                          *)
+(* Unless required by applicable law or agreed to in writing, software      *)
+(* distributed under the License is distributed on an "AS IS" BASIS,        *)
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+Require Import Coq.Init.Nat.
+Require Import Coq.micromega.Lia.
```
jadephilipoom, line 18: `Cava.Cava` now exports these; you don't need to import them separately.

acorn-examples/BitonicSorter.v
```diff
@@ -0,0 +1,209 @@
...
+
+  Lemma pow_lt n : 0 < 2 ^ n.
+  Proof.
+    induction n; [auto|].
+    simpl.
+    unfold "<".
+    lia.
+  Qed.
+
+  Definition vec_fst_halve {A n} (input : Vector.t A (pow 2 (S n)))
```
jadephilipoom, line 34: Is this doing the same thing as `divide`?

https://github.com/project-oak/silveroak/blob/d57a195e6acc37555904344d1d31d99f75d16a87/cava/Cava/Lib/Combinators.v#L340-L342

If so, it's probably better to use `divide` since we've already got some proofs about it. (And `divide` should be moved to `Util/Vector.v`, come to think of it.)

acorn-examples/BitonicSorter.v
```diff
@@ -0,0 +1,209 @@
...
+
+    input_snd <- packV (vec_snd_halve input) ;;
+    output_snd <- g (input_snd) ;;
+    output_snd_v <- unpackV output_snd ;;
+
+    packV (vec_pow_app output_fst_v output_snd_v).
+
+  Definition vec_length {A n} (a : Vector.t A n) : nat := n.
+  Fixpoint vec_alternate {A n} (a b : Vector.t A n) : Vector.t A (n + n).
+  Proof.
```
jadephilipoom, line 83: As we talked about over email, it would be nice to avoid the proof terms in definitions using `resize_default` and special matching.

jadephilipoom (commented):
I agree this is likely to be easier to define/prove with lists -- and if the lists are derived from vector signals, you should even get lengths guaranteed automatically anyway.

acorn-examples/BitonicSorter.v
```diff
@@ -0,0 +1,209 @@
...
+    cava (signal (Vec (Vec Bit bw) (pow 2 (S n)))).
+  Proof.
+    (* match n with *)
+    induction n as [|n' butterfly_rec].
+    (* | O => f input *)
+    { replace (2 ^ 1) with (2) in * by reflexivity.
+      exact (f (input)). }
+    (* | S n' => (interleave butterfly_rec >=> evens f) input *)
+    { exact ((interleave butterfly_rec >=> evens f) input). }
+  Qed.
```
satnam6502, line 191: It is possible for the butterfly descriptions to look pretty much identical to the Lava ones? i.e. a simple structural recursive description in terms of smaller butterflies?

satnam6502: Perhaps you or I or we should first have a go at implementing this circuit with lists as we do in Lava in Haskell?

satnam6502:
Also, since you know Agda, I recommend watching this video on the convoy pattern by Matthew Brecknell @mbrcknl
 https://matthew.brecknell.net/posts/pattern-matching-dependent-types-in-coq/ 
He also has come code at https://github.com/mbrcknl/ylj15-coq-pattern-match

acorn-examples/BitonicSorter.v
smore-lore: Removing these didn't seem to work for me.

acorn-examples/BitonicSorter.v
smore-lore: I update this patch to use a different version of pow to avoid some of the type transformations needed, but using splitat now (much nicer!)

acorn-examples/BitonicSorter.v
smore-lore: Done, thanks!

acorn-examples/BitonicSorter.v
smore-lore: This definition is just deceiving - it's using butterfly_rec as the recursive function. So, it is the same recursive definition, just very messy. This is fixed now.

acorn-examples/BitonicSorter.v
jadephilipoom: Sorry, my mistake -- you need `Cava.CavaProperties`, which contains tools for proofs such as `Lia`.

examples/BitonicSorter.v
```diff
@@ -0,0 +1,163 @@
...
+    (default : signal A)
+    (f : signal (Vec A 2) -> cava (signal (Vec A 2)))
+    : signal (Vec A (two_pow (S n))) -> cava (signal (Vec A (two_pow (S n)))) :=
+    match n with
+    | O => f
+    | S n' => interleave default (butterfly default f) >=> evens default f
+    end.
+
+  Definition signal_reverse {A n} : (signal (Vec A n)) -> cava (signal (Vec A n)) :=
+    unpackV >=> reverse >=> packV.
```
satnam6502, line 133: Perhaps worth adding `reverseV` to `Combinators.v` or `Vec.v` at some point.

satnam6502 (approved):
Great! I will add Verilog circuit generation and test-bench generation as a follow up PR.

examples/BitonicSorter.v
```diff
@@ -0,0 +1,163 @@
...
+
+  Lemma opaque_signal_vec {bw} : signal (Vec Bit bw). Admitted.
```
jadephilipoom, line 135: Hmm, this could cause problems for us down the line if we `Print Assumptions` or run an axiom-checker, because it will show up as an unproven axiom. Might be better to just use `defaultSignal` here. In a generic context, you have no information about what `defaultSignal` is, so it provides the same guarantee that the definition can't rely on its contents.

jadephilipoom (approved):
Nice work, LGTM!

examples/BitonicSorter.v
jadephilipoom: Already in `Vec.v`: https://github.com/project-oak/silveroak/blob/6abca5a4f296959c6f5d646463c29488ff1191d6/cava/Cava/Lib/Vec.v#L102-L105

acorn-examples/BitonicSorter.v
smore-lore: This still doesn't work as a drop in place solution, but I'd take IWYU arguments here.

examples/BitonicSorter.v
smore-lore: done

examples/BitonicSorter.v
smore-lore: done

Issue #639, opened 2021-03-11
Polish contents of Lib


jadephilipoom:
Now that we're exporting everything under `Lib` -- our combinators, adders, vector API, etc. -- in `Cava.Cava`, we should make sure these files are concise and easy to read. This issue tracks tasks related to making `Lib` into a nice, presentable core library. In approximate order of priority:

- [x] Remove lingering experimental code and any unused definitions (unused lemmas can generally stay)
- [x] Remove duplicated combinators (for instance, colL should be removed and colV renamed to col)
- [x] Make trees produce efficient results for input sizes that are not powers of 2
- [x] Add missing proofs for any definitions under `Lib` -- every definition we expose should have at least a couple of proofs describing its behavior for the most common use cases

Along the way, add explanatory comments, make the coding style more uniform, and organize files logically. I think all this would go a long way towards making our library definitions more discoverable and easier to use.

jadephilipoom:
With #645, I believe this one is done.

PR #641, opened 2021-03-15
Cleanup of Combinators.v and Adders.v

jadephilipoom:
Part of #639 

A set of interconnected changes that, together, hopefully make our combinators and adders more presentable, browsable, and usable.

- Remove all experimental code from `Combinators.v` and `Adders.v` -- we now have `tree` and `col` instead of six different trees and three different `col` combinators, and our adders are just `addN` (no carries) and `addC` (carries both in and out)
- Make `tree` combinator create binary trees of optimal depth for all input sizes, not just powers of 2
- Organize `Combinators.v` into logical sections
- Make xilinx adder take its carry on the right of the tuple instead of the left so that the tuple is flat (removes need for `xilinxAdderCarryFlat` in the tests to work with the netlist interface)
- Move proofs from `Combinators.v` to `CombinatorProperties.v`
- Move `eqb` and `all`, which are not really combinators, from `Combinators.v` to `CavaPrelude.v`
- Remove `zipWith` (also not a combinator, and duplicate of `Vec.map2`)
- Create new file `CavaPreludeProperties.v` to house `eqb` and `all` proofs, remove them from `CombinatorProperties.v`
- Change some adder argument names from `a`, `b`, and `cin` to `x`, `y` and `cin`, to make the distinction between the operands and the carry bit clearer
- Add correctness lemmas for the `col` combinator, use them to prove adders correct

### DIFF
--- a/cava/Cava/Lib/Adders.v
+++ b/cava/Cava/Lib/Adders.v
@@ -29,117 +29,58 @@ Section WithCava.
   Context `{semantics:Cava}.
 
   (* Build a half adder *)
-  Definition halfAdder '(a, b) :=
-    partial_sum <- xor2 (a, b) ;;
-    carry <- and2 (a, b) ;;
+  Definition halfAdder '(x, y) :=
+    partial_sum <- xor2 (x, y) ;;
+    carry <- and2 (x, y) ;;
     ret (partial_sum, carry).
 
   (* A full adder *)
-  Definition fullAdder '(cin, (a, b))
+  Definition fullAdder '(cin, (x, y))
                        : cava (signal Bit * signal Bit) :=
-    '(abl, abh) <- halfAdder (a, b) ;;
-    '(abcl, abch) <- halfAdder (abl, cin) ;;
-    cout <- or2 (abh, abch) ;;
-    ret (abcl, cout).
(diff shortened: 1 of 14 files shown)


cava/Cava/Lib/Adders.v
```diff
@@ -29,117 +29,58 @@ Section WithCava.
...
-            (inputs: signal Bit * (Vector.t (signal Bit * signal Bit)) n) :
-            cava (Vector.t (signal Bit) n * signal Bit) :=
-    colV fullAdder inputs.
-
-  (* Adder with a pair of inputs of the same size and no bit-growth. *)
+  Definition addC {n : nat}
+             (inputs : signal (Vec Bit n) * signal (Vec Bit n) * signal Bit) :
+    cava (signal (Vec Bit n) * signal Bit) :=
+    let '(x, y, cin) := inputs in
+    x <- unpackV x ;;
```
satnam6502, line 50: I'd vote for not reusing names?

satnam6502 (approved):

cava/Cava/Lib/Adders.v
satnam6502: Ignore my comment, I think in such a short definition this is totally fine!

PR #643, opened 2021-03-15
Merge main into gh-pages

jadephilipoom:
Redoing #632 (and adding the changes to `main` since), having learned the lesson to rebase instead of squash!


blaxill (commented):
I think it would be best to keep `gh-pages` clear of code and add `main` as a submodule if needed. If we go down that route, do you want to merge this in the meantime?

jadephilipoom:
Looks like we'll definitely have to go the other route. I can't merge this PR without squashing (rebasing has conflicts and merge commits are disallowed), which would just put us in the same position for the next merge. Created #644 to track this task.

I'm going to see if I can create the new repo quickly; if I can, I'll just close this. If it seems like it might take a while, I'll squash + merge this PR and push the updates to the tutorial that I have locally, so that they can be up while we wait on further setup. Does that sound reasonable?

blaxill:
Yep sounds good to me!

satnam6502 (approved):

jadephilipoom:
Closing in favor of #649 (see discussion on #644)

PR #645, opened 2021-03-16
Add missing combinator proofs

jadephilipoom:
For #639 

- Adds proofs for combinators that previously had no proofs
- Adds brief explanatory comments above each combinator (if one did not previously exist)
- Adds `dropl` and `dropr` combinators (I remember these from arrows and they were nice to have)
- Generalizes a few combinators that previously required `signal A` to work on any type `A`


blaxill (approved):
Lgtm!

PR #651, opened 2021-03-16
Tutorial: three xor-based examples

jadephilipoom:
For #609

This updates the tutorial with three progressively more complex (but all combinational) circuits:
- xor two bytes
- xor two n-bit bitvectors
- xor an m-length vector of n-bit bitvectors using the `tree` combinator

The idea here is to keep the example circuits conceptually simple, but introduce working with Cava datatypes and combinators, as well as demonstrate how inductive logic creates proofs that apply to huge circuits without requiring a lot of computation time.


demos/tutorial.v
```diff
@@ -244,10 +252,352 @@ To summarize, there are three things you can do with Cava circuits:
...
+Example 2 : Byte xor
+====================
+
+Our next example is a circuit that xors two bytes:
+|*)
+
+  Definition xor_byte :
+    Circuit (signal (Vec Bit 8) * signal (Vec Bit 8))
+            (signal (Vec Bit 8)) :=
+    Comb (fun '(v1, v2) => Vec.map2 xor2 v1 v2).
```
satnam6502, line 274: I wonder if we should have `xor2` take a pair? But currying makes it easier to specialize / partially apply?

demos/tutorial.v
```diff
@@ -244,10 +252,352 @@ To summarize, there are three things you can do with Cava circuits:
...
+the inverter, except that we change the input and output port types to match the
+circuit's type signature.
+|*)
+
+Definition xor_byte_interface
+  := sequentialInterface "xor_byte_interface"
+     "clk" PositiveEdge "rst" PositiveEdge
+     [mkPort "v1" (Vec Bit 8); mkPort "v2" (Vec Bit 8)]
+     [mkPort "o" (Vec Bit 8)]
+     [].
```
satnam6502, line 305: Perhaps I should remove this and make it available only in a more generate interface builder.

demos/tutorial.v
```diff
@@ -244,10 +252,352 @@ To summarize, there are three things you can do with Cava circuits:
...
+
+Compute (makeCircuitNetlist xor_byte_interface xor_byte).(module).
+
+(*|
+Tuples in the input or output types become lists of ports for the netlist
+interface, so ``signal (Vec Bit 8) * signal (Vec Bit 8)`` becomes ``[mkPort "v1"
+(Vec Bit 8); mkPort "v2" (Vec Bit 8)]``. The names of the ports ("v1", "v2", and
+"o") are just for readability and potentially reference by other netlists; they
```
satnam6502, line 313: potentially _for_ reference
_to_ _be_ referenced

demos/tutorial.v
```diff
@@ -244,10 +252,352 @@ To summarize, there are three things you can do with Cava circuits:
...
+can be named however you prefer.
+
+We can also, as before, simulate the circuit.
+|*)
+
+Compute
+  simulate xor_byte
```
satnam6502, line 320: Using scoping to remove the need for `%vector`?

satnam6502 (approved):
Cool!

demos/tutorial.v
jadephilipoom: `xor2` does take a pair, it's `Vec.map2` that doesn't -- but I actually have one last `Lib`-cleanup change queued that will change that!

demos/tutorial.v
blaxill: Curious to hear more about what you are thinking here Satnam

blaxill (approved):

satnam6502:
Well I added this as a hook for incorporating information about how a debug ILA core should be hooked up, I/O pads etc. It might be better to rename this general interface generation function and then make a specialized version that has this list set to empty for the common case.

PR #658, opened 2021-03-17
Cleanup of Vec, CavaPrelude, and BitVectorOps

jadephilipoom:
One last `Lib` cleanup #639 

Some potentially controversial changes here. Let me know your thoughts!
- Remove `BitVectorOps` (previously contained only bit-vector `xor`) and put those operations in their own section of `Vec.v`. My reasoning here is that, while bit vectors are sort of different from general vectors, bit vectors still conceptually fit in the vector part of the library, and I think it's clearer to write `Vec.xor` instead of `xorV`.
- Add optional notations for the bit vector operations `and`, `or`, and `xor`. Previously, we had `xorV`, which was uncurried (takes a tuple), and `xorv`, which was curried (takes two separate arguments). It's not ideal to have two versions, and we've generally standardized on having uncurried definitions. However, the uncurried version is awkward to chain, which is probably why we had the curried version. This line of four `xor`s, for instance, would need to become four lines:
https://github.com/project-oak/silveroak/blob/b4be19338faf54e4689c15a3e7bfac5960def10d/silveroak-opentitan/aes/Impl/MixColumnsCircuit.v#L100 This PR defines a notation that changes it to `data_o0 <- (data_i_1 ⊕ x_mul2_3 ⊕ x_1 ⊕ z_muxed_1) ;;`. I was a bit conflicted about whether or not to create it, because too many notations can make code inscrutable, but `&`, `|`, and `⊕` are fairly standard notations for these operations (unfortunately we can't use `^` because it conflicts with the power notation) and I think this is an overall win for readability in the case where several are chained together. The bit vector notations are not imported by default, and are currently only used in `MixColumnsCircuit`.

Other changes in this PR:
- Uncurry `eqb`, `Vec.fold_left2`, and `Vec.map2`
- Add some more definitions/proofs to fill gaps in our API
  - `any` as a natural counterpart to `all`
  - all boolean operations from `Cava` (e.g. `and`, `nor`) as bit-vector operations
  - `Vec.map_literal` to create a `Vec` from a map over a `Vector.t` that isn't necessarily a `SignalType`
  - versions of `packV` and `unpackV` that unpack multiple layers of vectors, up to four dimensions
  - Versions of `flatten`, `reshape`, and `resize_default` that work on `Vec`
  - Proofs stating `zero = false` and `one = true` in the identity-monad world so that `simpl_ident` does these conversions automatically


cava/Cava/Lib/Vec.v
```diff
@@ -93,31 +163,90 @@ Section WithCava.
...
+             .. (fun a => bind (and (a,z)) ret) .. ) x)
+      (at level 40, left associativity, z at next level, only parsing) : bitvec_scope.
+
+  Notation "x | y | .. | z" :=
+    ((fun a =>
+        bind (or (a,y))
+             .. (fun a => bind (or (a,z)) ret) .. ) x)
+      (at level 40, left associativity, z at next level, only parsing) : bitvec_scope.
+
+  Notation "x ⊕ y ⊕ .. ⊕ z" :=
```
blaxill, line 247: Is there a non-unicode symbol we can use? You mention `^` overlaps with power, can we not distinguish via scope or something?

jadephilipoom: Did some more digging and figured out why I was getting the conflict and a workaround. I was getting no results from `Locate "^"`, indicating no notation with `^` in scope, but errors (`Unknown interpretation for _ ^ _`) whenever I tried to use the notation. This is because in `Coq.Init.Notations`, part of Coq's core library, the notation is reserved:
https://github.com/coq/coq/blob/eeef63b0b09cf90f7a3022ce6f0d7e50a908484c/theories/Init/Notations.v#L51

The reservation, without providing a meaning for the notation, forces it to be at a certain notation level (30) and with a certain associativity, so when I tried to put it at level 40, my best guess is that the reserved version took precedence (and has no interpretation, hence the error that looked like the notation was not defined at all).

The fix is that I can simply define this notation at level 20, making them higher-priority than the reserved version. Seems to work locally -- I'll push an update momentarily.

blaxill (approved):
Lgtm

satnam6502:
LGTM

Issue #657, opened 2021-03-17
Implement an efficient constant equality for vectors


satnam6502:
Define a combinational circuit parameterized on the size of a vector `n` which takes as input an n-bit vector and returns `true` if the binary value of the vector represents a given compile time `nat` constant `v`, `false` otherwise. This circuit should have an efficient implementation which we hope to use as a component for a decoder circuit as part of another issue.

```coq
Section WithCava.
  Context {signal} `{Cava signal}.

  Definition vecConstEq (n v: nat)
                        (a: signal (Vec Bit n)) :
                        cava (signal Bit).
     (* Define! *)

End WithCava.
```

You should add your example the Cava's `Lib` directory in `VecConsteq.v`.

**Hint**. Convert the nat `v` into a bit-vector, then perform pair-wise comparison, and then use an AND-tree.
We will want help from @blaxill @jadephilipoom @smore-lore and @cipher1024 

Steps, submitting a PR for review after each step:

- [ ] First define any Gallina function that gives the correct result. Test your function with a few `Example` or `Goal` clauses.
- [ ] Iterate description until you have something that works by comparing bits pairwise and then AND-ing them together (folding the ANDs or using a tree). Check with some tests.
- [ ] Define a circuit interface and then extract both the circuit and a testbench. Add the invocation of the Verilator test-bench for your circuit and make sure the test passes.

satnam6502:
@blaxill @jadephilipoom Does this circuit already exist in some form as part of the AES core?

jadephilipoom:
> @blaxill @jadephilipoom Does this circuit already exist in some form as part of the AES core?

We already have `eqb`, which works for all signals and does indeed compare with an and-tree (`all`):
https://github.com/project-oak/silveroak/blob/b4be19338faf54e4689c15a3e7bfac5960def10d/cava/Cava/Lib/CavaPrelude.v#L45-L53

If I specialize to a `Vec Bit n` and simplify `eqb` and `all` (`Eval cbv [eqb all] in (fun n => eqb (t:=Vec Bit n))`), I get:
```coq
fun (n : nat) (x y : combType (Vec Bit n)) =>
       eq_results <- Vec.map2 (fun '(a, b) => xnor2 (a, b)) (x, y);; match n with
                                                                     | 0 => ret one
                                                                     | S _ => tree and2 eq_results
                                                                     end
```
Is this the circuit needed for the task?

satnam6502:
[...]

PR #662, opened 2021-03-17
Add Cava.Lib.VecConstEq

atondwal:
Implement an efficient constant equality for vectors #657

Currently we compare against `nat`s, so the constants can be out of range (in
which case they're truncated). Eventually these should be replaced by
`Fin 2^n`s, which should prevent this.

### DIFF
--- a/cava/Cava/Lib/VecConstEq.v
+++ b/cava/Cava/Lib/VecConstEq.v
@@ -14,50 +14,34 @@
 (* limitations under the License.                                           *)
 (****************************************************************************)
 
-Require Import Coq.Vectors.Vector.
-Require Import Coq.ZArith.ZArith.
-Require Import ExtLib.Structures.Monads.
-Import VectorNotations.
-Import MonadNotation.
-Local Open Scope vector_scope.
-Local Open Scope monad_scope.
-Local Open Scope N_scope.
-
 Require Import Cava.Cava.
-Require Import Cava.Lib.Vec.
-
 
 Section WithCava.
   Context {signal} `{Cava signal}.
 
   (* Check if [inp] is bit-equal to the first [n] bits of [v] *)
   (* TODO(atondwal): use [Fin] to disallow numbers larger than [2^n]. c.f. [ex5] *)
-  Definition vecConstEq {n v: nat}
+  Definition vecConstEq (n v: nat)
                         (inp: signal (Vec Bit n)) :
                         cava (signal Bit) :=
-    const <- bitvec_literal (N2Bv_sized n (N.of_nat v)) ;;
-    eq_results <- Vec.map2 (fun '(a, b) => xnor2 (a, b)) (inp, const) ;;
-    match n with
-    | 0%nat => ret one
-    | S _ => tree and2 eq_results
-    end.
-
-  Definition vecConstEq_combinator (n v: nat):= Comb (@vecConstEq n v).
+    const <- Vec.bitvec_literal (N2Bv_sized n (N.of_nat v)) ;;
+    eqb (inp, const).
 
 End WithCava.
 
 
-Example ex1 : simulate (vecConstEq_combinator 8 42) [N2Bv_sized 8 42] = [true].
+Example ex1 : vecConstEq 8 42 =<< Vec.bitvec_literal (N2Bv_sized 8 42) = ret true.
 Proof. trivial. Qed.
 
-Example ex2 : simulate (vecConstEq_combinator 8 43) [N2Bv_sized 8 42] = [false].
+Example ex2 : vecConstEq 8 43 =<< Vec.bitvec_literal (N2Bv_sized 8 42) = ret false.
 Proof. trivial. Qed.
 
-Example ex3 : simulate (vecConstEq_combinator 1 1) [N2Bv_sized 1 1] = [true].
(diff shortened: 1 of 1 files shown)


satnam6502:
Great! Can I change my mind and not have `n` and `v` implicit:
```coq
  Definition vecConstEq (n v : nat)
                        (inp: signal (Vec Bit n)) :
                        cava (signal Bit
```
For the tests you should be able to wrote:
```coq
Example ex1 : vecConstEq 8 42 (N2Bv_sized 8 42) = true.
Proof. trivial. Qed.
```

satnam6502:
Also if you moved `Local Open Scope N_scope.` to just before your tests, you don't neet to qualify `0%nat`.

cava/Cava/Lib/VecConstEq.v
```diff
@@ -0,0 +1,63 @@
...
+(*                                                                          *)
+(* Unless required by applicable law or agreed to in writing, software      *)
+(* distributed under the License is distributed on an "AS IS" BASIS,        *)
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+Require Import Coq.Vectors.Vector.
+Require Import Coq.ZArith.ZArith.
```
jadephilipoom, line 18: Why `ZArith`? Possibly unused import?

cava/Cava/Lib/VecConstEq.v
```diff
@@ -0,0 +1,63 @@
...
+Require Import ExtLib.Structures.Monads.
+Import VectorNotations.
+Import MonadNotation.
```
jadephilipoom, line 21: These should all be imported by `Cava.Cava` and shouldn't need to be imported separately -- let me know if there's a problem and I'll investigate.

cava/Cava/Lib/VecConstEq.v
```diff
@@ -0,0 +1,63 @@
...
+Local Open Scope vector_scope.
+Local Open Scope monad_scope.
+Local Open Scope N_scope.
+
+Require Import Cava.Cava.
+Require Import Cava.Lib.Vec.
```
jadephilipoom, line 27: `Vec` is `Require`d in `Cava.Cava` but not `Import`ed because it has lots of name collisions with Coq's `Vector`. Best practice here is to not import it and just write `Vec.map` where you need.

jadephilipoom (approved):
Thanks! LGTM as-is, but I've got some nitpicks and opinions below.

cava/Cava/Lib/VecConstEq.v
```diff
@@ -0,0 +1,63 @@
...
+
+
+Section WithCava.
+  Context {signal} `{Cava signal}.
+
+  (* Check if [inp] is bit-equal to the first [n] bits of [v] *)
+  (* TODO(atondwal): use [Fin] to disallow numbers larger than [2^n]. c.f. [ex5] *)
```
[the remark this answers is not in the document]
blaxill: > I strongly recommend avoiding Fin -- it includes hidden proof terms

I'm not suggesting we use Fin.t, but I don't think this is quite true, Fin.t is 

~~~Coq
Inductive t : nat -> Set :=
|F1 : forall {n}, t (S n)
|FS : forall {n}, t n -> t (S n).
~~~

albeit a lot of its functions use proofs internally (e.g. of_nat to_nat)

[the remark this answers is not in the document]
jadephilipoom: True -- I misspoke saying that it contains proof terms itself. However, using it effectively usually involves proof terms and most of the standard library functions that use it have proof terms.

atondwal:
For the tests you should be able to write:
```
Example ex1 : vecConstEq 8 42 (N2Bv_sized 8 42) = true.
Proof. trivial. Qed.
```

Hmm? `N2Bv_sized 8 42` is a Bvector 8 and I need a `signal (Vec Bit 8)`. I've rewritten the examples to be a bit more like that, though.

cava/Cava/Lib/VecConstEq.v
atondwal: Oh yep. I'd accidentally made Cava.Cava a circular dependency, imported all the stuff inside it I needed, and then fixed the circular dependency problem, but I forgot to human `build_cleaner` it.

satnam6502:
> For the tests you should be able to write:
> 
> ```
> Example ex1 : vecConstEq 8 42 (N2Bv_sized 8 42) = true.
> Proof. trivial. Qed.
> ```
> 
> Hmm? `N2Bv_sized 8 42` is a Bvector 8 and I need a `signal (Vec Bit 8)`. I've rewritten the examples to be a bit more like that, though.

The expression `signal (Vec Bit 8)` **is** `Bvector 8` i.e. `Vector.t bool 8` for the simulation/semantic interpretation.

atondwal:
Ah, so I can do that so long as they're outside the `WithCava` section, I see!

satnam6502 (approved):
Great work! Thank you!

atondwal:
I don't have permission to merge into `main`!

satnam6502:
> I don't have permission to merge into `main`!

Please try now.

blaxill (approved):
Lgtm too, thanks

PR #670, opened 2021-03-18
Update README.md

atondwal:
spent some time looking from a `.vs` file thanks to this typo :)


satnam6502 (approved):
Good catch! Thank you so much for being our canary.

PR #678, opened 2021-03-22
Add and prove an incrementer

### LINKED ISSUE #474: Replace IncrDecr with an implementation using synthesizable adders and subractors

jadephilipoom:
Resolves #474 

This incrementer (unlike the version in `examples/IncrDecr.v`) uses the synthesizable adder. I wanted an incrementer for a tutorial example, and it seemed like a good thing to add to our library anyway. Following the same pattern as our bit-vector adders `addN` and `addC`, there's one incrementer that has a carry bit to indicate overflow (`incrC`), and one without a carry (`incrN`).

### DIFF
--- a/cava/Cava/Lib/Adders.v
+++ b/cava/Cava/Lib/Adders.v
@@ -61,6 +61,29 @@ Section WithCava.
     '(sum, _) <- addC (xy, zero) ;;
     ret sum.
 
+  (* Increment an n-bit vector, representing result as an n-bit vector and a
+     carry bit *)
+  Definition incrC {n : nat} : signal (Vec Bit n) -> cava (signal (Vec Bit n) * signal Bit) :=
+    match n with
+    | 0 => fun input => ret (input, one) (* incrementing a 0-length vector always overflows *)
+    | S m =>
+      fun input : signal (Vec Bit (S m)) =>
+        (* use synthesizable adder to add 1 *)
+        onev <- packV [one] ;;
+        sum <- unsignedAdd (input, onev) ;;
+        (* resize (1 + Nat.max (S m) 1) to S (S m) *)
+        sum <- Vec.resize_default (S (S m)) sum ;;
+        (* separate highest bit from rest of sum *)
+        sum_low <- Vec.shiftout sum ;;
+        sum_high <- Vec.last sum ;;
+        ret (sum_low, sum_high)
+    end.
+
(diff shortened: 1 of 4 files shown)


cava/Cava/Lib/Adders.v
```diff
@@ -61,6 +61,29 @@ Section WithCava.
...
+    | S m =>
+      fun input : signal (Vec Bit (S m)) =>
+        (* use synthesizable adder to add 1 *)
+        onev <- packV [one] ;;
+        sum <- unsignedAdd (input, onev) ;;
+        (* resize (1 + Nat.max (S m) 1) to S (S m) *)
+        sum <- Vec.resize_default (S (S m)) sum ;;
+        (* separate highest bit from rest of sum *)
+        sum_low <- Vec.shiftout sum ;;
+        sum_high <- Vec.last sum ;;
```
blaxill, line 78: Perhaps we should have an 'unsnoc' similar to the `Vector.t` version to get both low+high.

blaxill (approved):
Lgtm

cava/Cava/Lib/Adders.v
```diff
@@ -61,6 +61,29 @@ Section WithCava.
...
 
+  (* Increment an n-bit vector, representing result as an n-bit vector and a
+     carry bit *)
+  Definition incrC {n : nat} : signal (Vec Bit n) -> cava (signal (Vec Bit n) * signal Bit) :=
+    match n with
+    | 0 => fun input => ret (input, one) (* incrementing a 0-length vector always overflows *)
+    | S m =>
+      fun input : signal (Vec Bit (S m)) =>
+        (* use synthesizable adder to add 1 *)
+        onev <- packV [one] ;;
```
satnam6502, line 72: I wonder if it is worth having variants of `bitvec_to_nat` and `nat_to_bitvec` which work over the overloaded constants `zero` and `one`? They need not be in a monadic context, it would be find to just use them as expressions. For example:
```coq
  sum <- unsignedAdd (input, nat_to_bitvec 1 1) ;;
```
or even
```coq
  sum <- unsignedAdd (input, nat_to_bitvec n 1) ;;
```

satnam6502 (commented):
Great! Thanks.

PR #681, opened 2021-03-22
Add constantV to CavaClass

jadephilipoom:
While trying to demonstrate loops with initial values, I noticed an inconvenient feature of our current setup. Loops and delays take `signal t` as their reset value, see here:

https://github.com/project-oak/silveroak/blob/e83fe025e5186c59ea6e31256f4c2f8f86cc5505/cava/Cava/Core/Circuit.v#L36-L42

However, in the usual circuit-definition context with a completely abstract `cava` monad, it's impossible to construct a `signal (Vec _ _)` that doesn't have a `cava` wrapped around it (since #597, `packV` wraps vectors in `cava`).

This PR introduces the ability to construct _constant_ vectors with `constantV`, similar to `constant` for bits. It comes at the cost of a new field for `CavaClass`.

@blaxill If you have a moment, it would be nice to confirm that this change doesn't affect the resource usage of our generated AES circuits.

Note: the changes to `CipherControlCircuit.v` are essentially undoing the changes to #597, where wrapping `packV` in `cava` forced a bunch of constants to be wrapped in `cava` too.

### DIFF
--- a/cava/Cava/Core/CavaClass.v
+++ b/cava/Cava/Core/CavaClass.v
@@ -32,6 +32,7 @@ Class Cava (signal : SignalType -> Type) := {
   monad :> Monad cava;
   (* Constant values. *)
   constant : bool -> signal Bit;
+  constantV : forall {A} {n : nat}, Vector.t (signal A) n -> signal (Vec A n);
   (* Default values. *)
   defaultSignal : forall {t: SignalType}, signal t;
   (* SystemVerilog primitive gates *)
@@ -58,6 +59,7 @@ Class Cava (signal : SignalType -> Type) := {
   (* Converting to/from Vector.t *)
   unpackV : forall {t : SignalType} {s : nat}, signal (Vec t s) -> cava (Vector.t (signal t) s);
   packV : forall {t : SignalType} {s : nat} , Vector.t (signal t) s -> cava (signal (Vec t s));
+  (* Dynamic indexing *)
   indexAt : forall {t : SignalType} {sz isz: nat},
             signal (Vec t sz) ->     (* A vector of n elements of type signal t *)
             signal (Vec Bit isz) ->  (* A bit-vector index of size isz bits *)
--- a/cava/Cava/Lib/Vec.v
+++ b/cava/Cava/Lib/Vec.v
@@ -34,9 +34,8 @@ Section WithCava.
   (* Convert back and forth from Cava and Coq vectors                         *)
   (****************************************************************************)
 
-  Definition bitvec_literal {n} (v : Vector.t bool n)
-    : cava (signal (Vec Bit n)) :=
-    packV (Vector.map constant v).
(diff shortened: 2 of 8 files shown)


blaxill:
> @blaxill If you have a moment, it would be nice to confirm that this change doesn't affect the resource usage of our generated AES circuits.

👍  i'll take a look

satnam6502:
I think it is wrong for initial (reset) values to be specified as a `signal t`. It should be `combType t` i.e. a compile time value.
```coq
  | DelayInitCE :
      forall {t} (resetval : combType t),
        Circuit (signal t * signal Bit) (signal t)
```
because the reset value (for our semantics) can never vary.

In the generated SystemVerilog the reset value is not an input pin to the register circuit, it is a compile time parameter attached to the instantiation and is used to produce the configuration bitstream for an FPGA circuit e.g. in VHDL:
```vhdl
FDRE_inst : FDRE
generic map (
INIT => 0, -- Initial value of register, 1’b0, 1’b1
-- Programmable Inversion Attributes: Specifies the use of the built-in programmable inversion
IS_C_INVERTED => ’0’, -- Optional inversion for C
IS_D_INVERTED => ’0’, -- Optional inversion for D
IS_R_INVERTED => ’0’ -- Optional inversion for R
)
port map (
Q => Q, -- 1-bit output: Data
C => C, -- 1-bit input: Clock
CE => CE, -- 1-bit input: Clock enable
D => D, -- 1-bit input: Data
R => R -- 1-bit input: Synchronous reset
);
```

jadephilipoom:
Will there be any problems extracting, for instance, a Coq `Vector.t` as a SystemVerilog constant vector?

satnam6502:
> Will there be any problems extracting, for instance, a Coq `Vector.t` as a SystemVerilog constant vector?

No, it's perfectly fine to have a vector literal in SystemVerilog. We already generate lots of them.

jadephilipoom:
> No, it's perfectly fine to have a vector literal in SystemVerilog. We already generate lots of them.

Okay, great! I can change that, then.

What do folks think about this change, given the original justification is gone? Is it still beneficial to have the ability to write constant vectors without the monadic wrappers?

jadephilipoom:
I think this change might be a prerequisite for https://github.com/project-oak/silveroak/pull/678#discussion_r598997698

satnam6502:
> > No, it's perfectly fine to have a vector literal in SystemVerilog. We already generate lots of them.
> 
> Okay, great! I can change that, then.
> 
> What do folks think about this change, given the original justification is gone? Is it still beneficial to have the ability to write constant vectors without the monadic wrappers?

Yes.

blaxill (approved):
Lgtm! I don't see any negative impact on resource usage (I've checked LUT usage for sbox_lut and mix_columns which I think is a sufficient test)

PR #682, opened 2021-03-23
Add proofs for simulating delays

jadephilipoom:
While writing the tutorial I realized we were completely missing proofs about the simulation of delays. This PR adds them for all four variants.


blaxill (approved):
Lgtm!

PR #693, opened 2021-03-24
Add the first two sequential examples to the tutorial

jadephilipoom:
For #609 

- `three_delays` circuit to introduce sequential circuit concepts
- `sum` circuit to introduce `Loop`, `LoopInit`, and loop invariant proofs

There will be one or two more "interesting" circuits following these two.

Also included here:
- I realized that we didn't have the adder lemmas added as hints to `simpl_ident`, so I added them
- The `demos` Makefile was for some reason writing to `_CoqProject` repeatedly, so there were n copies of the same lines. I didn't get far debugging it, so as a hotfix I just made the `_CoqProject` target remove the file if it already exists.


demos/tutorial.v
```diff
@@ -135,6 +135,9 @@ be, but it can be very useful when trying to exercise fine-grained control over
 circuit layout and area! At a first approximation, you can think of a monadic
 bind (``_ <- _ ;; ...``) as *naming a wire* in the circuit graph.
 
+If the monad notations are unfamiliar, the reference_ has more information on
```
satnam6502, line 138: What does the underscore mean?

demos/tutorial.v
```diff
@@ -614,8 +628,416 @@ At this point, we've covered pretty much everything you need to start building
...
+=====================================
+
+.. coq:: none
+|*)
+
+Section WithCava.
+  Context {signal} {semantics : Cava signal}.
+
+(*|
+The simplest sequential element is a delay (register). The delay takes in a
```
satnam6502, line 641: unit delay

satnam6502 (approved):
Thanks! That's great!

demos/tutorial.v
jadephilipoom: It's a link -- at the bottom of the file, there's `.. _reference : /reference`, which declares that every `reference_` should be replaced by a link to the relative path `/reference`.

PR #695, opened 2021-03-24
Add tutorial stylesheet

### LINKED ISSUE #694: Un-stylize arrows in Alectryon output
The default font settings for Alectryon stylize arrows, which makes it very hard to tell `>=>` and `>==>` apart. It uses the [iosevka](https://typeof.net/Iosevka) font family, which is actually very configurable. I think the way forward here is likely to use Iosevka's custom engine to create our own version of "the Coq style but minus the arrow stylization" and include it as a static resource. Alectryon should then pull it in automatically.

jadephilipoom:
Fixes #694 

There ended up being a much simpler solution to this problem than I originally thought; I can just add a `.. raw:: html` block and a custom stylesheet that overrides the arrow-stylization.

### DIFF
--- a/demos/tutorial.v
+++ b/demos/tutorial.v
@@ -1,4 +1,8 @@
 (*|
+.. raw:: html
+
+   <link rel="stylesheet" href="tutorial.css" type="text/css" />
+
 .. coq:: none
 |*)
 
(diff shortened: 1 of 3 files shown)


satnam6502 (approved):

cpitclaudel:
👍 Alternatively, I think `font-variant-ligatures: none;` might work (and that would guarantee that ligatures don't appear for visitors whose default monospace font has ligatures)

jadephilipoom:
>  Alternatively, I think `font-variant-ligatures: none;` might work (and that would guarantee that ligatures don't appear for visitors whose default monospace font has ligatures)

It does work, even better. Thanks again!

PR #697, opened 2021-03-24
BitonicSorter proofs

smore-lore:
WIP

### DIFF
--- a/examples/BitonicSorter.v
+++ b/examples/BitonicSorter.v
@@ -141,9 +141,8 @@ Section WithCava.
 End WithCava.
 
 Section Test.
-  Definition bw := 8. (* only test numbers < 2**8 *)
-
   Definition run_sort (n : nat) (v : Vector.t N (two_pow n)) : Vector.t N (two_pow n) :=
+    let bw := 8 in
     Vector.map Bv2N (@bitonicSorter combType _ _ n
                        (N2Bv_sized bw 0)
                        (Vector.map (N2Bv_sized bw) v)).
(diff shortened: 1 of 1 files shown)


examples/BitonicSorter.v
```diff
@@ -159,3 +158,78 @@ Section Test.
...
+           parl
+           packV
+           unpackV
+           Combinational.CombinationalSemantics
+           unriffle
+           riffle
+           vec_riffle
+           vec_unriffle].
+      simpl.
+      do 2 rewrite VectorSpec.splitat_append.
```
smore-lore, line 205: I don't yet have the intuition for how to proceed with this proof. I'm planning on building some properties around butterfly networks next, but definitely missing intuition here.

satnam6502: Over my head, I defer to @jadephilipoom

satnam6502:
However, I see you thread through a `default` value. Why not just use `defaultSignal`?  Then you would not need to pass around this default, because we have already defined a default signal for every type of signal expression.

examples/BitonicSorter.v
jadephilipoom: From what I can tell looking at the source, it might be the case that you need some more intermediate lemmas. Unfolding multiple recursive functions at once can quickly give you a goal that's too complicated to easily reason about. It often helps me to write down an argument on paper or in a comment for why I think a function is correct, in terms of the behavior of its subroutines (and why their behavior is correct in terms of their subroutines).

smore-lore: I was hoping to get more intuition from Satnam since he co-wrote the paper I was basing this off of. The paper states that this particular butterfly network will always sort a list whose first half is in order and second half is reversed. However, I didn't understand why (and I haven't read the Springer book yet).

PR #702, opened 2021-03-25
Add fibonacci example to tutorial

### LINKED ISSUE #609: Tutorial for Cava

jadephilipoom:
Resolves #609 

With this addition, our tutorial is first-draft complete! (I'm also going to add the exponentiation-by-squaring circuit that I was thinking about including, but as a separate example rather than included with the tutorial because it's a little too complex for this context.)

Doing this example made me realize that our system is a little annoying for doing proofs of circuits with sequential bodies, because there are a bunch of extra `unit`s in the `circuit_state` type representing the combinational parts. The loop invariant proofs also struggle to unify the type of the circuit state without being provided the body structure explicitly. I'll make a new issue about those wrinkles.

However, I think once this is merged I'll start asking people to beta-test the tutorial and give me feedback on it!


demos/tutorial.v
```diff
@@ -1043,5 +1056,258 @@ Lemma sum_concise_correct n (input : list (combType (Vec Bit n))):
...
+circuit how many timesteps it should run for.
+
+It's also possible to write the ``fibonacci`` circuit as two nested loops with a
+combinational body (essentially a mealy_ machine).
+|*)
+
+  Definition fibonacci_mealy {sz}
+    : Circuit (signal Void) (signal (Vec Bit sz)) :=
+    let v1 : combType (Vec Bit sz) := N2Bv_sized sz 1 in
+    let v_negative1 : combType (Vec Bit sz) := Vector.const one sz in
```
satnam6502, line 1113: At some point in the future we could look into more civilized types for unsigned and signed values.

satnam6502 (approved):

PR #708, opened 2021-03-29
fix #704

### LINKED ISSUE #704: Rework decoder to not use packV and unpackV
The current decoder is defined as:
```coq
  Definition map2VC {n a b c}
    (f: a -> signal b -> cava (signal c))
    (v1: Vector.t a n)
    : signal (Vec b n)
    -> cava (signal (Vec c n))
    :=  unpackV
    >=> Vector.map2 (fun f v => f v) (Vector.map f v1)
    >=> mapT_vector id
    >=> packV.

  (* A decoder from binary to one-hot. Both are big endian *)
  Definition decoder {n : nat}
    : signal (Vec Bit n) -> cava (signal (Vec Bit (2^n)))
    := forkN (2^n) >=> map2VC (vecConstEq n) (Vector.vseq 0 (2^n)).

```
which is mainly my fault. It uses `packV` and `unpackV` but really I feel these two operators should only get used in our `Lib/Vec` library to define our overloaded vector operations. So let's rework this. Here are some suggestions:

* Use just a regular `map` (not `map2`).
* Rework `vecConstEq` so it takes a pair, so it can be used as an argument to `map`.
* Define a combinator in `Combinators.v` that "injects a first element: `injFst {a b : SignalType} (x : signal a) (y : signal b) : cava (signal a * signal b)`.
* Now you should be able to write `forkN (2^n) >=> injFst (vseq 0 (2^n)) >=> zipWith vecConstEq` to define the decoder.

CC: @jadephilipoom

atondwal:


satnam6502 (approved):
Thank you!
I have more ideas, but later...

PR #711, opened 2021-03-31
CavaClass: factor out indexConst

smore-lore:
CavaClass represents the abstraction layer for some subset of our
circuits. Limiting this surface limits the ability of an implementation
of CavaClass to conflict with itself, and increases the usefulness of
proofs against a specific implementation of CavaClass (usually
combinational semantics).

For instance, one undocumented behavior of indexConst is that when the
index exceeds the bounds of the provided vector, the default signal is
returned. In the existing implementation, we have no guarantee that this
semantic is preserved between NetlistGeneration and our cominational
semantics.

In this case, unpackV is provides a strict superset of functionality to
indexConst. Specifically, lifting the use of an underlying index
function above the CavaClass function, based on the semantics of
unpackV, should produce the exact same underlying code.

Contrast this with other cases (such as and/xor/etc..) for which
replacing them with compositions of each other may produce more complex
circuits. For instance, imagine one FPGA uses a different basis of gates
to achieve functional completeness.

The hope is that this CL establishes a place and methodology to simplify
CavaClass. For maximum assurance far into the future, CavaClass would
also need to expose a list of properties which are required of the
underlying implementation, and we would want to base proofs off of this
instead of the combinational semantics (or we would want proofs that
other semantics are in some sense isomorphic to the combinational
cases). In either case, this simplification process reduces the API
surface we have to live/deal with.


satnam6502 (approved):
As discussed offline, this looks like a great simplification and it is good to remove one "axiom" from `CavaClass`.

cava/Cava/Core/CavaClass.v
```diff
@@ -90,3 +85,15 @@ Class Cava (signal : SignalType -> Type) := {
...
+Require Import ExtLib.Structures.Monads.
+Import MonadNotation.
+
+Section Derivative.
+  Context {signal} `{Cava signal}.
+
+  Definition indexConst {t : SignalType} {sz : nat} (v : signal (Vec t sz)) (i : nat) : cava (signal t)
+    := v' <- unpackV v ;;
+       ret (nth_default defaultSignal i v').
+End Derivative.
```
jadephilipoom, line 99: I think this bit belongs in `Cava/Lib`. My recommendation would be to put it in `Cava/Lib/Vec.v` and call it `nth_default` for consistency; the other vector operators are all named to match the Coq vector library. And then, if you wouldn't mind, update https://github.com/project-oak/silveroak/blob/main/docs/reference.md to change the name!

jadephilipoom (approved):
Thanks for noticing this! I agree, it's great to remove things from `CavaClass` wherever we can.

cava/Cava/Core/CavaClass.v
blaxill: I agree with Jade, but we can also do this in a follow up if you want to merge first.

blaxill (approved):
Lgtm thanks!

cava/Cava/Core/CavaClass.v
smore-lore: Great - I'll merge now. I'm going to take a pause from this project for a while to work on some Android things. Thank you!

PR #712, opened 2021-04-01
Prove encode/decode

atondwal:
I could probably be better about using more automation.


tests/DecoderProofs.v
```diff
@@ -0,0 +1,347 @@
...
+    rewrite H.
+    trivial. }
+Qed.
+
+Lemma vec_const_eq_correct' : forall n k v,
+  VecConstEq.vecConstEq n k v = eqb (N2Bv_sized n (N.of_nat k), v).
+Proof.
+  intros.
+  cbv [VecConstEq.vecConstEq eqb].
+  simpl.
```
jadephilipoom, line 45: Try using the `simpl_ident` tactic, which pulls in existing lemmas to change the library circuits into their specifications. Haven't checked, but here I would try:
```suggestion
  cbv [VecConstEq.vecConstEq].
  simpl_ident.
```
`simpl_ident` will change `eqb` into `combType_eqb` on both sides.

You can add your own lemmas to the `simpl_ident` tactic using hints, e.g. https://github.com/project-oak/silveroak/blob/27d1002d2fe0424fda833f2f9283f7e580d1750a/cava/Cava/Lib/VecProperties.v#L43

tests/DecoderProofs.v
```diff
@@ -0,0 +1,347 @@
...
+  replace (N2Bv_sized n (N.of_nat k)) with (fst ((N2Bv_sized n (N.of_nat k)),v)).
+  2:trivial.
+  replace v with (snd ((N2Bv_sized n (N.of_nat k)),v)) at 3.
+  2:trivial.
+  rewrite <- @eqb_correct with (t:=Vec Bit n).
+  apply vec_const_eq_correct'.
+Qed.
+
+Lemma In_destruct : forall n A (a h:A) (v : Vector.t A n),
+  Vector.In a (h :: v) -> a = h \/ Vector.In a v.
```
jadephilipoom, line 69: As I suspect you've found, `Vector.In` is a real pain. We've defined our own more computation-friendly version here: 
 https://github.com/project-oak/silveroak/blob/27d1002d2fe0424fda833f2f9283f7e580d1750a/cava/Cava/Util/Vector.v#L1197-L1205

I recommend using it instead of the standard library version to avoid axioms, and we've even got a lemma about `map` there already.

blaxill:
Looking good!

atondwal:
Proofs are now complete. Seems the next step is to rewrite them using `InV` instead of `In` to avoid use of Eqdep? And then I figure I should probably move stuff like `map2_assoc` and the `InV` theorems to `Vector.v`. Is there a better place for me to put the decoder proofs (and what about `vec_consteq_correct`?)

satnam6502:
`Vec.v`
`VecProperties.v`
Is the pattern we have been following.

satnam6502:
Yay! Excellent, thank you very much!

Issue #715, opened 2021-04-02
Make a specific 8-bit adder instance of Xilinx IP Catalog Adder/Subtractor


satnam6502:
The objective is to iteratively refine our own plug-in compatible IP block for Xilinx's Adder/Subtractor core: https://www.xilinx.com/products/intellectual-property/adder_subtracter.html

PDF documentation: https://www.xilinx.com/support/documentation/ip_documentation/addsub/v12_0/pg120-c-addsub.pdf

When we're done we will have a Coq/Gallina function that will have arguments that correspond to the configuration options for the adder/subtractor. First let's do a simple combinational adder with two 8-bit unsigned inputs and one 9-bit output, configured as:

![image](https://user-images.githubusercontent.com/6164339/113445984-b4fb7080-93ab-11eb-9463-ec6eb51b64df.png)

This should produce an adder with the following interface:
```vhdl

COMPONENT c_addsub_0
  PORT (
    A : IN STD_LOGIC_VECTOR(7 DOWNTO 0);
    B : IN STD_LOGIC_VECTOR(7 DOWNTO 0);
    S : OUT STD_LOGIC_VECTOR(8 DOWNTO 0)
  );
END COMPONENT;
```

or in Verilog:

```verilog

c_addsub_0 your_instance_name (
  .A(A),  // input wire [7 : 0] A
  .B(B),  // input wire [7 : 0] B
  .S(S)  // output wire [8 : 0] S
);
```

The IP block should live in an `ip` subdirectory of `Cava` in a file `AdderSubtractor.v` with tests in the `silveroak/tests` directory.

This issue can be easily satisfied by simply instantiating our existing adder.
In subsequent issues we will add pipelining (clocking, registers) as well as the subtractor, or adder/subtract functionality.

PR #717, opened 2021-04-05
Add and test AdderSubstractor IP Core

atondwal:
towards replicating Xilinx Adder/Subtractor
cf #715


atondwal:
(git history's a bit of a mess, but I figure we're going to squash it anyways...)

satnam6502:
Thanks! Let me take my time to look over this carefully... hopefully tonight.

tests/AdderSubstractorTests.v
```diff
@@ -0,0 +1,54 @@
...
+Proof. trivial. Qed.
+
+Example ex2 :
+  c_addsub_0 (bv_511, bv_511) = N2Bv_sized 9 1022.
+Proof. trivial. Qed.
+
+Definition adderInterface
+  := combinationalInterface "c_addsub_0"
+     [mkPort "x" (Vec Bit 8); mkPort "y" (Vec Bit 8)]
+     [mkPort "z" (Vec Bit 9)].
```
satnam6502, line 37: Please can you name the port *exactly* as they are named in the generated Verilog/VHDL.
This is important for using these components as drop-in replacements.

satnam6502 (approved):
Great! Thanks.
I'll make a follow-up PR which runs the Xilinx tools to implement this so we can measure the resource utilization. I will also elaborate the Xilinx IP core to see if we can reuse any of its testbenches.

PR #718, opened 2021-04-06
Add sum8 implementation for Cava tutorial

satnam6502:
This PR add material to the sum example show:

- How to generate a circuit netlist.
- Write a testbench.
- Sample output of testbench SystemVerilog simulation.
- Viewing a VCD waveform of the simulation.
- Showing an ILA trace captured from the circuit running on an actual FPGA board.


demos/tutorial.v
```diff
@@ -921,14 +922,60 @@ version, just a delay connecting the loop's output to its own input.
...
-  := sequentialInterface "sum_interface"
+  := sequentialInterface "sum8"
      "clk" PositiveEdge "rst" PositiveEdge
      [mkPort "i" (Vec Bit n)]
      [mkPort "o" (Vec Bit n)].
 
 Compute
   (makeCircuitNetlist sum_interface (sum (n:=8))).(module).
 
+Local Open Scope N_scope.
```
jadephilipoom, line 933: Please remove this line, or put `..coq::none` before it so it doesn't show up in the generated HTML (looking at the code below, I think you may be just fine without opening `N_scope` here). At the tutorial level, I want to avoid opening up the can of worms that is notation scopes; I think it could be a distraction from the main points.

demos/tutorial.v
```diff
@@ -921,14 +922,60 @@ version, just a delay connecting the loop's output to its own input.
...
+
+Definition sum8Netlist := makeCircuitNetlist (sum_interface (n := 8)) sum.
+Definition sum8_tb_inputs := map (N2Bv_sized 8) [3; 5; 7; 2; 4; 6].
+Definition sum8_tb_expected_outputs :=  (simulate sum sum8_tb_inputs).
+
+Definition sum8_tb :=
+  testBench "sum8_tb" (sum_interface (n := 8)) sum8_tb_inputs sum8_tb_expected_outputs.
+
+(*|
+The circuit netlist and testbench can be converted in SystemVerilog and
```
jadephilipoom, line 943: Explain how the netlist is converted to SystemVerilog?

demos/tutorial.v
```diff
@@ -921,14 +922,60 @@ version, just a delay connecting the loop's output to its own input.
...
      [mkPort "o" (Vec Bit n)].
 
 Compute
   (makeCircuitNetlist sum_interface (sum (n:=8))).(module).
 
+Local Open Scope N_scope.  
+
+Definition sum8Netlist := makeCircuitNetlist (sum_interface (n := 8)) sum.
+Definition sum8_tb_inputs := map (N2Bv_sized 8) [3; 5; 7; 2; 4; 6].
+Definition sum8_tb_expected_outputs :=  (simulate sum sum8_tb_inputs).
```
jadephilipoom, line 937: Since this block hasn't appeared for previous examples, would you mind adding a little bit of intro text explaining that you're adding a test for the generated SV?

PR #721, opened 2021-04-06
Add dune build files without requiring dune

blaxill:
Enables building cava via dune but doesn't default to it


third_party/Makefile
```diff
@@ -39,6 +39,9 @@ build-extlib:
 build-record-update:
 	$(MAKE) -C coq-record-update
 
+build-koika:
+	$(MAKE) -C koika
```
jadephilipoom, line 43: Does this change add a Koika dependency?

jadephilipoom (approved):

third_party/Makefile
blaxill: Ahh this leaked in as I was building koika but dont intend to add it to third_party anymore: For Koika (and possibly other opam'able modules like coq-ext-lib/coq-record-update, when they have up to date releases and aren't making modifications) my intention is to require they be opam installed rather a custom build, only vendoring if we making changes.

jadephilipoom: I think that would be a great change!

Issue #722, opened 2021-04-06
Add makefile to create OpenTitan directories with replacements 


blaxill:
This is currently manually possible, but we want this to be as easy as possible for comparison etc.

PR #725, opened 2021-04-07
Add files for OpenTitan {gold,silveroak} comparison

blaxill:
- Update `silveroak-opentitan/aes` bash files
- Add `make diff_opentitan` to `silveroak-opentitan/aes/Makefile`
   This will create copies of `third_party/opentitan` as subdirectories `opentitan_gold` and `opentitan_silveroak` in `silveroak-opentitan/aes` with `opentitan_silveroak` containing silveroak replacements. Both directories should be buildable.
- Update doc

#722


satnam6502 (approved):
Thank you!

PR #728, opened 2021-04-08
Performance tweaks for Haskell execution

blaxill:
A bit of an extraction experiment: Having `nats` in extraction code is slow. This includes as constants and indices to vectors. I only modified SubBytes here as it was the easiest to modify, but it knocked 20s off `ghc --make`. Test vectors for `MixColumns` and `AddRoundKeyNetlist` are unmodified (they include vecs with nat indices) and still extract slow. Parallel ghc (`ghc -j`) halves the execution time.

### DIFF
--- a/silveroak-opentitan/aes/Impl/Makefile
+++ b/silveroak-opentitan/aes/Impl/Makefile
@@ -62,7 +62,7 @@ coq:		Makefile.coq
 		$(MAKE) -f Makefile.coq
 
 AESSV:		coq AESSV.hs
-		ghc --make AESSV.hs
+		ghc --make -j AESSV.hs
 		./AESSV
 
 minimize-requires: coq
--- a/silveroak-opentitan/aes/Impl/SubBytesCircuit.v
+++ b/silveroak-opentitan/aes/Impl/SubBytesCircuit.v
@@ -20,7 +20,8 @@ Import Pkg.Notations.
 Local Open Scope vector_scope.
 
 Section SboxLut.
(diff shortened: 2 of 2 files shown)


satnam6502:
Noted!

jadephilipoom (approved):
I'm happy to make this change for `sub_bytes`. `N` is faster but more difficult in proofs because induction on `N` is on bits, and natural-number induction is often better. It doesn't matter for `sub_bytes` because the proofs are all by computation, not induction.

However, I think before making these changes to any other part of the AES block we need to consider whether we're committing to Haskell extraction long-term. We've previously discussed porting the Haskell code to Coq in order to remove extraction and ghc from our TCB, and if we ended up connecting to other tools in Coq (e.g. Koika) then extraction would likely be cut out anyway. I wouldn't want to make tradeoffs in favor of faster Haskell execution if we're just going to get rid of the Haskell later.

blaxill:
> However, I think before making these changes to any other part of the AES block we need to consider whether we're committing to Haskell extraction long-term. 

Yes this is why I didn't change the other immediately obvious cases. I am curious if we will run into performance issues with pure Coq generation though as Haskell "shouldn't be slow". 

> ... and if we ended up connecting to other tools in Coq (e.g. Koika) then extraction would likely be cut out anyway. 

The current plan would be to use Cava modules as Koika external modules which would still require us to generate SystemVerilog outside the Koika flow (whether Haskell or Coq based).

PR #729, opened 2021-04-12
Add bedrock2 experiment with small circuit

jadephilipoom:
Everything here is in `investigations`, so it should have no effect on CI.

This PR:
- does a little organization in the preexisting `investigations/bedrock2` directory (adds a Makefile and moves earlier `aes_init` experiment to its own directory)
- Adds bedrock2 circuit spec, firmware program, and proof of program correctness for a very simple signalling-incrementer circuit

OpenTitan's AES block works, in broad strokes, by consuming input that software writes to registers, writing output to registers, and signalling when the output is ready by setting a flag in a special "status" register that software can observe. The firmware reads the status register until its value indicates that the AES block is done processing, and only then reads the output. For writing input, the firmware similarly observes the status register until it indicates the circuit is ready to receive new input.

To start off the bedrock2 experimentation, I designed an extremely simple circuit to mimic this signalling strategy. It has two registers, `VALUE` and `STATUS`. `STATUS` has three flags: `IDLE`, `BUSY`, and `DONE`. Like in AES, the software can only interact with the circuit by reading/writing registers. 

The circuit spec is based on a state machine with the state options:
```coq
  Inductive state :=
  | IDLE
  | BUSY (input : word)
  | DONE (answer : word)
  .
```
Note that this is the state of the circuit _from the software's perspective_; the software might read a `BUSY` status and think the circuit is still `BUSY`, but in the meantime it might have finished processing and actually be `DONE`. In other words, this state is meant to track what we already know about the circuit, and cannot predict its outputs in general.

The state transitions are defined by the inductive `step` predicate (small-step semantics), which for a given start state, action (read or write), list of argument values, list of return values, and new state, states that the action with specified arguments and return values is a valid transition between the start state and new state. The transition rules are:
- Write new input: `IDLE -> write [VALUE; v] [] -> BUSY v`
- Finish processing: `BUSY v -> read [STATUS] [DONE] -> DONE (v+1)`
- Read output: `DONE v -> read [VALUE] [v] -> IDLE` 
- No state change: `forall s : state, s -> read [STATUS] [<expected status for s>] -> s`

Big-step semantics are described by `execution`, which is a predicate on a bedrock2 trace and a state that says the given state represents the observed state of the circuit after the trace has been executed.

Some notes:
- The timing properties of the circuit are not visible to software; there's no way for the program to know when the circuit will indicate that it's `DONE`
- Currently, we're assuming our program is the only thing interacting with the circuit (no concurrently operating programs)
- The program currently just checks if the circuit is done for a specified number of attempts before exiting (in bedrock2, it's possible to write infinite loops but in order to prove anything about them you need to prove somehow that they terminate; the number of attempts is a kludge that fills this requirement)
- In the C printout, the constants are literals; the `aes.c` code uses a number of global constants imported from `aes_regs.h`, so it would be ideal to print these constants by name
- There's a lot of spaghetti proof because bedrock2's proof automation appears to assume the word implementation is concrete rather than abstract. We could fix this by making it concrete, but that seems like a leaky abstraction to me

Next steps:
- Replace the "number of attempts" kludge with a more realistic `while (true)` loop and a more sophisticated state machine that also tracks some circuit timing properties. We know that software can't read `STATUS` more than once per cycle, and this circuit would have deterministic timing properties. Therefore, we should be able to write a provable spec for the circuit that specifies it is guaranteed to be `DONE` after a certain number of `STATUS=BUSY` reads.
- Fix the handling of constants so that C printouts can print them as C
- Write the corresponding circuit and prove that it satisfies the spec?
- The version of bedrock2 we're using is pretty old and it would be good to update it; this requires moving to Coq 8.13 (which we probably also want to do anyway)
- Once this small example looks good, try to apply the same strategy to write a spec for the AES block and prove correctness of functions like `aes_data_put_wait`

CC @atondwal


investigations/bedrock2/IncrementWait/IncrementWaitSemantics.v
```diff
@@ -0,0 +1,192 @@
...
+
+  Global Instance ok : Semantics.parameters_ok semantics_parameters.
+  Proof.
+    split; cbv [env locals mem semantics_parameters]; try exact _.
+    { cbv; auto. }
+    { exact (SortedListString.ok _). }
+    { exact (SortedListString.ok _). }
+  Qed.
+
+  (* COPY-PASTE this *)
```
blaxill, line 187: Hmm, not sure what this comment is referring to?

investigations/bedrock2/incrementwait.c.out
```diff
@@ -0,0 +1,42 @@
+     = "#include <stdint.h>
+#include <memory.h>
+
```
blaxill, line 3: In general I'd prefer not to include generated files, what are your thoughts?

investigations/bedrock2/IncrementWait/IncrementWaitProperties.v
```diff
@@ -0,0 +1,344 @@
...
+           | H : reg_addr _ _ |- _ => invert_nobranch H
+           | H : read_step _ _ _ _ |- _ => invert_nobranch H
+           | H : execution _ (DONE _) |- _ => invert_nobranch H; destruct_products
+           | H : step _ _ _ _ (DONE _) |- _ => invert_nobranch H
+           | H1 : execution ?t _, H2 : execution ?t _ |- _ =>
+             pose proof execution_unique _ _ _ H1 H2; subst;
+             clear H2; one_goal_or_solved ltac:(try congruence)
+           end.
+
+  Lemma put_wait_get_correct :
```
blaxill, line 166: Is it possible to break this proof into smaller subproofs or perhaps the intermediate goals are complex and so its not particularly meaningful? Don't worry about actually doing it here for the investigation.

blaxill (approved):
Looking good! A few minor comments inline. 

> Write the corresponding circuit and prove that it satisfies the spec?
This sounds worth doing- I'll do it, assuming you don't already have a stacked PR waiting?

Also general question: Am I right saying bedrock2 semantics are prop based and not `Compute`able, does this give added benefit in someway? Perhaps we wouldn't need to e.g. avoid type rewriting with `resize_default` & co. that we do in Cava- if we were to iterate on Cava in the future is this worth looking into?

investigations/bedrock2/IncrementWait/IncrementWaitProperties.v
jadephilipoom: Will definitely make this nicer before I'm done with it; it's just spaghetti code for the first pass so I could get something working as soon as possible.

investigations/bedrock2/incrementwait.c.out
jadephilipoom: I'd like to include it in this investigation because it will help me track how the output changes. I'm happy to not include generated files if I move this stuff out of `investigations`.

investigations/bedrock2/IncrementWait/IncrementWaitSemantics.v
jadephilipoom: The comment is copied from bedrock2's `FE310Semantics.v`, which was my template for these semantics. It means that if you want to use the `ring` tactic on words, you can copy the line below and add it to your file (with your parameter instance in context).

jadephilipoom:
> > Write the corresponding circuit and prove that it satisfies the spec?
> This sounds worth doing- I'll do it, assuming you don't already have a stacked PR waiting?

No stacked PR waiting, although I will continue to iterate on this example. Maybe wait on this for now -- changing to a circuit spec that includes some timing properties might change the circuit spec significantly.

> Also general question: Am I right saying bedrock2 semantics are prop based and not `Compute`able, does this give added benefit in someway? Perhaps we wouldn't need to e.g. avoid type rewriting with `resize_default` & co. that we do in Cava- if we were to iterate on Cava in the future is this worth looking into?

The bedrock2 semantics are "weakest-precondition" style ([wikipedia overview](https://en.wikipedia.org/wiki/Predicate_transformer_semantics#Weakest_preconditions)). It's not quite as simple as "not computable". They _are_ computable; given a postcondition and a function body, you can compute the `Prop` representing the goal you need to satisfy for the postcondition to hold after the function. So, for instance, instead of computing the state of the memory after the function, you get a `Prop` that includes at some point "if your memory includes x and y, then that (plus a bunch of other context) implies your postcondition".

As to whether we should look into it for Cava, I'm not sure what problem it would solve for us. We could try to write circuit semantics in weakest-precondition style, but it would make our semantics much less intuitive. I think bedrock2 uses this to get around things like not necessarily knowing the complete state of the memory, but that's not a problem we have; we can compute the full state of every circuit at every timestep.

PR #735, opened 2021-04-12
Use a while(true) loop in bedrock2 experimental circuit

jadephilipoom:
In #729 I mentioned that the bedrock2 firmware experiment had a bit of a hacky solution in it to prove that its while loop terminated; the program took an argument that was a maximum number of times to try reading the result before giving up. This PR fixes that and changes the loop to a more realistic `while (true)` loop. The logic is now much more similar to `aes.c`.

Although the software is still not directly aware of circuit timing information, the circuit specification now includes, for the `BUSY` state, a maximum number of future reads that can happen before the state is `DONE` (the circuit is permitted to move to the `DONE` state sooner). It should be possible to prove that a circuit meets this specification by simply plugging in the number of cycles for the maximum reads; each read takes one cycle, so the guarantee should hold.

### DIFF
--- a/investigations/bedrock2/IncrementWait/IncrementWait.v
+++ b/investigations/bedrock2/IncrementWait/IncrementWait.v
@@ -20,42 +20,26 @@ Section Impl.
   Local Notation "0" := (expr.literal 0) (in custom bedrock_expr).
   Local Notation "1" := (expr.literal 1) (in custom bedrock_expr).
 
-  (* Sets the VALUE register, waits for a DONE status until a particular timeout
-     (in terms of number of attempted reads), reads the result, and returns
-     it. Fails if the circuit is not in IDLE state at the start, or if the
-     status is not DONE by the time the timeout is reached. *)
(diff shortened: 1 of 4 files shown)


investigations/bedrock2/IncrementWait/IncrementWait.v
```diff
@@ -20,42 +20,26 @@ Section Impl.
...
+        (* write input value *)
+        output! WRITE (VALUE_ADDR, input) ;
+        (* initialize status to 0 *)
+        status = 0 ;
+        (* wait for status to be DONE *)
+        while ((status & (1 << STATUS_DONE)) == 0) {
+          io! status = READ ( STATUS_ADDR )
+        };
+        (* read the value and exit *)
+        io! out = READ ( VALUE_ADDR )
```
jadephilipoom, line 43: New program body, for easier scanning among all the proof changes

blaxill (approved):
Lgtm!

PR #737, opened 2021-04-13
Option 1 of 2: Interpret bedrock2 constants as literals for proofs, vars for printing

jadephilipoom:
In #729 I mentioned as part of "next steps" that the constant values were printed as literals, and therefore couldn't be copy-pasted into an environment using global constants. The crux of the problem here is that bedrock2 doesn't support global constants at all (the proof logic starts functions bodies with an empty context).

This is 1 out of 2 possible solutions to the problem. Here, I've defined a `constant` typeclass, and a typeclass for interpretation functions that change a `constant` into a bedrock2 `expr`. There are two instances of the interpretation typeclass: one interprets the constants using `expr.var` (that is, by referencing the string part of the constant as a variable) and the other interprets the constants using `expr.literal` (referencing the Z part of the constant as a literal value).

The end result is that in proofs, we reason about constants as if they were literals (although the proof is still parameterized over their exact values), but the C printout shows them as local variables:
```
uintptr_t put_wait_get(uintptr_t input) {
  uintptr_t STATUS_DONE, status, STATUS_ADDR, out, VALUE_ADDR;
  REG32_SET(VALUE_ADDR, input);
  status = (uintptr_t)0ULL;
  while (((status)&(((uintptr_t)1ULL)<<(STATUS_DONE)))==((uintptr_t)0ULL)) {
    status = REG32_GET(STATUS_ADDR);
  }
  out = REG32_GET(VALUE_ADDR);
  return out;
}
```

One major disadvantage of this approach is that it does not protect against global constants being overwritten. For instance, if I had written a new value to a local variable called `STATUS_ADDR`, the proof logic could still use the literal value from the global constant. This is a bit of a weird edge case, but still is a concern.

Another minor disadvantage is that the global variables declarations at the top of the function will need to be removed before copy-pasting this into a C file.

CC @atondwal

### DIFF
--- a/investigations/bedrock2/Constants.v
+++ b/investigations/bedrock2/Constants.v
@@ -1,0 +1,23 @@
+Require Import Coq.Strings.String.
+Require Import Coq.ZArith.ZArith.
+Require Import bedrock2.Syntax.
+Require Import coqutil.Word.Interface.
+
+Module constant.
+  (* Typeclass for constant values *)
+  Definition constant (name : string) : Type := Z.
+  Existing Class constant.
+
+  (* Typeclass for functions that convert a constant to a bedrock2 expression *)
+  Definition interp : Type := forall name, constant name -> expr.expr.
+  Existing Class interp.
+
+  (* Use `Existing Instance constants.var_interp` to interpret constants as
+     named variables -- useful for printing C that will use those constants *)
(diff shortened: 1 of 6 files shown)


jadephilipoom:
Closed in favor of #738

PR #738, opened 2021-04-13
Option 2 of 2: Interpret bedrock2 constants as arguments

jadephilipoom:
Alternative to #737; only one of the two should be merged.

This solves the same problem as #737 but with a different method. Here, the constants are interpreted as _arguments_ to the function, both in the C printout and in the proofs. The C printout looks like:
```
uintptr_t put_wait_get(uintptr_t VALUE_ADDR, uintptr_t STATUS_ADDR, uintptr_t STATUS_IDLE, uintptr_t STATUS_BUSY, uintptr_t STATUS_DONE, uintptr_t input) {
  uintptr_t status, out;
  REG32_SET(VALUE_ADDR, input);
  status = (uintptr_t)0ULL;
  while (((status)&(((uintptr_t)1ULL)<<(STATUS_DONE)))==((uintptr_t)0ULL)) {
    status = REG32_GET(STATUS_ADDR);
  }
  out = REG32_GET(VALUE_ADDR);
  return out;
}
```

Compared to #737, this solution:
- (+) avoids the issue in #737 where overwriting global constants can lead to discrepancies between C and proofs
- (+) is likely easier to copy-paste; the functions have extra arguments, but these will all be in the exact same order and can be removed with a quick `sed`
- (-) has more boilerplate; constants have to be listed multiple times and loop invariants need to prove that global constants were not overwritten

Curious for thoughts on which of these solutions is better!

CC @atondwal

### DIFF
--- a/investigations/bedrock2/IncrementWait/IncrementWaitSemantics.v
+++ b/investigations/bedrock2/IncrementWait/IncrementWaitSemantics.v
@@ -20,24 +20,40 @@ Definition unique_words
   List.dedup word.eqb l = l.
 
 Module constants.
-  Class constants :=
-    { (* compile-time constants *)
-      VALUE_ADDR : Z;
-      STATUS_ADDR : Z;
+  Class constants T :=
+    { VALUE_ADDR : T;
+      STATUS_ADDR : T;
+      STATUS_IDLE : T;
+      STATUS_BUSY : T;
+      STATUS_DONE : T }.
 
-      (* positions of status flags *)
-      STATUS_IDLE : Z;
-      STATUS_BUSY : Z;
-      STATUS_DONE : Z;
-    }.
+  Definition constant_vars
+           {names : constants string}
+    : constants expr :=
+    {| VALUE_ADDR := expr.var VALUE_ADDR;
+       STATUS_ADDR := expr.var STATUS_ADDR;
+       STATUS_IDLE := expr.var STATUS_IDLE;
+       STATUS_BUSY := expr.var STATUS_BUSY;
+       STATUS_DONE := expr.var STATUS_DONE |}.
 
-  Class ok {word : word.word 32} {word_ok : word.ok word}
-        (consts : constants) :=
-    { addrs_unique : word.of_Z VALUE_ADDR <> word.of_Z STATUS_ADDR;
+  Definition constant_names : constants string :=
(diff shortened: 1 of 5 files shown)


atondwal:
At least thus far, this seems preferable to me --- the side conditions on each invariant that it preserves globals should always be trivial. (We could probably have one tactic that takes care of them all and saves just much of the boilerplate, at least at the proof level.)

blaxill:
> * (-) has more boilerplate; constants have to be listed multiple times and **loop invariants need to prove that global constants were not overwritten**

This should be relatively straight forward though? Or in more complex loops would this be an issue?

jadephilipoom:
In answer to both questions -- yes, it is very easy to automate the proofs that constants are not overwritten. There's actually a tactic that solves these cases in the code already (`map_lookup`).

investigations/bedrock2/IncrementWait/IncrementWaitSemantics.v
```diff
@@ -20,24 +20,40 @@ Definition unique_words
...
-      (* positions of status flags *)
-      STATUS_IDLE : Z;
-      STATUS_BUSY : Z;
-      STATUS_DONE : Z;
-    }.
-
-  Class ok {word : word.word 32} {word_ok : word.ok word}
-        (consts : constants) :=
-    { addrs_unique : word.of_Z VALUE_ADDR <> word.of_Z STATUS_ADDR;
+  Class constants T :=
```
blaxill, line 23: This could be specialized to `word`?

blaxill (approved):
Lgtm! This method seems to be preferable 👍

investigations/bedrock2/IncrementWait/IncrementWaitSemantics.v
jadephilipoom: There are actually three instances with different `T` arguments. One is `constants string`, and provides the names of the constants. The other is `constants expr` and is derived from the `string` one -- it just applies `expr.var` (variable expression constructor) to each field. The last (which is not actually instantiated here, but is in the `Context` for proofs) is `constants word`, and provides the actual values. The reason for all the different instances is so that I can write e.g. `VALUE_ADDR` in different contexts for all different ways of referring to the constant. I can write it directly into a program and get a bedrock2 `expr`, write it into the circuit spec and get a `string`, and write it in the proofs and get a `word`.

Issue #672, opened 2021-03-19
Implement a binary decoder library module


satnam6502:
Following the pattern for `vecConstEq`, implement a binary decoder circuit:
```coq
Require Import Coq.Arith.PeanoNat.

Section WithCava.
  Context `{semantics:Cava}.

  Definition decoder {n : nat} (bv : signal (Vec Bit n)) :
                     cava (signal (Vec Bit (2^n))).

End WithCava.
```

The output is a [one-hot](https://en.wikipedia.org/wiki/One-hot) encoding of the input.

Define in `Decoder.v` in `Lib` and `DecoderTest.v`. in `tests`.

Use `vecConstEq` to help make this circuit. For example, the first output bit can be defined as `vecConstEq 0 input`.

PR #739, opened 2021-04-13
Decoder: proof of adjointness, reverse direction

atondwal:
cf  #672

### DIFF
--- a/cava/Cava/Lib/DecoderProperties.v
+++ b/cava/Cava/Lib/DecoderProperties.v
@@ -38,36 +38,40 @@ Existing Instance CombinationalSemantics.
 
 Import Circuit.Notations.
 
-Lemma fold_mono {A n} {a:A} {f unit decider}
-  (dec_true : forall a b, decider a b = true <-> a = b)
-  (right_id : forall b, f b unit = b)
-  (neutral : forall b, f b b = b)
+Lemma fold_mono {A B n} {a:A} {b: A-> B} {f unit decider}
+  (dec_true : forall c d, decider c d = true <-> c = d)
+  (right_id : forall c, f c unit = c)
+  (neutral : forall c, f c c = c)
   (vec : Vector.t A n) :
-  Vector.fold_left f a (Vector.map (fun v => if decider a v then a else unit) vec) = a.
+  Vector.fold_left f (b a) (Vector.map (fun v => if decider a v then b v else unit) vec) = b a.
 Proof.
   induction vec; [ trivial | ].
   cbn.
-  case_eq (decider a h);
-    intros;
-    [rewrite neutral | rewrite right_id];
-    apply IHvec.
+  case_eq (decider a h); intro H.
+  { rewrite dec_true in H. subst.
+    rewrite neutral.
+    apply IHvec. }
+  { remember (dec_contrapositive dec_true) as dec_false.
+    rewrite dec_false in H. subst.
+    rewrite right_id.
+    apply IHvec. }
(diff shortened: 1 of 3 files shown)


atondwal:
cf  #672

cava/Cava/Lib/DecoderProperties.v
```diff
@@ -193,3 +197,195 @@ Proof.
...
+  }
+Qed.
+
+Theorem map_ext_guard
+  (A B : Type) (f g : A -> B)
+  (P : A -> Prop)
+  (ext : forall a : A, P a ->  f a = g a)
+  (n : nat) (v : VectorDef.t A n)
+  (Guard: forall a : A, InV a v -> P a)
+  : VectorDef.map f v = VectorDef.map g v.
```
blaxill, line 224: I think `VectorDef` here is the same as using `Vector`. Although its just a module name alias in Coq, Haskell doesn't like extractions that contain mixed usage of `VectorDef` and `Vector`.

cava/Cava/Lib/DecoderProperties.v
```diff
@@ -193,3 +197,195 @@ Proof.
...
+    reflexivity.
+    }
+  { intros. apply Guard.
+    cbn. right.  assumption. }
+Qed.
+
+Lemma N_lt_step a : (N.of_nat a < N.of_nat (S a))%N.
+Proof. rewrite Nat2N.inj_succ. apply N.lt_succ_diag_r. Qed.
+
+Lemma N_lt_inj a b : a < b -> (N.of_nat a < N.of_nat b)%N.
```
jadephilipoom, line 242: This proof might be easier using `Nat2N.inj_compare`, `Nat.compare_lt_iff`, and `N.compare_lt_iff`.

cava/Cava/Lib/DecoderProperties.v
```diff
@@ -193,3 +197,195 @@ Proof.
...
+            { rewrite <- eq. apply Nat.lt_succ_diag_r. }
+          }
+          { apply N_lt_step. }
+        }
+      }
+    }
+  }
+Qed.
+
+Theorem dec_enc_inv (n:nat) (k : {i: nat | i < 2^n}):
```
jadephilipoom, line 271: Why a sigma type here? The proof will likely be easier to apply later with the proof and `nat` separate, e.g.:
```suggestion
Theorem dec_enc_inv (n:nat) (k : nat):
  k < 2^n ->
```
This way, if you _do_ have a sigma type in context, you can still use the lemma with `k:=proj1_sig x` and then prove the side condition with `proj2_sig x`. But you wouldn't need to construct a sigma type in order to use the lemma if that's not what your context looks like.

cava/Cava/Util/Bool.v
```diff
@@ -23,3 +23,8 @@ Proof.
     exfalso.
     apply notQ. apply H. trivial. }
 Qed.
+
+Lemma dec_contrapositive {A decider}
```
jadephilipoom, line 27: Might be a good idea to use `coqutil`'s `BoolSpec` for this one; it'll get picked up by nice automation like the `destr` tactic.

jadephilipoom (approved):
LGTM

cava/Cava/Lib/DecoderProperties.v
atondwal: Hmm, somehow `Search Nat.lt` didn't give me anything useful ... thanks!

PR #741, opened 2021-04-14
Upgrade to Coq 8.13

### LINKED ISSUE #730: Update to Coq 8.13
[Coq 8.13 is out](https://coq.inria.fr/news/coq-8-13-0-is-out.html) (as of January). We should make sure we're compatible by updating the CI to either check both 8.13 and 8.12, or just 8.13. If the latter, we should update our own local installations as well.

I'd personally lean towards checking just one version at this stage of the project; although it's nice to support multiple, it can slow down PRs and it's annoying to debug a version of Coq you don't have.

jadephilipoom:
~Testing CI, please ignore for now.~ CI passed!

Resolves #730 

This PR upgrades us to use Coq 8.13 in the CI. It is most likely not a backwards-compatible change due to a change to module names in the Haskell extraction, and possibly also due to the coq-record-update upgrade that was needed for 8.13 to build. Before this PR is merged, everyone should update their local development environment to 8.13 or higher. I've upgraded mine to 8.13.2 and everything builds, so any 8.13.x version is likely fine.

### DIFF
--- a/cava/Cava2HDL.cabal
+++ b/cava/Cava2HDL.cabal
@@ -87,7 +87,7 @@ library
                      Byte
                      Decimal
                      Hexadecimal
-                     Numeral
+                     Number
                      Specif
                      MonadExc
                      MonadFix
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ requests.
 
 Please install the following components:
 
-* The [Coq proof assistant](https://coq.inria.fr/) version 8.12.0.
+* The [Coq proof assistant](https://coq.inria.fr/) version 8.13.0.
 * The [GHC Haskell compiler](https://www.haskell.org/ghc/) version 8.6.5 or later
 * [Verilator](https://www.veripool.org/wiki/verilator) version 4.028 (as specified by the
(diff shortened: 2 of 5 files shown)


jadephilipoom:
To be clear, I'm happy to hold off on this change for a while as people make local upgrades.

blaxill (approved):
LGTM

> Before this PR is merged, everyone should update their local development environment to 8.13 or higher. I've upgraded mine to 8.13.2 and everything builds, so any 8.13.x version is likely fine.

Is it specifically failing on 8.12? I'm in the process adding multi Coq-version CI builds, it would be good if we could still build on 8.12  (at least partially)

cava/Cava2HDL.cabal
```diff
@@ -87,7 +87,7 @@ library
                      Byte
                      Decimal
                      Hexadecimal
-                     Numeral
+                     Number
```
jadephilipoom, line 90: @blaxill I haven't checked for sure whether it fails on 8.12, but I think this naming change is likely to be an issue.

blaxill: Good point, don't worry about it for now!

Issue #742, opened 2021-04-14
Source tree clean up and building and CI modernisation 


blaxill:
In the direction of cleaning up source trees, adding support for matrix CI builds and dune building, I'm proposing the following:

- Keep everything simply buildable via `make`
- [ ] - Make haskell files built by cabal by default
- [ ] - Add dune build files but dont make the switch yet
- [ ] - Upgrade nix files to support multi coq,ghc,verilator versions 
- _to be expanded_

Then once everything has settled:
- Add a dependency on dune (everything will still be buildable via `make`)
- _to be expanded_

And should enable:
- Generated non-source files (`.vo` etc), kept in build directories (dune -> _build, cabal -> dist) 
- Generated source files (`.hs`, `.sv`, etc) in-tree in marked directories

PR #743, opened 2021-04-14
Add cabal support to Haskell modules

blaxill:
- Removes the requirement to install Cava2HDL globally

#742

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,4 +24,7 @@
 **/obj_dir
 **/*.vcd
 **/*/_CoqProject
+**/dist/
+**/dist-newstyle/
+_build/
 cava.code-workspace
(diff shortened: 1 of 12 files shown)


jadephilipoom (approved):

satnam6502:
What's the reason to use cabal to build programs other than the Cava2HDL library?
I would have guessed it is simpler and less brittle to just `ghc make` these binaries?

blaxill:
Cava2HDL is the main reason for using cabal, the other binaries could be made with ghc make. Apart from some initial speed bumps cabal should not be more brittle than ghc.

atondwal:
Modern cabal (after new-build etc became the default c. cabal v3.0) should
actually be less brittle than `ghc --make`!

PR #744, opened 2021-04-15
Cava/ip -> Cava/IP

blaxill:
Coq expects the directory name to match module name usage. The module is used with a capitalised path and so the directory should be capitalised too.


satnam6502:
Sorry for getting this wrong. Thanks for fixing it.

atondwal:
lol, I put a workaround for this in the `Makefile` because I thought it was intentional :upside_down_face:

Issue #746, opened 2021-04-15
Cabal build is not working for me


jadephilipoom:
I don't know if I have the wrong cabal version or a missing dependency (`base`?), but after a checkout of `main` and a `make clean` I get this error from the cava haskell target:
```
./FixupHaskell
mv Ascii2.hs Ascii.hs
mv ByteVector2.hs ByteVector.hs
cabal build
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: Cava2HDL-0.1.0.0 (user goal)
[__1] trying: base-4.13.0.0/installed-4.1... (dependency of Cava2HDL)
[__2] next goal: cava-examples (user goal)
[__2] rejecting: cava-examples-0.1.0.0 (conflict:
base==4.13.0.0/installed-4.1..., cava-examples => base>=4.14 && <4.15)
[__2] fail (backjumping, conflict set: base, cava-examples)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: base, Cava2HDL, cava-examples
```

satnam6502:
I've not been following along about how we got here, but why are we using cabal for the examples?

jadephilipoom:
That's probably it, then -- I have version 8.8.4. That's also the latest version on the gLinux package manager, though, and I'd prefer not to have to maintain a from-source installation if possible.

blaxill:
> I've not been following along about how we got here, but why are we using cabal for the examples?

There are definitely some benefits to using cabal:
- builds are sandboxed and don't require installing packages system wide
- project level configuration effecting our cava/pinmux/aes etc 
- cabal is pretty ubiquitous in the haskell world these days, and issues from prior years around cabal dependency solving issues aren't such an issue anymore 

The cabal changes will be easiest to roll back if there are usability issues, but i expect things to be simpler going forward (after potentially some speed bumps such as this). In the long run I still see us potentially removing Haskell and moving all logic to Coq.  

I'd like to make a few more changes to simplify our source tree in #742 such that we only have source/project files in our source tree and artefacts/compilation objects are put into other directories (dune being the simplest way to achieve this for Coq). That in turn will simplify browsing through the directory hierarchy but also more tangible benefits like simplifying our makefile building and cleaning rules.

blaxill:
> That's probably it, then -- I have version 8.8.4. That's also the latest version on the gLinux package manager, though, and I'd prefer not to have to maintain a from-source installation if possible.

[...]

PR #749, opened 2021-04-15
Relax cabal GHC dependency and update documentation

blaxill:
#746

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ requests.
 Please install the following components:
 
 * The [Coq proof assistant](https://coq.inria.fr/) version 8.13.0.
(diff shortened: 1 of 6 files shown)


jadephilipoom (approved):
The compiler version is still higher than the one I have, but let's see if it works.

blaxill:
> The compiler version is still higher than the one I have, but let's see if it works.

I think you're at 8.8.4? Thats newer than 8.6 (base 4.12 corresponds to 8.6.*)

jadephilipoom:
> I think you're at 8.8.4? Thats newer than 8.6 (base 4.12 corresponds to 8.6.*)

Yes, you're right -- sorry, I misrememered it as 8.4. Thanks!

PR #751, opened 2021-04-15
Update README to include cabal dependency

### LINKED ISSUE #745: Document cabal dependency
After #743 building the Haskell targets requires `cabal` (just got a build error due to not having it). We should add this dependency to our README. @blaxill, do you know what version we require?

atondwal:
fixes #745

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Please install the following components:
 
 * The [Coq proof assistant](https://coq.inria.fr/) version 8.13.0.
 * The [GHC Haskell compiler](https://www.haskell.org/ghc/) version 8.6 or later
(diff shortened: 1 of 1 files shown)


atondwal:
nix/glinux both pull in 3.4.0.0, but it should work fine with 3.0.0.0

blaxill (approved):
Perhaps even go down to cabal 2.4 as it is a closer match to ghc 8.6 I think? (in fact older versions of cabal should work for us too) Lgtm either way thanks

atondwal:
I don't think we want to be in the business of supporting both 2.4 (old-style) and 3.0 (nix-style) building; that's bound to get confusing. (though one can always get Cabal 2.4 to bevave like cabal 3.0 with v2-build, etc)

PR #752, opened 2021-04-15
Consistent name for make target `clean_all`

atondwal:
This was called `clean_all` elsewhere

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ clean:
 	done
 
 # clean everything *including* third_party
-cleanall:
+clean_all:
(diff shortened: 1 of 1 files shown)


jadephilipoom:
> This was called `clean_all` elsewhere

Where? We've generally used `-` over `_` in our make targets, I think, and I'd like to keep that consistent. With a quick grep I can't find any instance of `clean_all` outside `third_party`.

atondwal:
Ah, yep, I was thinking of in bedrock. What was throwing me off was that I was looking in my shell history and kept getting the wrong one.

PR #753, opened 2021-04-16
[investigations] Strengthen aes_init proof

jadephilipoom:
The `aes_init` proof is now much less trivial; instead of saying that the value in the control register matches the sequence of ands and ors that is in the code, the specification now says that there exists some value in that register such that, when we mask out the bits for each control-register field, we get the right value. This required a lot of bit-arithmetic logic that was a bit painful but probably useful to set up. Some of the more obvious bit-arithmetic lemmas are currently admitted.

Next step, I think, is to do a little organization and put the bit-arithmetic proofs in their own file, which will make the main files shorter and easier to work with. Then I'll try some of the other functions from `aes.c`.

CC @atondwal


blaxill (approved):
Lgtm!

PR #755, opened 2021-04-20
[WIP] first crack at pipelining addsub

atondwal:
This is my first time working with `Circuit` and timing, so I'm not sure this is exactly what you had in mind, but here's the shape of what I'm thinking re: pipelining the adder


satnam6502:
I don't totally understand what is going on here. Let's talk about this over video chat.

satnam6502:
This has been closed for the moment since we no longer have Anish wish us. We may pick up this thread later.

PR #756, opened 2021-04-20
Simplifying netlist/tupling

blaxill:
- Pulls tupling operations into generic tuple module that also can do function currying. Currying is a bit weird as tupleR are "reversed" coq tuples that have first arg as first element of the tuple, but of course Coq tuples will be the other way around so there are two curried representations `curried`&`curriedR`
- Added `TaggedEdge` as clock/reset are sometime optional but always passed around with edge
 
Possibly quite a few ways to simplify things in the future but this starts to reduce some of the complexity of tupling


cava/Cava/Lib/VecProperties.v
```diff
@@ -126,17 +126,19 @@ Hint Rewrite @shiftout_correct using solve [eauto] : simpl_ident.
 
 Lemma transpose_correct A n m (v : combType (Vec (Vec A n) m)) :
   Vec.transpose v = transpose v.
-Proof. crush. Qed.
+Proof. crush.
```
blaxill, line 129: @jadephilipoom my changes have broken automation here, i can fix it manually but I can't seem to figure out the changes required to `crush`/`simpl_ident` without forcing rewrites of `Vec.unpackV2`/`Vec.packV2`

cava/Cava/Semantics/Combinational.v
```diff
@@ -106,12 +122,14 @@ Create HintDb simpl_ident.
 Hint Rewrite @foldLM_ident_fold_left using solve [eauto] : simpl_ident.
 Ltac simpl_ident :=
   (* simplify identity monad and most projections from Cava *)
-  cbn [fst snd bind ret Monad_ident monad
+  cbn [fst snd bind ret Monad_ident monad mcompose
```
jadephilipoom, line 125: If I had to guess, I'd guess that this change is what broke the automation, just because none of the other changes should have affected `Vec`. If I need to simplify `mcompose`, I've been just writing `cbv [mcompose]` before calling `simpl_ident` (the `crush` tactic does do so, so it's a bit odd that adding it would cause issues, but I'd still try removing it and see if that fixes the `Vec` proofs). `mcompose` used to be in this `cbn` list, but I removed it because sometimes you might want to keep the `mcompose` operations instead of unfolding them.

cava/Cava/Semantics/Combinational.v
```diff
@@ -38,6 +44,14 @@ Definition xnorb b1 b2 : bool := negb (xorb b1 b2).
 (* interpretation.                                                            *)
 (******************************************************************************)
 
+Definition port_signal p := combType (port_type p).
```
jadephilipoom, line 47: I think it would be clearer to just write `port_signal combType` everywhere `port_signal` is used, rather than having two different definitions called `port_signal`.

jadephilipoom (approved):
A few comments, but overall seems okay to me. It would be nice if we could get these parts of the interface to be simpler still.

cava/Cava/Core/CavaClass.v
```diff
@@ -15,14 +15,19 @@
...
 
 Require Import ExtLib.Structures.Monad.
+Require Import ExtLib.Structures.Functor.
 Require Import Cava.Core.Netlist.
 Require Import Cava.Core.Signal.
+Require Import Cava.Util.Tuple.
 
 Local Open Scope type_scope.
 
+Import FunctorNotation.
```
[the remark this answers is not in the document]
blaxill: Removed

cava/Cava/Core/Netlist.v
```diff
@@ -326,111 +333,81 @@ Definition initState : CavaState
...
-  o <- circuit i ;;
-  let outType := circuitOutputs intf in
-  instantiateOutputPorts outType o.
-
-Definition makeNetlist (intf : CircuitInterface)
-                        (circuit : tupleNetInterface (circuitInputs intf) ->
-                                    state CavaState (tupleNetInterface (circuitOutputs intf))) : CavaState
+Check makeNetlist.
+
+                        (* makeNetlist intf (curry _ _ (curry_helper _ _ (interpCircuit c))). *)
```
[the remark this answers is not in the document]
blaxill: Accidental inclusion- removed

cava/Cava/Semantics/Combinational.v
blaxill: Done

PR #758, opened 2021-04-21
[investigations] Cleanup/reorganization of bedrock2 AES code

jadephilipoom:
The AES experimental files were getting a bit hard to work with because they were too long, so I've done a bit of a refactor to make my life easier ahead of trying more proofs.
- Move constant definitions into their own file, since they're very verbose
- Move generic word-related lemmas and tactics into their own file
- Make a `StateMachineSemantics` file that defines bedrock2 semantics in terms of a state machine with read/write steps, and define AES semantics in terms of these


blaxill (approved):
Lgtm! Also can start thinking about moving out of investigations as you see fit, maybe top level `coverification/software` or something

jadephilipoom:
> Lgtm! Also can start thinking about moving out of investigations as you see fit, maybe top level `coverification/software` or something

I might want to do a little more performance engineering first -- some of the proofs here are a little slow. And at the moment I'm not 100% committed to bedrock2, so probably will hold off until I've done some more experimentation.

PR #759, opened 2021-04-21
[investigations] Implement and prove aes_iv_put, implement aes_key_put

jadephilipoom:
- Implementation in bedrock2 of two functions from `aes.c`, which write the key and initialization-vector registers
- Proof that `aes_iv_put`, the simpler of the two, obeys a specification in terms of the AES state machine
- `read_step` and `write_step` are now postcondition-style `Prop`s instead of inductives, which makes the case analysis less painful
- Lemma that allows you to unfold while loops if you know exactly how many iterations they will take, and prove them as if they are unrolled (this is nicer than loop invariants for `aes_iv_put`)
- Automation improvements
  - `push_unsigned` tactic to simplify `word`-arithmetic goals
  - `ring_simplify_load_addr`, which runs `ring_simplify` on memory addresses so you don't have to prove that `ptr + 4 + 4 = ptr + (2 * 4)`


blaxill (approved):
Lgtm!

PR #763, opened 2021-04-22
[investigations] Prove aes_key_put

jadephilipoom:
This proof is a little more complicated than `aes_iv_put` because we have to handle three different cases of key length and then do two while loops with different numbers of iterations depending on the case. I've added a little more automation for this but opted to keep short, repetitive lines of code for each iteration of each loop in each case -- 24 in all. (I could have made a tactic to remove the repetition, but that makes it harder to debug each case.) Ultimately, this should be replaced by a loop-invariant proof to make it faster and less verbose, but this works for a first pass.


blaxill (approved):
> (I could have made a tactic to remove the repetition, but that makes it harder to debug each case.) Ultimately, this should be replaced by a loop-invariant proof to make it faster and less verbose, but this works for a first pass.

Yeah fair enough, LGTM!

PR #764, opened 2021-04-23
[investigations] Prove status-flag checking functions and separate out some logic

jadephilipoom:
- Factor out proofs that have to do with state machines in general, or while loops in general, into `StateMachineProperties.v` and `WhileProperties.v` respectively, to declutter `AesProperties.v`
- Fix a mistake in `is_flag_set` definition that meant the boolean was reversed
- Implement and prove `aes_idle`, `aes_data_ready`, and `aes_data_valid` functions from `aes.c`


investigations/bedrock2/StateMachineProperties.v
```diff
@@ -0,0 +1,76 @@
...
+  let t := lazymatch goal with
+           | |- cmd _ (cmd.interact _ _ _) ?t _ _ _ => t end in
+  eapply interact_mmio; [ tac | ]; intros;
+  lazymatch goal with
+  | H : execution (_ :: t) _ |- _ =>
+    pose proof H; apply invert1_execution in H; destruct H as [? [? ?]]
+  end; invert_step.
+
+(* Remove [execution] hypotheses that are superceded by later ones; improves
+   proof performance *)
```
blaxill, line 70: I'm curious, does this affect batch mode performance as well or just interactive proof mode performance? (I've remember situations previously where large goals slowed down interactive mode but not batch mode so much)

blaxill (approved):
Lgtm! Out of interest, did you catch the flipped is_flag_set by eye or failing proof goal?

investigations/bedrock2/StateMachineProperties.v
jadephilipoom: I didn't check, but that seems plausible.

jadephilipoom:
> Lgtm! Out of interest, did you catch the flipped is_flag_set by eye or failing proof goal?

Failing proof goal. It's difficult to prove things that are false!

PR #765, opened 2021-04-23
[investigations] Implement + prove aes_data_put and aes_data_put_wait

jadephilipoom:
These functions write input data to the AES block and (in my simplified version of the state machine) trigger the transition from IDLE to BUSY states.


blaxill (approved):
Lgtm!

PR #766, opened 2021-04-26
[investigations] Implement + prove aes_data_get and aes_data_get_wait

jadephilipoom:
Very similar to #765, but for reading output data instead of writing input data.

With this, we have a more-or-less complete version of OpenTitan's `aes.c` :tada: (we're missing `aes_clear`, but the simplified AES state machine I'm using for these proofs doesn't have a clearing functionality).

The next steps I'm thinking of taking from here are:
- Plug this code into `aes.c` and check that we pass the AES tests
- Prove the admitted bit-arithmetic lemmas in `Word.v`
- Change the `aes_key_put` proof, which takes quite a while to check because it destructs into 3 cases of key length * 8 loop iterations per case, into a proof using a loop invariant
- Add a test program that composes operations from the `aes.c` API (e.g. initialize, write IV and key registers, write input, wait for output, read output) and prove it correct, to ensure that the function specifications compose properly (i.e. that the postconditions of one operation will fulfill the preconditions of the next)
- Try to compile to RISC-V using the bedrock2 compiler, maybe see if we can run the AES tests on the assembly code
- Modify the state machine model to be less simplified (for instance, allow the AES unit to take new input while it is still processing the last input)

### DIFF
--- a/investigations/bedrock2/Aes/Aes.v
+++ b/investigations/bedrock2/Aes/Aes.v
@@ -151,6 +151,28 @@ Section Impl.
     ))).
 
   (**** aes.c
+    void aes_data_get(void *data) {
+      // Read the four output data registers.
+      for (int i = 0; i < AES_NUM_REGS_DATA; ++i) {
+        ((uint32_t * )data)[i] = REG32(AES_DATA_OUT0(0) + i * sizeof(uint32_t));
+      }
+    }
+   ***)
+  Definition aes_data_get : func :=
+    let data := "data" in
+    let val := "val" in
+    let i := "i" in
+    ("b2_data_get",
+     (aes_globals ++ [data], [], bedrock_func_body:(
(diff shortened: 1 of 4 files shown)


blaxill (approved):
Lgtm!

> With this, we have a more-or-less complete version of OpenTitan's aes.c 🎉 (we're missing aes_clear, but the simplified AES state machine I'm using for these proofs doesn't have a clearing functionality).

Awesome job!

satnam6502:
Looks like `aes.c` is crushed and we should move onto more challenging examples?
Well done!

jadephilipoom:
> Looks like `aes.c` is crushed and we should move onto more challenging examples?

I'd like to do some of those "next step" items to wrap up -- especially actually compiling this code to RISC-V and seeing if we can get it to pass the AES tests. But broadly, yes!

PR #767, opened 2021-04-27
Add AES example program

jadephilipoom:
This program serves as sort of a test that the AES API proofs compose together properly. The example program initializes the AES block, writes the key, IV, and input registers, waits for the output, and reads the results into an array pointer provided by the caller. The specification says that the circuit is idle, the input arrays (keys, IV, and input) are unchanged, and the result array pointer now points to values that match the AES specification for the inputs (the specification in terms of words is provided by the `AesSemantics.parameters` instance in terms of words; we could plug in Cava's AES specification by making an instance of the parameters for it, but I haven't done so here because I think it's nicer for the proofs to be parameterized over how one chooses to express the spec).

Found a fun copy-paste typo where I accidentally made the name of `aes_key_put` and `aes_iv_put` the exact same, causing the wrong function specification to get pulled in as a precondition! That one took some figuring out.

### DIFF
--- a/investigations/bedrock2/Aes/AesExampleProperties.v
+++ b/investigations/bedrock2/Aes/AesExampleProperties.v
@@ -1,0 +1,191 @@
+Require Import Coq.Strings.String.
+Require Import Coq.Lists.List.
+Require Import Coq.micromega.Lia.
+Require Import Coq.ZArith.ZArith.
+Require Import bedrock2.Array.
+Require Import bedrock2.Map.Separation.
+Require Import bedrock2.Map.SeparationLogic.
+Require Import bedrock2.ProgramLogic.
+Require Import bedrock2.Scalars.
+Require Import bedrock2.Semantics.
+Require Import bedrock2.Syntax.
+Require Import bedrock2.WeakestPrecondition.
+Require Import bedrock2.WeakestPreconditionProperties.
+Require Import coqutil.Word.Interface.
+Require Import coqutil.Word.Properties.
+Require Import coqutil.Map.Interface.
+Require Import coqutil.Map.Properties.
+Require Import coqutil.Tactics.Tactics.
+Require Import Cava.Util.List.
+Require Import Cava.Util.Tactics.
+Require Import Bedrock2Experiments.StateMachineSemantics.
+Require Import Bedrock2Experiments.StateMachineProperties.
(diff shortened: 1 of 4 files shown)


investigations/bedrock2/Aes/AesExampleProperties.v
```diff
@@ -0,0 +1,191 @@
...
+                   * array scalar32 (word.of_Z 4) iv_ptr iv_arr
+                   * array scalar32 (word.of_Z 4) key_ptr key_arr
+                   * array scalar32 (word.of_Z 4) ciphertext_ptr
+                           [out0; out1; out2; out3] * R)%sep m'
+                (* ...and there are no return values *)
+                /\ rets = []).
+
+
+  (* TODO: the nregs_populated preconditions are annoying; replace with
+       reg_lookup = Some everywhere? *)
```
blaxill, line 91: What causes this, is it the assumption that registers might be initially undefined?

blaxill (approved):
Lgtm!

> Found a fun copy-paste typo where I accidentally made the name of aes_key_put and aes_iv_put the exact same, causing the wrong function specification to get pulled in as a precondition! That one took some figuring out.

Argh!

investigations/bedrock2/Aes/AesExampleProperties.v
jadephilipoom: Since I left the map implementation abstract, `nregs_populated` (which is stated in terms of a fold over the map) can't be decided by computation; I have to apply lemmas that transform e.g. `nregs_populated (map.put m (reg_addr r) v) KeyReg` into `if reg_category_eqb (reg_addr r) KeyReg then S (nregs_populated m KeyReg) else nregs_populated m KeyReg`. There are a lot of `map.put`s, so it's a bit slow and cumbersome. The main possible fixes I see are a) the fix suggested in the comment; replace `nregs_populated` preconditions with several preconditions like `reg_lookup KEY0 = Some v` or `reg_lookup KEY0 = None` or b) not make the map implementation abstract (pros: bedrock2 automation seems to assume it's not abstract, so it would remove some pain points there, plus we'd get nice proofs by computation; cons: could potentially make proof performance tricky if you have to do complex dances to avoid letting Coq unfold a big map, plus we couldn't freely swap out map implementations).

PR #769, opened 2021-04-28
Koika J1 experimentation 

blaxill:
A first non-trivial Koika experiment- a simple stack based CPU based on James Bowman's open source J1. I did not immediately model our AES component as the interaction between Cava and Koika is non-trivial partly due to Koika's external   functions expecting combinational only external functions. Although not an issue for our AES sub components, the interface exposed externally (e.g. register interface for top level AES) must also currently use Koika external functions. As this interface is clearly not combinational, special care has to be made to when using them as Koika may optimise them away.

### DIFF
--- a/investigations/koika/AesCore.v
+++ b/investigations/koika/AesCore.v
@@ -1,0 +1,313 @@
+Require Import Koika.Frontend.
+Require Import AesCore.TileLink.
+
+Definition key_t := bits_t 128.
+Definition state_t := bits_t 128.
+Definition keypair_t := bits_t 256.
+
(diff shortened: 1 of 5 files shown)


jadephilipoom:
> the interaction between Cava and Koika is non-trivial partly due to Koika's external functions expecting combinational only external functions

I'm not sure what _exactly_ this interface looks like, but I think what we could possibly do here is use the `step` function. `step c` provides the combinational function for circuit c from its input and previous state to its output and next state -- an automatic mealy machine conversion, in other words.

PR #770, opened 2021-04-29
Test bedrock2 functions in OpenTitan

jadephilipoom:
- Modifies `AesToC.v` to strip out global-variable arguments and print out a modified version of OpenTitan's `aes.c` in which all the functions except `aes_clear` are wrappers for the bedrock2-generated functions
- Adds a `make` target for creating the new `aes.c`
- Adds a script (based on @blaxill 's preexisting script for the AES SystemVerilog files) that calls the make target, copies the new `aes.c` into OpenTitan, builds everything needed, runs the test, and prints the output

Tests passed on the first try! I even had to break something intentionally just to make sure it was actually picking up the right file :slightly_smiling_face:


investigations/bedrock2/Aes/AesToC.v
```diff
@@ -11,5 +14,192 @@ Local Open Scope Z_scope.
 
 Existing Instance constant_names.
 
-Definition funcs := [aes_init].
-Redirect "aes.c" Compute c_module funcs.
+(* Redefinition from bedrock2's ToCString; the only difference is that globals
+   are stripped out of the arguments list in function calls *)
```
blaxill, line 18: Is this because you want the functions to use the globals from the OpenTitan files?

blaxill (approved):
> Tests passed on the first try! I even had to break something intentionally just to make sure it was actually picking up the right file 🙂

Great! :)

investigations/bedrock2/Aes/AesToC.v
jadephilipoom: Yes. There's no support for global values in bedrock2, so I had to pretend that every function is getting all the globals passed in as arguments. I could have just made the wrappers manually pass in the globals instead of stripping them out here, but that would have created a lot of boilerplate and introduced significant room for typos or copy/paste errors, so this solution seemed more robust.

PR #774, opened 2021-04-30
Add tactic reference and fix some typos

jadephilipoom:
This has been on my to-do list for a while. I've noticed that it's hard to discover some of the proof automation in the repo, so I've added information to the reference that lists and explains useful tactics and autorewrite databases. I also saw a couple of typos or awkwardly-reading lines, so I fixed those while I was there.


blaxill (approved):
Nice! I wonder if we should separate some of these tactics so you can share them with the community as a self contained package

jadephilipoom:
> Nice! I wonder if we should separate some of these tactics so you can share them with the community as a self contained package

Before making my own, I think I'd upstream them to `coqutil`, since that's pretty much the purpose of `coqutil` anyway! We could also upstream half of the lemmas in `Util` if we wanted.

PR #777, opened 2021-05-04
Remove comment

benlaurie:
a) Not very informative
b) # is not a comment introducer in zsh so this caused me some confusion when I cut'n'pasted...

### DIFF
--- a/demos/tutorial.v
+++ b/demos/tutorial.v
@@ -56,7 +56,7 @@ quick-start instructions for installing Cava::
 
   $ git clone https://github.com/project-oak/silveroak.git
   $ cd silveroak
-  $ git submodule update --init --recursive # initialize submodules
+  $ git submodule update --init --recursive
   $ make -j4 cava-coq
 
 You can now make the Cava libraries visible to your project by adding the


satnam6502:
> a) Not very informative
> b) # is not a comment introducer in zsh so this caused me some confusion when I cut'n'pasted...

I thought that was quite an informative comment. And this is not a `zsh` script.

satnam6502:
> > a) Not very informative
> > b) # is not a comment introducer in zsh so this caused me some confusion when I cut'n'pasted...
> 
> I thought that was quite an informative comment. And this is not a `zsh` script.

Although I agree, for copying and pasting it can be troublesome. Why not have the tutorial user invoke a rule in the top `Makefile` to just build the core?

jadephilipoom:
> Although I agree, for copying and pasting it can be troublesome. Why not have the tutorial user invoke a rule in the top `Makefile` to just build the core?

My reasoning for submodule initialization as a separate line was this:
- The `third_party` makefile used to update/initialize submodules, but doesn't any more as of #691 -- I was surprised to notice this change when testing. However, I think it makes sense not to automatically update/initialize submodules, because that setup made it a huge pain to update the submodules (`make` would reset the update -- that's why we have the now-redundant `no-update` target). Regardless of what the best answer is here, I think it's evident enough that the behavior of that Makefile changes frequently, so I'd rather not make the tutorial assume anything about its behavior. If I don't assume the makefile will initialize the submodules, then the user has to. One option would be to have `make submodule-init` which just initializes the submodules, but I think that's no more readable than `git submodule`.
- I could have instead suggested people clone with the `--recursive` option, which initializes the submodules as part of the clone, but then some people might have cloned already and not know how to recover, and/or not notice the `--recursive`.

blaxill:
Closing this in favour of #778 which adds a top level command and replaces the command referenced here with the new one in the linked PR

PR #779, opened 2021-05-07
Fix a few tiny things in the tutorial

fshaked:


demos/tutorial.v
```diff
@@ -54,7 +54,7 @@ idea of how Cava works, skip to the next section. If you are trying to write
 your own circuits or step through this tutorial locally, here are the
 quick-start instructions for installing Cava::
 
-  $ git clone https://github.com/project-oak/silveroak.git
+  $ git clone --recurse-submodules https://github.com/project-oak/silveroak.git
```
jadephilipoom, line 57: Did the build fail without this? The `make update-third-party` line should initialize the submodules, if it's not doing so then something is wrong.

jadephilipoom (approved):
Thanks!

demos/tutorial.v
fshaked: My build failed, but it probably had nothing to do with that. The main reason the build failed was the version of coq I used. This was one of the things I tried before pinning the coq version. 

I just checking out a clean copy and build it (without --recurse-submodules) and it didn't fail, so you can discard this fix.

fshaked: Removed

PR #781, opened 2021-05-12
State monad AST representation of circuits

jadephilipoom:
Experimental alternative representation for Cava's core logic that removes the combinational/sequential separation.


cava/Cava/Lib/Vec.v
```diff
@@ -37,9 +40,11 @@ Section WithCava.
...
     constantV (Vector.map constant v).
 
-  Definition map_literal {A B n} (f : A -> cava (signal B)) (v : Vector.t A n)
-    : cava (signal (Vec B n)) :=
-    out <- mapT f v ;;
+  (* TODO(blaxill): Make Cava.Util.IxTraversable to gather states ? *)
+  Definition map_literal {A B n st} (f : A -> cava st (signal B)) (v : Vector.t A n)
+    : cava (concat (repeat st n)) (signal (Vec B n)) :=
+    (* : cava st (signal (Vec B n)) := *)
+    out <- Cava.Util.IxTraversable.mapT f v ;;
```
blaxill, line 47: @jadephilipoom the traversable instances here might get a bit hairy, `cava (concat (repeat st n))` is the "shape" of state for a `map f` but in general an indexed Traversable will have a state shape specific to an instance. Maybe this is fine.

blaxill:
I'm going to replace all the definitions with this new `Cava` class, but only using the deep embedded binders for the simulation as changing the other `Cava` instances to use the deep embedding should not be too hard to do later

cava/Cava/Lib/Vec.v
jadephilipoom: I think this might be a sign that we're going in the wrong direction here. It just seems like we're needing more and more complicated things to make this solution work.

blaxill: to be fair it would be the equivalent of the existing traversable, but I agree this makes me feel a little uneasy

blaxill: Getting rid of the explicitly typed state and using existentials in the new inductive type would allow us to return to regular monads

jadephilipoom: Maybe we should continue pursuing this approach, but in parallel also try to remove the explicitly typed state and/or try rewriting the netlist stuff so it's working directly on the AST? It's an advantage to let the netlist code work off of a monad abstraction instead of an AST abstraction, but it's looking to me like that advantage might not be worth the complexity it costs. What do you think?

blaxill: Thinking about it, making the state existential is probably best as nobody is ever going to want to write a complex type signature for the state, and although it will always be inferable it is then pointless to carry around?

blaxill: > Maybe we should continue pursuing this approach, but in parallel also try to remove the explicitly typed state and/or try rewriting the netlist stuff so it's working directly on the AST? 

Yeah how about here remove the explicitly typed state but carry on with new inductive structure (but state becomes internal to it)

> It's an advantage to let the netlist code work off of a monad abstraction instead of an AST abstraction, but it's looking to me like that advantage might not be worth the complexity it costs. What do you think?

Actually the netlist implementation here is straight forward, its the vec util functions (and any other library functions) that require modification, but I agree; cons out way the benefits of the explicit state here

jadephilipoom: > Yeah, how about here remove the explicitly typed state but carry on with new inductive structure (state becomes internal to it)

Sounds good if we can make it work, but I remember having problems when I tried this earlier. Essentially, as soon as you're using Gallina binds, you can't recurse on the circuit to get the state type (without going full PHOAS and using a dummy `var` type). So you can't determine the state of the circuit unless it's obvious from the type.

blaxill: Thats a good point, what about just a deep embedding of the operations then

jadephilipoom: > Thats a good point, what about just a deep embedding of the operations then

Deep-embedding the operations doesn't fix the problem, I think, unless you use PHOAS; you still have binders that you can't recurse under. Maybe PHOAS is the right thing to do, but I'm a little hesitant because it does also introduce a fair amount of complexity. Right now, if I had unlimited time I would experiment with:
- Netlist/library functions operating on the AST instead of normal monads
- A true deep embedding in PHOAS
- Going back to the drawing board and doing something totally different

I'm not going to take the lock on any of those right now, because I'm currently working on other things and don't expect to get back to this for a couple of days. So feel free to claim any of those or pursue some other avenue! I think the best thing we can do is experiment as much as possible and see what works.

blaxill: By a deep embedding of just the operations I mean the is same as your AST. The monad interface can still be used to construct the functions, but the all the bindings get effectively unfolded and it removes the sharing.

jadephilipoom: Wouldn't that cause problems when we want to be aware of sharing later on?

blaxill: I don't know a solution that works for all cases- this would allow a monad without separate circuit type, and give an inductive structure with state representation for evaluation, and the netlist interpretation would still have sharing. This version wouldn't have an inductive structure for the netlist interpretation and so doesn't attempt to help with compiler proofs.

jadephilipoom: Ah, I didn't understand that the netlist interpretation would still have sharing. In that case it sounds like a good idea. If I'm understanding correctly, you're saying that the netlist interpretation would still use a monad-based interpretation, but the proofs would use an AST with no binds?

jadephilipoom:
We've long since converged on a new core representation, so I'll close this experiment.

Issue #784, opened 2021-05-14
Cava2HDL does not appear to be installed anymore


satnam6502:
I can't load up generated Haskell files or compile them with `ghc --make`.

```
~/silveroak/examples$ ghci AdderTree.hs
GHCi, version 8.10.3: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /usr/local/google/home/satnam/.ghci
[1 of 1] Compiling AdderTree        ( AdderTree.hs, interpreted )

AdderTree.hs:7:1: error:
    Could not find module ‘Adders’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
  |
7 | import qualified Adders
  | ^^^^^^^^^^^^^^^^^^^^^^^

AdderTree.hs:8:1: error:
    Could not find module ‘BinNat’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
  |
8 | import qualified BinNat
  | ^^^^^^^^^^^^^^^^^^^^^^^
```

satnam6502:
And attempting to install it does not seem to work.
```
$ cabal install Cava2HDL
Warning: The package list for 'hackage.haskell.org' is 483 days old.
Run 'cabal update' to get the latest list of available packages.
Wrote tarball sdist to
/usr/local/google/home/satnam/silveroak/dist-newstyle/sdist/Cava2HDL-0.1.0.0.tar.gz
Wrote tarball sdist to
/usr/local/google/home/satnam/silveroak/dist-newstyle/sdist/cava-examples-0.1.0.0.tar.gz
cabal: Error: Could not find module: NandLUT with any suffix:
["gc","chs","hsc","x","y","ly","cpphs","hs","lhs","hsig","lhsig"]. If the
module is autogenerated it should be added to 'autogen-modules'.
```

blaxill:
`cabal repl` is the cabal equivalent for getting `ghci` up, that is working for me.

blaxill:
You should never need to install any haskell packages with the cabal setup, its automatic

PR #785, opened 2021-05-15
Generate HTML for sum8 ILA capture

satnam6502:
Add documentation to show the implementation of the sum of sequence circuit, its simulation and its execution on an FPGA board. Also fix the generation of the Cava2HDL library #784


demos/tutorial.v
```diff
@@ -970,7 +969,7 @@ run and observe this actually running on an FPGA and capture its output:
    :width: 70%
    :alt: Logic analyzer trace capture for the sum8 circuit.
 
-Reassuring the actual circuit behaves as predicted by the Cava model
+Reassuringly the actual circuit behaves as predicted by the Cava model
```
jadephilipoom, line 972: nit:
```suggestion
Reassuringly, the actual circuit behaves as predicted by the Cava model
```

jadephilipoom (approved):
LGTM. I left some comments on #718, but fine to merge as-is.

cava/Cava2HDL/Makefile
```diff
@@ -20,6 +20,9 @@ all:		haskell
 
 haskell:
 		cabal build
+		runhaskell Setup.hs configure --user
+		runhaskell Setup.hs build
+		runhaskell Setup.hs install
```
blaxill, line 25: I don't think Setup.hs is required? Were you seeing issues without it?

blaxill (changes requested):

cava/Cava2HDL/Makefile
blaxill: With the cabal setup, installing is not require. `cabal repl` should work directly to get a `ghci` prompt with the files loaded.

satnam6502: Well, I don't want to use `cabal repl`! OK... I am going to do your bidding and remove these lines and use `cabal repl` :-(

satnam6502:
@blaxill PTAL, thank you kindly.

blaxill (approved):
Lgtm, but the `Setup.hs` can be dropped too now I think.

satnam6502:
> Lgtm, but the `Setup.hs` can be dropped too now I think.

Ah! Right! Done! Thanks.

PR #786, opened 2021-05-17
Sketch of desired Cava API for loop and delay

satnam6502:
WIP sketch for the API I think we need to have for loops and delay in Cava for a viable release candidate.


satnam6502:
@jadephilipoom @blaxill why do I have to fully qualify `ExtLib.Structures.MonadState.put` and `ExtLib.Structures.MonadState.get` here? I get an error if I don't fully qualify them. This seems odd because the import statement is there... and that's exactly how I import them from `MonadState` in the main Cava code. Thank you.

jadephilipoom:
> @jadephilipoom @blaxill why do I have to fully qualify `ExtLib.Structures.MonadState.put` and `ExtLib.Structures.MonadState.get` here? I get an error if I don't fully qualify them. This seems odd because the import statement is there... and that's exactly how I import them from `MonadState` in the main Cava code. Thank you.

Sounds like something else in context is called `put`/`get`. What is the output of `Locate put`?

blaxill:
> @jadephilipoom @blaxill why do I have to fully qualify `ExtLib.Structures.MonadState.put` and `ExtLib.Structures.MonadState.get` here? I get an error if I don't fully qualify them. This seems odd because the import statement is there... and that's exactly how I import them from `MonadState` in the main Cava code. Thank you.

You're importing the concrete instance from __`Data.Monads.StateMonad`__ but, as you note, the class and its methods are defined in __`Structures.MonadState`__, if you add `Require Import ExtLib.Structures.MonadState.` they will work without qualification.

satnam6502:
> > @jadephilipoom @blaxill why do I have to fully qualify `ExtLib.Structures.MonadState.put` and `ExtLib.Structures.MonadState.get` here? I get an error if I don't fully qualify them. This seems odd because the import statement is there... and that's exactly how I import them from `MonadState` in the main Cava code. Thank you.
> 
> You're importing the concrete instance from **`Data.Monads.StateMonad`** but, as you note, the class and its methods are defined in **`Structures.MonadState`**, if you add `Require Import ExtLib.Structures.MonadState.` they will work without qualification.

Thank you! Fixed.

PR #788, opened 2021-05-17
Add a circuit that computes the square of a bitvector

jadephilipoom:
I needed this for the demo I'm developing, and it seems like a useful thing to have in the library anyway.


satnam6502:
Ah, specifically, this is for computing a square of an unsigned value without using multiplication.

satnam6502 (approved):
Looks cool!

PR #791, opened 2021-05-18
Update bedrock2

jadephilipoom:
Bump bedrock2 (and its dependencies) up to the latest version, in order to capture some upstream changes that we submitted:
https://github.com/mit-plv/bedrock2/pull/187
https://github.com/mit-plv/riscv-coq/pull/27

The gist is that those two PRs add the ability for the bedrock2 compiler to handle code that uses _hardware_ guarantees to prove things about program logic (namely loop termination), which is something we need for the firmware proofs. The AES code loops until it sees a "done" status flag; without the guarantee from hardware that the status flag will eventually appear, these would be infinite loops.


satnam6502 (approved):
Stamped.

blaxill:
Rerunning CI because GH actions failed without reason..

PR #792, opened 2021-05-18
Compile IncrementWait firmware

jadephilipoom:
Using the recent bedrock2 updates, compile the minimal "increment-wait" example firmware all the way to RISC-V and prove that the setup satisfies the bedrock2 compiler's preconditions. This proof of concept should pave the way to doing the same for AES; it's designed to be a minimal example of the AES signalling structure.

- Adds the bedrock2 compiler (which is separate from the language) as a dependency
- Changes our `coq-record-update` dependency to go via bedrock2 (the bedrock2 compiler also uses it and otherwise there are issues with having the same submodule twice -- open to better fixes here)
- Changes the constant-handling for IncrementWait to match AES
- Changes the hardware spec for IncrementWait to provide the guarantees that the compiler needs (the hardware spec `ext_spec` simultaneously provides the _requirements_ for software and the _givens_ for the compiler -- I didn't quite understand this interface before)
- Proves `FlatToRiscvCommon.compiles_FlatToRiscv_correctly` for interactions based on this hardware spec, which states that the bedrock2 compiler will correctly compile MMIO interactions
- Compiles `put_wait_get`, which writes a value to an register, repeatedly reads a status register until it sees a "done" status flag, and then reads the value from the register and returns it


investigations/bedrock2/IncrementWait/IncrementWaitToRiscV.v
```diff
@@ -0,0 +1,135 @@
...
+     lw      x12, x2, 24
+     lw      x6, x2, 28    // restore values of argument registers
+     lw      x7, x2, 32
+     lw      x8, x2, 36
+     lw      x9, x2, 40
+     lw      x10, x2, 44
+     lw      x11, x2, 48
+     lw      x1, x2, 52    // load ra
+     addi    x2, x2, 84    // increase stack pointer
+     jalr    x0, x1, 0     // return
```
jadephilipoom, line 133: The punchline: assembly for `put_wait_get`!

investigations/bedrock2/IncrementWait/IncrementWaitProperties.v
```diff
@@ -139,49 +136,91 @@ Section Proofs.
...
              clear H2; one_goal_or_solved ltac:(try congruence)
            | H : BUSY _ _ = BUSY _ _ |- _ => invert_nobranch H
            | H : ?x = ?x |- _ => clear H
            end.
 
-  Hint Rewrite word.unsigned_add word.unsigned_sub word.unsigned_mul
-       word.unsigned_mulhuu word.unsigned_divu word.unsigned_and
-       word.unsigned_or word.unsigned_xor word.unsigned_sru word.unsigned_slu
-       word.unsigned_ltu @word.unsigned_of_Z_0 @word.unsigned_of_Z_1
-       using solve [typeclasses eauto] : push_unsigned.
```
blaxill, line 180: Was this not helpful? Perhaps autorewriting wasn't reducing to the equivalent forms you needed?

investigations/bedrock2/IncrementWait/IncrementWaitToRiscV.v
blaxill: Nice!

blaxill (approved):
Lgtm! Notes are optional

investigations/bedrock2/IncrementWait/IncrementWaitMMIO.v
```diff
@@ -0,0 +1,663 @@
...
+Proof.
+  intros. unfold compile_ext_call.
+  destruct (String.eqb _ _); destruct binds; try destruct binds;
+  try destruct args; try destruct args; try destruct args;
+  cbv; intros; discriminate.
+Qed.
+
+Lemma compile_ext_call_length': forall binds f args,
+    Z.of_nat (List.length (compile_ext_call binds f args)) <= 7.
+Proof. intros. rewrite compile_ext_call_length. blia. Qed.
```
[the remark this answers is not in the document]
jadephilipoom: I think they might be needed later, potentially. This file is based on https://github.com/mit-plv/bedrock2/blob/master/compiler/src/compilerExamples/MMIO.v, which includes those lemmas. I've proved that interactions are compiled correctly, but not yet instantiated the compiler correctness proof to say _this particular program_ is correctly compiled, so it's possible I'll need them there -- otherwise I'll remove.

investigations/bedrock2/IncrementWait/IncrementWaitProperties.v
jadephilipoom: It was helpful, but these hints are now in `Word.v` (added since I last worked on the minimal example) so there was no need to duplicate them.

investigations/bedrock2/IncrementWait/IncrementWaitMMIO.v
```diff
@@ -0,0 +1,663 @@
...
+    pose proof addrs_aligned as Halign.
+    rewrite Forall_forall in Halign.
+    apply Halign, reg_addrs_complete.
+  Qed.
+
+  Lemma compile_ext_call_correct: forall resvars extcall argvars,
+      FlatToRiscvCommon.compiles_FlatToRiscv_correctly
+        (@FlatToRiscvDef.compile_ext_call compilation_params)
+        (FlatImp.SInteract resvars extcall argvars).
+  Proof.
```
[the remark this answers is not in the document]
jadephilipoom: Possibly, but it's a lightly edited version of the proof in the bedrock2 repo, and I haven't gone through the logic in enough detail to know. I think for now there's some advantage in sticking close to what they've got, so that we can reference their code when upstream changes.

PR #795, opened 2021-05-20
Instantiate compiler correctness proof for increment-wait

jadephilipoom:
This PR builds on #792 to write a fully-instantiated correctness proof for the assembly code in the increment-wait minimal example, in order to make sure that everything ties together and all of the compiler proofs' preconditions are satisfied.
- Creates a `main` function that simply reads an input value from a constant location in memory, calls `put_wait_get`, and stores the output to another constant location in memory (the `main` function is not allowed to have arguments)
- Compiles the `main` function and the `put_wait_get` function together
- Proves that, if the input/output pointers are valid, a RISC-V machine executing the resulting assembly will remain in a valid state and satisfy the postcondition of `main` (which says that the circuit has returned to the IDLE state, the value at the input location has stayed the same, and the value at the output location in memory is now the result of incrementing the input value).

### DIFF
--- a/investigations/bedrock2/IncrementWait/Constants.v
+++ b/investigations/bedrock2/IncrementWait/Constants.v
@@ -35,6 +35,18 @@ Definition constant_vars
      STATUS_DONE := expr.var STATUS_DONE;
   |}.
 
+(* Given the Z values of all the constants, convert them to exprs with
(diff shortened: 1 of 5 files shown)


blaxill (approved):
Lgtm!

PR #797, opened 2021-05-20
Add demos/ to CI and update to use cabal

blaxill:


demos/cava-tutorial.cabal
```diff
@@ -0,0 +1,15 @@
+cabal-version:       >=1.10
+
+name:                cava-tutorial
+version:             0.1.0.0
+author:              The SilverOak Team
+maintainer:          benblaxill@google.com
+build-type:          Simple
+
+executable examples
+  main-is:             TutorialSV.hs
```
satnam6502, line 10: Is it really worth caballing since source file Haskel programs?

satnam6502 (approved):

demos/cava-tutorial.cabal
blaxill: Technically its two files :) But the main reason is library discovery. The current Cabal usage design (this wasn't always the case) is to not install packages globally, but that does then require that files have a cabal file to discover dependencies.

PR #800, opened 2021-05-24
Update AES semantics to be compatible with compiler proofs

jadephilipoom:
This PR contains some changes to the AES semantics and proofs that match the changes I made earlier to the increment-wait example circuit (see e.g. #792) so that it would be compatible with the bedrock2 compiler proofs. (The `ext_spec` definition forms part of the interface between program and compiler proofs; the previous version worked for proofs of programs but wouldn't have been strong enough to work with the compiler proofs.) This paves the way for actually compiling our AES code and getting proofs about the assembly.

Some of the code churn here is a change to how the state data (e.g. register values) is represented; since I was changing all these proofs anyway, it seemed worth a refactor because this representation was a major pain point before. Still not necessarily totally happy with the new solution, but it's at least much less painful than the old one (a lot of code got deleted from `AesProperties`).

### DIFF
--- a/investigations/bedrock2/Aes/Aes.v
+++ b/investigations/bedrock2/Aes/Aes.v
@@ -6,6 +6,7 @@ Require Import bedrock2.Syntax.
 Require Import bedrock2.NotationsCustomEntry.
 Require Import bedrock2.ToCString.
 Require Import coqutil.Word.Interface.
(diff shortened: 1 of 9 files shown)


satnam6502 (approved):
I've read the PR but the contents are beyond my competency to give a detailed review, although I am very happy to stamp it!

PR #801, opened 2021-05-25
State machine mmio

jadephilipoom:
- Adds `StateMachineMMIO.v`, which constructs the compiler setup for the generic state-machine semantics used by AES and proves a key lemma that says the compiler compiles MMIO interactions correctly. This file is almost exactly a copy of `IncrementWaitMMIO.v`, which is itself based on bedrock2's `compilerExamples/MMIO.v`.
- Threads a few compiler preconditions (e.g. MMIO addresses are word-aligned) through the state machine semantics and AES proofs.
- Fixes the `aes.c` template to use "MMIOREAD" and "MMIOWRITE" (should have been in #800, missed this in the previous changes)

The fact that this proof passes means that the firmware/hardware interface is set up in a way that's compatible with the compiler (i.e. #800 worked)!


satnam6502 (approved):
Stamped.

PR #802, opened 2021-05-25
Compile AES code

jadephilipoom:
Compile the AES firmware to assembly! :tada: 

Unfortunately, the bedrock2 compiler has a restriction on the number of arguments a function can have (arguments + return values + 5 < 32), which meant the previous strategy of passing all global constants into all functions would result in too many arguments. This PR therefore modifies each function to take only the globals it actually needs as arguments.

### DIFF
--- a/investigations/bedrock2/Aes/AesToRiscV.v
+++ b/investigations/bedrock2/Aes/AesToRiscV.v
@@ -1,0 +1,202 @@
+Require Import Coq.ZArith.ZArith.
+Require Import Coq.Strings.String.
+Require Import Coq.micromega.Lia.
+Require Import Coq.derive.Derive.
+Require Import bedrock2.Syntax.
+Require Import compiler.FlatToRiscvDef.
+Require Import compiler.MemoryLayout.
+Require Import compiler.Pipeline.
+Require Import compiler.RiscvWordProperties.
+Require Import coqutil.Word.Interface.
+Require Import coqutil.Map.Z_keyed_SortedListMap.
+Require Import coqutil.Z.HexNotation.
+Require Import Bedrock2Experiments.WordProperties.
+Require Import Bedrock2Experiments.StateMachineMMIO.
+Require Import Bedrock2Experiments.StateMachineSemantics.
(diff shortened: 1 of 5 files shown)


satnam6502:
So we have a first task for Sam next week then to fix this restricition?

jadephilipoom:
Here's the result of compilation: https://gist.github.com/jadephilipoom/9abdf453cfa551b8f79262da1db01d48

jadephilipoom:
> So we have a first task for Sam next week then to fix this restriction?

Maybe so! I think the reason is that the compiler doesn't currently apply register spilling to arguments.

investigations/bedrock2/Aes/AesToRiscV.v
```diff
@@ -0,0 +1,202 @@
...
+  MemoryLayout.heap_pastend  := word.of_Z (8*2^10);
+  MemoryLayout.stack_start   := word.of_Z (8*2^10);
+  MemoryLayout.stack_pastend := word.of_Z (16*2^10);
+                              |}.
+
+(* Magic number for aes base address found in
+   third_party/opentitan/hw/top_earlgrey/sw/autogen/top_earlgrey.h:
+
+   #define TOP_EARLGREY_AES_BASE_ADDR 0x40110000u *)
+Definition AES_BASE_ADDR : Z := Ox"40110000".
```
satnam6502, line 54: Noted. Let's keep an eye on this in case it has changed since the commit we use. Perhaps this is now the USB base address at HEAD?

satnam6502 (approved):
Stamped.

investigations/bedrock2/Aes/AesToRiscV.v
jadephilipoom: That file, despite being "autogen", is actually version-controlled, so I just checked. Looks like it has indeed changed. Current master has `0x41100000u`: https://github.com/lowRISC/opentitan/blob/8b9fe4bf2db8ccfac0b26600decf07cf41867e07/hw/top_earlgrey/sw/autogen/top_earlgrey.h#L592

PR #803, opened 2021-05-26
Proofs about AES assembly

jadephilipoom:
- Compiles the AES example program (which uses the AES API to perform a full round of encryption)
- Uses the bedrock2 compiler proofs to show that the compiled assembly has the expected behavior

Right now, the specification that describes what exactly the circuit does with the input is an axiom. I think a good next step from here would possibly be to plug in the real AES spec from our circuit verification, create an example program that runs _decryption_, and prove that composing the two is the identity all the way at the assembly level -- that would give us some assurance that we can build on top of the proofs.

### DIFF
--- a/investigations/bedrock2/Aes/AesExampleProperties.v
+++ b/investigations/bedrock2/Aes/AesExampleProperties.v
@@ -47,8 +47,7 @@ Section Proofs.
                     key0 key1 key2 key3 key4 key5 key6 key7
                     iv0 iv1 iv2 iv3 : Semantics.word)
         (* initial values of output array (used only for determining length) *)
-        (ciphertext_arr : list Semantics.word)
-        (s : state),
+        (ciphertext_arr : list Semantics.word),
         let plaintext_arr := [plaintext0; plaintext1; plaintext2; plaintext3] in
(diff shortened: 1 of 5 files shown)


jadephilipoom:
Thinking a little more about the current status: with this PR, I think all of the essential objectives for the AES firmware verification are met. The interface with the control logic should be the AES state machine; if that connection works, we should have the whole thing, software and hardware, end-to-end.

There are some remaining nice-to-haves:
- Test the assembly by embedding it in C and running OpenTitan's aes.c tests on it
- Prove that decryption is the inverse of encryption in software
- Update the state machine model to allow writing new input while the previous input is still processing (this should probably wait until the control-logic connection is working with the simpler model, just to minimize complexity)
- Various ergonomic improvements (e.g. representing registers as nats instead of an inductive so that the uniqueness of register addresses is easier to prove)

However, none of these things seem essential to me, so I may go ahead and tackle some unrelated backlog I've been meaning to get to before moving forward with them.

satnam6502 (approved):
Stamped! Looking forward to Sam joining so he can do some code reviews :-)

PR #805, opened 2021-05-27
Refactor circuit simulation definition

jadephilipoom:
- Changed `fold_left_accumulate` list definition to more closely mimic the behavior of circuit loops
- Defined `simulate` directly in terms of `fold_left_accumulate`, so now we can reason about loops using more than just invariants
- Update everything to use the new definition of `simulate`

This resolves some awkwardness of reasoning about loops with sequential bodies, and hopefully unblocks some changes like introducing a circuit-retiming abstraction that were previously blocked on the awkwardness of reasoning about `simulate`.


blaxill (approved):
Lgtm. I suppose this drops the key expansion proofs as we dont provide a verified key expansion routine?

jadephilipoom:
> Lgtm. I suppose this drops the key expansion proofs as we dont provide a verified key expansion routine?

Pretty much. The dropped file used to basically provide logic that helped deal with reasoning about precomputed keys vs on-the-fly computed keys, and it's no longer needed.

PR #806, opened 2021-06-01
AesSemantics.v: reuse injectivity proof from Coq's standard lib

samuelgruetter:
`Injective f <-> NoDup (map f domain_of_f)` is already proven
in Coq's standard lib, so we can reuse it instead of writing
a quadratically slow proof.
Also removed one unnecessary usage of unique_words.


samuelgruetter:
(these are just two things I wondered while studying the aes code, I'll try to do less bikeshedding in the future :wink:)

jadephilipoom (commented):
Great, thanks! This was one of the slowest proofs, so I'm very happy to see a better solution.

jadephilipoom (approved):

PR #814, opened 2021-06-04
sketch riscv machine model that concurrently runs a cava-defined device

samuelgruetter:
I'm trying to work on the following question: How could we prove that a Cava-defined hardware accelerator satisfies the assumptions that a bedrock2-defined source program makes about the Cava-defined hardware accelerator?
Right now, these assumptions are given as a state machine in eg [AesSemantics.v](https://github.com/project-oak/silveroak/blob/3b6f8d26cff1d215ac58bb84ce95c14eb439c6fa/investigations/bedrock2/Aes/AesSemantics.v), but if the Cava device really satisfies them, we should be able to state an end-to-end theorem where this state machine definition cancels out, so I tried to sketch what such a theorem statement would look like.

Since we're not attempting to verify the RISC-V processor, we need some kind of hybrid RISC-V machine model, where the processor is just a spec, but the devices are written in Cava.

The core of this sketch is to define a step of the combined hybrid system, `system_step`, as a nondeterministic choice between a step of the processor spec, `processor_step`, and a step of the Cava device, `device_step`.

I don't know if this makes sense without further explanations, I didn't write too many source comments, but it might be good to discuss this next week. Also, I don't know if this kind of concurrency is easy to deal with in proofs, I haven't done this before, so there's a risk that this sketch is a bad idea, and that we should take a completely different approach.

A note on terminology: I call the state of the Cava device "extra state", and I call communication between the processor model and the Cava device "internal MMIO", and don't record it in the event trace, and call communication with the external world (not currently present in this sketch) "external MMIO" and do record it in the event trace.


investigations/bedrock2/RiscvMachineWithCavaDevice/CavaIncrRiscvMachine.v
```diff
@@ -0,0 +1,98 @@
...
+  Definition incr_update_v0: unit * (signal Bit * signal (Vec Bit 32)) ->
+                    cava (unit * (signal Bit * signal (Vec Bit 32))) :=
+    fun '(_, (busy, value)) =>
+            busy' <- ret (constant false);;
+            value' <- incrN value;;
+            ret (tt, (busy', value')).
+
+  (* Search (?signal (?A * ?B)%type) (?signal ?A * ?signal ?B)%type. *)
+  Fail Definition incr: Circuit unit unit := Loop (Comb incr_update_v0).
+  (* How to go from `signal A B` to `(signal A * signal B)`? *)
```
jadephilipoom, line 25: There is no `signal (A * B)`; arguments to `signal` are `SignalType`s, not Coq types, and products are not permitted. Although `Circuit` takes arbitrary type arguments, you'll run into lots of issues if it's anything other than some tupling of `signal`s (can be nested tuples), so `Circuit unit unit` is not recommended (we've been going back and forth about how to make this type system more obvious without getting loads of type-mismatch errors from Coq when it can't see how to reconstruct a signal tuple).

 To get the same meaning, try `Circuit (signal Void) (signal Void)` (which will be `Circuit unit unit` in the Coq semantics). I think it will work to write something like:
```coq
  Definition incr_update_v0: signal Void * signal Bit * signal (Vec Bit 32) ->
                             cava (signal Void * signal Bit * signal (Vec Bit 32)) :=
    fun '(i, busy, value) =>
      busy' <- ret (constant false);;
      value' <- incrN value;;
      ret (i, busy', value').

Definition incr : Circuit (signal Void) (signal Void) := Loop (Loop (Comb incr_update_v0)).
```

The tutorial might be helpful if you want to understand the circuit definitions, it's a bit more polished than our examples directory: https://project-oak.github.io/silveroak/demo/tutorial

jadephilipoom (commented):
Thanks for taking the initiative on this! Let's discuss it soon. I think our "seminar" slot this week is unclaimed; it would be fine by me for you to claim it for this discussion.

samuelgruetter:
This is still very much WIP, but it might be good if @jadephilipoom and/or @blaxill already have a look at the interface between the RISC-V processor and the Cava simulator that I made up:

https://github.com/project-oak/silveroak/blob/82d4af8ca3a8fbf5d679c59930f985aa21f9478e/investigations/bedrock2/RiscvMachineWithCavaDevice/InternalMMIOMachine.v#L21-L62

And here's a very simple instantiation of this interface (with an increment device):

https://github.com/project-oak/silveroak/blob/82d4af8ca3a8fbf5d679c59930f985aa21f9478e/investigations/bedrock2/RiscvMachineWithCavaDevice/DetIncrMachine.v#L67-L78

And here's how the RISC-V processor model invokes the methods of the `device` class defined above:

https://github.com/project-oak/silveroak/blob/82d4af8ca3a8fbf5d679c59930f985aa21f9478e/investigations/bedrock2/RiscvMachineWithCavaDevice/InternalMMIOMachine.v#L110-L154

Do you think it will be possible to make an HMAC and/or AES Cava device conform to this interface? Any suggestions for changes? Also, I have not yet written any proofs involving this interface, so if you have a completely different interface/approach in mind, I'd by happy to try that too!

blaxill:
> This is still very much WIP, but it might be good if @jadephilipoom and/or @blaxill already have a look at the interface between the RISC-V processor and the Cava simulator that I made up:

This looks reasonable to me so far, the device hardware side is general enough that we can fit anything to it and the software side fits my understanding.

investigations/bedrock2/RiscvMachineWithCavaDevice/InternalMMIOMachine.v
```diff
@@ -0,0 +1,235 @@
...
+
+    (* poll whether the previous read request has been answered, if None,
+       we'll have to wait, ie call run1 again and again until we get Some *)
+    readResp: state -> option word;
+
+    (* make a write request *)
+    writeReq(num_bytes: nat)(addr value: word): state -> state;
+
+    (* poll whether the previous write request has been answered *)
+    writeResp: state -> option unit;
```
jadephilipoom, line 50: The hardware side doesn't include anything about serving read/write requests right now, so that's something we'd have to add if we wanted to reason about responding to those based on the circuit state. In particular, we'd need a way to remember what the previous request was. But this isn't an issue if the read/write results are always answered immediately.

investigations/bedrock2/RiscvMachineWithCavaDevice/Run.v
```diff
@@ -0,0 +1,135 @@
...
+  end.
+
+(* Useful for debugging: display (ok-flag, pc, output) after each cycle:
+Compute snd (trace 100 initial).
+*)
+
+Definition res(nsteps: nat): (bool * Z * Z) := outcomeToLogElem (run sched nsteps initial).
+
+(* We can vm_compute through the execution of the IncrementWait program,
+   riscv-coq's processor model, and Cava's reaction to the IncrementWait program: *)
```
jadephilipoom, line 132: Nice!

investigations/bedrock2/RiscvMachineWithCavaDevice/MMIOToCava.v
```diff
@@ -0,0 +1,106 @@
...
+    unfold stepH in H0.
+    unfold nth_step.
+    (*
+      above the line:
+      Run.run1 with Bedrock2Experiments.Riscv.MetricMinimalMMIO.IsRiscvMachine
+      below the line:
+      Run.run1 with Bedrock2Experiments.RiscvMachineWithCavaDevice.InternalMMIOMachine.IsRiscvMachine
+
+      would like to say "forall monadic programs, L behaves like H" --> L needs to go through
+      monad materialization as well *)
```
jadephilipoom, line 72: Shot in the dark here, but does it help to know that in the Coq semantics for Cava circuits, the monad entirely disappears? For circuit definitions, we use the monads to preserve sharing for the netlist generation, but for proofs it's not needed and is instantiated with an identity monad. If you're having issues with monads in proofs, you should be able to just remove them without issues; it's safe to talk about `A -> B` instead of `A -> cava B`.

jadephilipoom:
Looks good to me! I think the interfaces make sense.

samuelgruetter:
After the discussion in yesterday's meeting, I wanted to try if I can change the interface so that the Cava device does not have to remember the latest request, so here's a new interface with a `run1` that includes the request as an argument:

https://github.com/project-oak/silveroak/blob/c6cfe1cc0429fa39d0b29f2bc23665b1419b6a6a/investigations/bedrock2/RiscvMachineWithCavaDevice/InternalMMIOMachine.v#L22-L52

This signature of `run1` of the form `state -> input -> state * output` matches the Cava's `step` used in `simulate`, so that it fits more naturally with this sample implementation of `IncrementWait`:

https://github.com/project-oak/silveroak/blob/c6cfe1cc0429fa39d0b29f2bc23665b1419b6a6a/investigations/bedrock2/RiscvMachineWithCavaDevice/DetIncrMachine.v#L22-L68

In particular, I don't need to modify internal Cava state any more, but all interaction now happens over input/output of Cava's `step`.

And the RISC-V processor code invoking the methods of the `device` class defined above now looks as follows:

https://github.com/project-oak/silveroak/blob/c6cfe1cc0429fa39d0b29f2bc23665b1419b6a6a/investigations/bedrock2/RiscvMachineWithCavaDevice/InternalMMIOMachine.v#L106-L126

However, one caveat with this approach is that in order to get a response from the Cava device, you have to run it for at least one cycle (and after that, maybe a few more cycles until `is_resp` becomes true), that is, 0-cycle responses are not possible any more. @jadephilipoom @blaxill do you think that's a problem? And would you prefer the previous version or the new version of this interface?

investigations/bedrock2/RiscvMachineWithCavaDevice/CavaIncrRiscvMachine.v
samuelgruetter: A belated thank you for showing me the trick to apply the `Loop` combinator several times, once for each component of the state, that worked!

investigations/bedrock2/RiscvMachineWithCavaDevice/InternalMMIOMachine.v
samuelgruetter: Good point, addressed in https://github.com/project-oak/silveroak/pull/814/commits/c6cfe1cc0429fa39d0b29f2bc23665b1419b6a6a, in such a way that the processor keeps feeding the same request to the device as long as it waits for a response, so there's no need to remember what the previous request was.

investigations/bedrock2/RiscvMachineWithCavaDevice/MMIOToCava.v
samuelgruetter: This was actually about the riscv-coq monad, not the Cava monad, but your comment applies there as well, there's only one monad instance, and it can be simplified away and entirely disappears, and I do make use of this!

samuelgruetter:
I think I've reached a milestone where it would be good have this reviewed and merged soon.

This PR now contains an end-to-end theorem for the `IncrementWait` example spanning the whole stack, all the way down to a risc-v machine that executes the Cava simulator to serve the MMIO requests made by the bedrock2 driver program.

The `IncrementWait_end2end_correct` theorem is now almost admit-free: Besides the propositional & functional extensionality axioms, there are only three more admits, which all are uninteresting properties about words (`naive_riscv_ok`, `incrN_word_to_bv` and `MMIOToCava.bv_to_word_word_to_bv`).

Here's an overview of the different layers of specification:

1) High-level spec: `scalar (word.of_Z output_ptr) (word.add (word.of_Z 1) input)` in the separation logic predicate of the postcondition of `IncrementWait_end2end_correct`, saying that the output pointer points to `input+1` (adding 1 to the input is all that this minimal demo does, but in the most complicated way I've ever implemented this operation :wink:)
2) bedrock2 weakest-precondition generator (invoking the state machine defined in `IncrementWaitSemantics` on external calls)
3) bedrock2 Inductive semantics (invoking the state machine defined in `IncrementWaitSemantics` on external calls)
4) a risc-v machine model compatible with the bedrock2 compiler (`investigations/bedrock2/Riscv/MinimalMMIO.v` and `investigations/bedrock2/Riscv/MinimalMMIO.v`), invoking the state machine defined in `IncrementWaitSemantics` on MMIO loads and stores
5) a risc-v machine model that executes the Cava simulator running the `CavaIncrementDevice` to serve MMIO loads and stores

And here's how each pair of two adjacent layers is bridged by a theorem:

* 1)-2): Lemma `put_wait_get_correct` in `IncrementWaitProperties`
* 2)-3): Lemma `bedrock2.WeakestPreconditionProperties.sound_cmd`
* 3)-4): The generic compiler correctness lemma `compiler.Pipeline.compiler_correct`, where the external calls compiler has been instantiated to `Bedrock2Experiments.StateMachineMMIO.compile_ext_call`, which compiles external function calls called `MMIOWRITE` to a single risc-v store instruction, and external function calls called `MMIOREAD` to a single risc-v load instruction, and is proven correct in the same file, and this external calls compiler correctness proof is passed to the generic compiler correctness lemma
* 4)-5): Lemma `Bedrock2Experiments.RiscvMachineWithCavaDevice.MMIOToCava.stateMachine_to_cava` proves that if all level-4 risc-v machine executions satisfy a postcondition, then all level-5 risc-v machine executions also satisfy that postcondition. This proof is generic over the state machine and Cava device, and takes a `device_implements_state_machine` hypothesis (defined in the same file), which is proven for the increment device in `investigations/bedrock2/IncrementWait/CavaIncrementDevice.v`.

In `Bedrock2ToCava.v`, a theorem spanning layers 2)-5) is proven generically for all possible state machines and related Cava devices, so that when we want to prove end-to-end theorems for other devices and their drivers, we can reuse this one and only need to write a driver, a Cava device, and a state machine against which we prove both the driver and the Cava device, but once we've done that, we should be able to obtain an end-to-end theorem for this new device, and it should only require a proof as short as the one in `IncrementWaitToCava.v` (less than 70 lines, and probably shrinkable further).

blaxill:
Great work! At a high level this all makes sense to me, but it's going to take me a few days to digest the code.

blaxill (approved):

blaxill (approved):
Lgtm, I dont think I have anything useful to add, hopefully I can test out your `InternalMMIOMachine` machinery for something larger soon.

PR #815, opened 2021-06-04
Finish mask ROM uart.c transcription (functional test passes)

dayeol:
This PR completes the transcription of `sw/device/silicon_creator/lib/drivers/uart.c`.

Note: 
`memset` function used in `_br2_store` and `_br2_load` somehow breaks the functional test.
Thus, I have temporarily made it volatile memory access so that the functional test passes.
Also, unittest will not run because we don't have mock functions for `b2_abs_mmio_write32` and `b2_abs_mmio_read32`.

You can test it by manually running the following commands:

In `investigations/bedrock2`:
```
make uart.c
cp uart.c <opentitan dir>/sw/device/silicon_creator/lib/drivers/uart.c
```

And then in opentitan dir:
```
ninja -C build-out all

build/lowrisc_systems_chip_earlgrey_verilator_0.1/sim-verilator/Vchip_earlgrey_verilator \
    --meminit=rom,build-bin/sw/device/boot_rom/boot_rom_sim_verilator.scr.40.vmem \
    --meminit=otp,build-bin/sw/device/otp_img/otp_img_sim_verilator.vmem \
    --meminit=flash,build-bin/sw/device/silicon_creator/testing/sw_silicon_creator_lib_driver_uart_functest_sim_verilator.elf
```

This should produce output like
```
TEST PASSED CHECKS
 _____         _                                  _ _
|_   _|__  ___| |_   _ __   __ _ ___ ___  ___  __| | |
  | |/ _ \/ __| __| | '_ \ / _` / __/ __|/ _ \/ _` | |
  | |  __/\__ \ |_  | |_) | (_| \__ \__ \  __/ (_| |_|
  |_|\___||___/\__| | .__/ \__,_|___/___/\___|\__,_(_)
                    |_|
```


investigations/bedrock2/Uart/Constants.v
```diff
@@ -127,3 +127,34 @@ Definition constant_names : uart_constants string :=
     kErrorUartInvalidArgument := "kErrorUartInvalidArgument";
     kErrorUartBadBaudRate := "kErrorUartBadBaudRate";
   |}.
+
+(* This list includes all the constants and is prepended to functions' argument
+   lists to initialize their environment *)
```
jadephilipoom, line 132: This comment is a bit outdated (guessing you got it from the AES code, so I probably need to update it there!). Previously, I used to define the arguments of every function as `globals ++ <other arguments>`; however, that ended up exceeding limits in the bedrock2 compiler for how many arguments a function can have, so I later changed to passing each function only the global values it actually needs.
```suggestion
(* This list includes all the constants *)
```

investigations/bedrock2/Uart/UartToC.v
```diff
@@ -15,6 +15,137 @@ Local Open Scope Z_scope.
...
+    c_cmd indent c2
+  | cmd.skip =>
+    indent ++ "/*skip*/" ++ LF
+  | cmd.call binds f arges =>
+    (* the below line filters out globals from arguments in the declaration *)
+  (*let arges := filter (fun e => match e with
+                               | expr.var s => negb (existsb (String.eqb s) uart_globals)
+                               | _ => true
+                               end) arges in
+     indent ++ c_call (List.map c_var binds) (c_fun f) (List.map c_expr arges) *)
```
jadephilipoom, line 52: This line strips out globals from calls to other functions; you'll probably need to uncomment it once you add globals to calls between UART functions.

investigations/bedrock2/Uart/Uart.v
```diff
@@ -160,7 +167,8 @@ Section Impl.
     let byte := "byte" in
     let reg := "reg" in
     let cond := "cond" in
-    ("br2_uart_putchar", ([byte], [],
+    ("b2_uart_putchar", ([UART_WDATA_WDATA_MASK; UART_WDATA_WDATA_OFFSET;
+    TOP_EARLGREY_UART0_BASE_ADDR; UART_WDATA_REG_OFFSET; byte], [],
     bedrock_func_body:(
       unpack! cond = uart_tx_full();
```
jadephilipoom, line 173: Since `uart_tx_full` now takes global arguments, you'll need to call it with those arguments for the proofs to work. (The C strips out these arguments from the function declaration, which is why the mismatch doesn't cause problems there.) Same goes for any other calls between UART functions.
```suggestion
      unpack! cond = uart_tx_full(TOP_EARLGREY_UART0_BASE_ADDR, UART_STATUS_REG_OFFSET, UART_STATUS_TXFULL_BIT);
```

jadephilipoom (approved):
Great work!

jadephilipoom:
> Also, unittest will not run because we don't have mock functions for `b2_abs_mmio_write32` and `b2_abs_mmio_read32`.

Maybe we could write them; I'll make an issue to look into this.

investigations/bedrock2/Uart/UartToC.v
dayeol: how do we want to handle the actual arguments that are global variables?
we'd always copy them to local variables before we pass them as arguments?

jadephilipoom: I see what you mean. Yes, I think the best solution here is to copy them to local variables, although unfortunately that makes the correspondence with the source code less obvious.

investigations/bedrock2/Uart/Uart.v
dayeol: done

investigations/bedrock2/Uart/Constants.v
dayeol: done

investigations/bedrock2/Uart/UartToC.v
dayeol: fixed

PR #822, opened 2021-06-09
study the `timed A := nat -> A` monad

samuelgruetter:
@jadephilipoom pointed me to the files [`TimedMonad.v`](https://github.com/project-oak/silveroak/blob/0adc80b5194a54afff21b7648dea4d3993dc2ea2/cava/Cava/Acorn/TimedMonad.v) and [`SequentialTimed.v`](
https://github.com/project-oak/silveroak/blob/0adc80b5194a54afff21b7648dea4d3993dc2ea2/cava/Cava/Acorn/SequentialTimed.v), which simulate circuits using `timed A` instead of `list A`, where `timed A` is defined as `nat -> A`. [`SequentialTimed.v`](
https://github.com/project-oak/silveroak/blob/0adc80b5194a54afff21b7648dea4d3993dc2ea2/cava/Cava/Acorn/SequentialTimed.v) contains a loop definition `loopSeqF` with a builtin delay.

If I understood @satnam6502 correctly, he would like to define `loop` without a builtin `delay`, and put explicit `delay` elements inside the "body" of the loop, so I was curious to see if the approach in these two files would allow this, and made some stand-alone experimentation file based on them.

I managed to write some code that *seems* to create a loop without builtin delay: You can write a loop body like

```coq
Definition sample_loop_body_with_delay(inp_signal: timed (combType Nat)):
  timed (combType Nat) -> timed (combType Nat) :=
  fun sum_signal =>
    inp <- inp_signal;;
    sum <- delay sum_signal;;
    res <- add inp sum;;
    res' <- incr res;;
    res'' <- decr res';;
    ret res''.
```

which explicitly calls `delay` on `sum_signal` inside its body, and then you can make a definition like

```coq
Definition sum_countUp: timed (combType Nat) :=
  loop (sample_loop_body_with_delay countUp).
```

This looks too good to be true, so here are the (severe) limitations:
1) I don't actually define the loop combinator as a `Definition`, but as a `Notation`, in order to delay the termination checking to the actual usage. This also means that it can only be applied to concrete expressions for which Coq's termination checker succeeds, and it won't be possible to prove general theorems about the loop combinator.
2) I changed the signature of the loop body: It doesn't take an input signal any more (that seemed to have become unnecessary, but I could add it back), and instead of taking just the current state as an argument, it takes a `timed State`, ie the whole time series of state values.
3) I rely on the termination checker unfolding the loop body and (potentially) also nested definitions inside it, as well as the definition of `delay`, until it is convinced that the recursive calls are on structurally smaller `nat`s. When it works, it happens all automatically, but it will be hard to debug when it doesn't, and will not scale well with bigger expressions where unfolding everything could lead to huge terms and thus huge slowdown in the termination checking.

Given these limitations, I don't think we should use such an approach, but just wanted to share this file in case someone else is about to redo similar experiments.

### DIFF
--- a/investigations/separate_loop_from_delay_in_timed_monad.v
+++ b/investigations/separate_loop_from_delay_in_timed_monad.v
@@ -18,18 +18,20 @@ Notation "x <- c1 ;; c2" := (bind c1 (fun x => c2))
   (at level 61, c1 at next level, right associativity).
 Notation "f >=> g" := (mcompose f g) (at level 61, right associativity).
 
-Inductive SignalType := Bit | Nat.
+Inductive SignalType := Bit | Nat | Pair(A1 A2: SignalType).
 
-Definition combType(A: SignalType) :=
+Fixpoint combType(A: SignalType) :=
   match A with
   | Bit => bool
   | Nat => nat
+  | Pair A1 A2 => (combType A1 * combType A2)%type
   end.
 
-Definition defaultCombValue(A: SignalType): combType A :=
+Fixpoint defaultCombValue(A: SignalType): combType A :=
   match A with
   | Bit => false
   | Nat => 0
+  | Pair A1 A2 => (defaultCombValue A1, defaultCombValue A2)
   end.
 
 Definition delayWithInit{A}(init: combType A)(x: timed (combType A)): timed (combType A) :=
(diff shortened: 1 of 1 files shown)


satnam6502:
At first sight this looks very interesting. Let me take a closer look and see what I can do with it...

blaxill:
Interesting idea!

This is a special case of your disadvantage (1) but unfortunately it won't be able to be used in higher order circuits definitions that put their arguments in a loop either (as at notation time the argument will be unresolved).

samuelgruetter:
Update: I managed to lift restrictions (1) and (3).

The trick is that if the loop body delays its output by at least one time step, then a fuel amount of `t+1` is sufficient to simulate the loop, so the fact that we use fuel can be hidden as an implementation detail in the definition of the `loop` combinator.
That is, we can define `loop f` simply as applying `f` to itself `t+1` times:

```coq
Definition loop{B: SignalType}(f: timed (combType B) -> timed (combType B)): timed (combType B) :=
  fun t => power_fun f (S t) (fun _ => defaultCombValue B) t.
```

And we can prove that the `loop` combinator behaves like a fixpoint combinator:
```coq
Theorem loop_fix{A}(f: timed (combType A) -> timed (combType A))(d: nat):
  0 < d ->
  delaysAtLeast f d ->
  loop f = f (loop f).
```

And as an example, I made a circuit `fib` with two delays computing Fibonacci numbers:

```
      -------      -------   ---
      |delay|      |delay|   | |
  +-->|with |--+-->|with |-->|a|
  |   |init |  |   |init |   |d|
  |   |  1  |  |   |  0  |   |d|--+----->
  |   -------  |   -------   |e|  |
  |            |             |r|  |
  |            +------------>| |  |
  |                          ---  |
  |                               |
  +-------------------------------+
```

and proved that it behaves like the Gallina specification `fib_spec`.

blaxill:
Awesome! I'm going to take a deeper look in a minute but this sounds very promising. 

Currently I think this would not work for a circuit such as `loop ( swap >=> fsT delay )` (using your definition of loop), here each path has a guaranteed after two unrolls. Increasing the assumed gas would work right? Is there a way to make this work for a loop that needs an arbitrary 'n' unrolls?

blaxill:
FWIW I'll also push some experiments on indexed monads soon, but I think your approach is more promising if we can cover all cases.

samuelgruetter:
Yeah, `loop (swap >> fsT delay)` would not work with this approach, and doubling the fuel in the definition of `loop` would fix this particular example, but I don't quite know how to best generalize it and how to adapt the `loop_fix` theorem...

jadephilipoom:
Very cool! I think it would be okay to not deal with loops that delay different state values different amounts; those could be represented with nested loops. However, it would be good to handle loops with different amounts of delay on the _state_ values and on the _output_ values. The current `loop` signature doesn't distinguish feedback and output, if I understand correctly, which is suitably general (you can always not use all your state). However, if we add back the distinction, perhaps it would be possible to reason about a loop with different amounts of delay on the values that are and are not fed back.

One note here that might or might not help with proofs is that, if the state value is first and the output value is second, loop (x >> snD delay) is equivalent to loop x >> delay.

samuelgruetter:
@jadephilipoom, are you suggesting a solution to the `loop (swap >> fsT delay)` example that involves nested loops and/or separating the loop body output into a "state" and "output" part? Or is your comment unrelated to that example?

jadephilipoom:
> @jadephilipoom, are you suggesting a solution to the `loop (swap >> fsT delay)` example that involves nested loops and/or separating the loop body output into a "state" and "output" part? Or is your comment unrelated to that example?

My comment is related to that example -- if I understand correctly, the problem is that the loop delays only one part of its state, so things get out of sync. But if we separate loop "state" (which is fed back to the beginning of the loop block) and loop "output" (which is not), then we can represent different amounts of delay on the state parts with nested loops and different amounts of delay on the output parts won't be a problem because we don't need to loop them back. It would also allow loops with no delay at all on the output.

samuelgruetter:
> if I understand correctly, the problem is that the loop delays only one part of its state, so things get out of sync.

I see another problem: In the body of this loop, there is one path that does not get delayed at all. Just to make sure we're discussing the same circuit (I used different signatures than standard Cava), here's a pic of the circuit I'm considering:

IMG_20210616_101127

The upper input doesn't get delayed at all! Yet, this is a valid circuit, but in order to prove it, you have to put your finger one of the input wires, and follow the wire until you get back to where you started, and note that no matter where you started, by the time you get back to your starting point, you have traversed at least one delay, but this journey went through the loop body **twice**, and the `delaysAtLeast` judgment I used in this PR can't be used to prove this kind of "advanced loop body validity".

jadephilipoom:
Ah, I see now! Thanks for the detailed clarification :slightly_smiling_face: 

I think that this circuit _can_ be expressed as a loop that would pass the validity check: `loop (swap >> fsT delay)` is equivalent to `loop delay >> fork2`, no? I'm not sure if it's the case that _every_ circuit can be transformed this way, though. At any rate, this validity check, even if it would fail some valid circuits, is certainly an improvement over the delay automatically being fused to the loop.

blaxill:
> Ah, I see now! Thanks for the detailed clarification 🙂
> 
> I think that this circuit _can_ be expressed as a loop that would pass the validity check: `loop (swap >> fsT delay)` is equivalent to `loop delay >> fork2`, no? I'm not sure if it's the case that _every_ circuit can be transformed this way, though. At any rate, this validity check, even if it would fail some valid circuits, is certainly an improvement over the delay automatically being fused to the loop.

I'm not suggesting we put too much effort into it, but it would be great if we could crack this case as otherwise loopy combinators will be more restrictive (and if the combinator is providing the loop it will be less clear how to do the circuit transform to a valid representation)

samuelgruetter:
A simple hack might be to pass an additional `nat` argument to `loop`, with `loop n body` standing for "a loop whose proof that all paths contain at least one delay requires `n` unfoldings of the loop body", or, equivalently, "a loop whose output at time `t` can be correctly simulated by using a fuel amount of `n*(t+1)`". In the case of `loop (swap >> fsT delay)`, for example, `n` would be 2. But I don't know if unfolding a loop body `n` times will always allow to prove validity, I think the reason why a circuit is valid can be arbitrarily complex if you have loops whose body contains a multiplexer whose selector signal represents a complicated expression?

samuelgruetter:
Currently, Cava2 does not provide any loop-like combinator that does not also introduce delays and state on its own, but it seems that for the applications we're interested in right now, this feature is not needed. I still believe that if this feature is desired, the approach in this PR would be a good one to try out, but in that case it would have to be applied to the Cava2 code, so in the meantime, I'm closing this PR.

PR #829, opened 2021-06-16
Apply Changes for Global Variables (#818) to Uart

dayeol:
We have been passing global variables and constants to bedrock2
functions as arguments.
This has been fixed by #818 thus apply the changes to Uart.


satnam6502 (approved):
Great! Nice to see the global variables contained.

PR #830, opened 2021-06-17
[WIP] Experiment with Adam Chlipala's variant record suggestion

satnam6502:
Adam said:

> At first, I couldn't see anything hardware-specific about your method 
> (in Lava and Cava) of writing down circuit generators. There is the 
> well-known duality between record and variant types, and your encoding 
> with type classes is basically the record version (instances being the 
> records) of writing out the same syntax trees with GADTs (variant 
> version), which is quite natural in Coq.  I imagine you will find it 
> impossible to prove key metatheorems about generic circuits, e.g. where 
> you need to instantiate a generator with two different instances and 
> know that the results are related.

I'm trying to understand exactly what Adam means when he says "variant version". Here I've tried to sketch an
alternative to our typeclass version which uses an inductive data-type instead. Is this what Adam means?
Or does he really mean a `Record` type?

### DIFF
--- a/investigations/acorn2/Acorn2.v
+++ b/investigations/acorn2/Acorn2.v
@@ -1,0 +1,49 @@
+(****************************************************************************)
+(* Copyright 2021 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
+(* you may not use this file except in compliance with the License.         *)
+(* You may obtain a copy of the License at                                  *)
+(*                                                                          *)
+(*     http://www.apache.org/licenses/LICENSE-2.0                           *)
+(*                                                                          *)
+(* Unless required by applicable law or agreed to in writing, software      *)
+(* distributed under the License is distributed on an "AS IS" BASIS,        *)
+(* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *)
+(* See the License for the specific language governing permissions and      *)
+(* limitations under the License.                                           *)
+(****************************************************************************)
+
+Require Import Coq.Strings.Ascii Coq.Strings.String.
+Require Import Coq.Strings.Ascii.
+Require Import Coq.Lists.List.
+Require Import Coq.ZArith.ZArith.
+Require Import ExtLib.Structures.Monad.
+Require Import ExtLib.Data.Monads.StateMonad.
+Require Import ExtLib.Data.Monads.IdentityMonad.
+Require Import ExtLib.Structures.MonadState.
+
+Require Import Coq.Numbers.DecimalString.
+
+Import ListNotations.
+Import MonadNotation.
+Local Open Scope monad_scope.
+Local Open Scope type_scope.
+
+Inductive SignalType :=
+  | Bit : SignalType 
+  | Nat : SignalType.
+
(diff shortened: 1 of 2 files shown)


jadephilipoom:
Yes, my interpretation is similar -- "record" means defining a record type or class (class in Coq is just record plus typeclass inference, so I don't think there's a difference whether we do `Record` or `Class`) and then creating instances of that record/class for different interpretations. I'm not familiar with the term "variant types" to describe it, but the rest of the context seems to imply using a single underlying structure, interpreted in two different ways, to get the different interpretations of the circuit that we want.

investigations/acorn2/Acorn2.v
```diff
@@ -0,0 +1,49 @@
...
+  | NatType : N -> Signal Nat.
+
+Inductive Acorn : Type -> Type :=
+| Inv : Signal Bit -> Signal Bit -> Acorn (Signal Bit -> Signal Bit)
+| And2 : Signal Bit -> Signal Bit -> Signal Bit -> Acorn (Signal Bit * Signal Bit -> Signal Bit)
+| Compose : forall {t1 t2 t3}, Acorn (t1 -> t2) -> Acorn (t2 -> t3) -> Acorn (t1 -> t3).
+
+Definition nandGate : Acorn (Signal Bit * Signal Bit -> Signal Bit)
+  := Compose (And2 (BitType 1) (BitType 2) (BitType 3))
+             (Inv (BitType 3) (BitType 4)).
```
samuelgruetter, line 48: What does `BitType n` mean? Does it mean "the wire with id n"? If so, do you want to require the user to ensure that wire ids are globally unique?

blaxill:
Its not obvious to me that he would mean this structure with fields containing references (nats) vs the actual inductive structure. For example in Koika:

https://github.com/mit-plv/koika/blob/37a4d8b1fe9b4b9a131e193281ebd33b28ce1c35/coq/CircuitSyntax.v#L25

Or my example of reifying from the Acorn monad to an inductive structure (with an example evaluation method):

https://github.com/project-oak/silveroak/pull/825/files#diff-b4a03022b46b8decadfecad67ef4f2a3600ffad0302f103a7f96242e9dfa1ee5R58

I'm constructing the inductive `CavaDeep` instances in my example from the Acorn monad class to follow the Acorn design. Terms could instead be written directly with and evaluated with denotation function as above, or as in my arrow examples last year e.g. `investigations/Arrow/CircuitSemantics.v`

samuelgruetter:
I wrote a sketch of how I understood Adam's comment, and took the liberty to push it to Satnam's cava2 branch (file `AcornGADT.v`), so that it appears on this PR as well and can be discussed together.

It contrasts with the typeclass-based approach as follows:
- Instead of starting with a typeclass with one method per circuit component kind, we start with a GADT (`Inductive Acorn`) indexed over the `SignalType` that it produces, with one inductive constructor per circuit component kind.
- Instead of providing different instances of the same typeclass, we provide different interpretation functions for the same `Inductive`. In this example, they are `Simulation.interp` and `Netlist.interp`.

Note that this is just a quick sketch to illustrate the concept, and important questions (eg, how to feed input into such circuits) are left as an exercise :wink:

samuelgruetter:
I think this has been implemented in Cava2, so I guess we can close this?

Issue #816, opened 2021-06-07
Mock mmio write/read functions so OT's unit tests run


jadephilipoom:
> Also, unittest will not run because we don't have mock functions for `b2_abs_mmio_write32` and `b2_abs_mmio_read32`.

Maybe we could write them; I'll make an issue to look into this.

_Originally posted by @jadephilipoom in https://github.com/project-oak/silveroak/issues/815#issuecomment-855738626_

satnam6502:
@jadephilipoom can this be closed via #831 ?

jadephilipoom:
Yep, looks like it!

PR #831, opened 2021-06-17
translate hmac (sha256) to bedrock2 (functest & unittest pass)

samuelgruetter:
This PR translates the silicon creator `hmac.c` into bedrock2, and adds a script that pretty-prints the bedrock2 code to C, inserts it in opentitan, and runs the functional as well as unit tests. It requires a recent opentitan checkout (the opentitan submodule of silveroak is too old). I also added a script that runs the uart functest & unittest. Previously, only the uart functest could be run. The trick to also get the unittest to run is to not pretty-print `abs_mmio_write32`/`abs_mmio_read32` but use the existing opentitan versions of them, which also makes more sense from a verification point of view, because the only thing these functions do is one MMIO, and MMIO is modeled differently in bedrock2 and C: in bedrock2, MMIOs are external function calls, whereas in C, MMIO actually looks like accessing the memory.

In order to run the unit tests and functional tests, make a fresh clone of the opentitan repo, cd into it, and from there, invoke `/path/to/silveroak/investigations/bedrock2/run_hmac_in_opentitan.sh`, after following the instructions at the top of that script.

@dayeol could you please review the files `AbsMMIO.v`, `MMIOLabels.v`, `UartToC.v`, `run_uart_in_opentitan.sh` and tell me if you're ok with them because they concern the UART part?


samuelgruetter:
Also, if I understand correctly this will fix https://github.com/project-oak/silveroak/issues/816.

dayeol:
LGTM! Thanks for writing uart script!

jadephilipoom (approved):
LGTM! I'm liking the new strategy for constants, should save us a good deal of pain.

PR #836, opened 2021-06-21
Sketch Hmac state machine

samuelgruetter:
This PR models the subset of the functionality of the Hmac module that is needed by the sha256 hmac code in silicon_creator.
I inferred this specification from [the doc](https://docs.opentitan.org/hw/ip/hmac/doc/) and from [hmac.c](https://github.com/lowRISC/opentitan/blob/98ca8dbea38c65d1d93ae78c44ce476781d3961a/sw/device/silicon_creator/lib/drivers/hmac.c).

I wanted to use `word` as the register type for `StateMachineSemantics` instead of a separate `Inductive Register`, to avoid converting back and forth between them. This required some changes to `StateMachineSemantics`, `StateMachineMMIO`, and `AesSemantics`, so that the Hmac code can still benefit from the state-machine-agnostic compiler correctness proof in `StateMachineMMIO`.


investigations/bedrock2/Hmac/HmacSemantics.v
```diff
@@ -0,0 +1,229 @@
...
+Require Import coqutil.Byte.
+Require Import coqutil.Map.Interface.
+Require Import coqutil.Word.Interface.
+Require Import coqutil.Word.LittleEndianList.
+Require Import Bedrock2Experiments.StateMachineSemantics.
+Require Import Bedrock2Experiments.Hmac.Constants.
+
+(* Note: AesSemantics.v seems to completely ignore its control register,
+   which results in a wrong specification in the sense that the specification
+   might say that a transition is possible even though it isn't.
```
jadephilipoom, line 13: The mode section of the control register is ignored (that's a TODO for if/when we actually support multiple modes), but the decryption/encryption flag is [used](https://github.com/project-oak/silveroak/blob/858fdead1fad2e33a3ee641ab9d1db0d9598668d/investigations/bedrock2/Aes/AesSemantics.v#L282) to compute the expected output. Are you talking about the mode section, or some other issue?

In any case, let's maybe make an issue for this comment instead of adding it into the source code; I always worry about references to other files in comments getting out of date and becoming misleading.

investigations/bedrock2/Hmac/HmacSemantics.v
```diff
@@ -0,0 +1,229 @@
...
+      write_step (CONSUMING b)
+                 (word.of_Z (TOP_EARLGREY_HMAC_BASE_ADDR + HMAC_MSG_FIFO_REG_OFFSET)) v
+                 (CONSUMING (b ++ le_split 4 (word.unsigned v)))
+  | write_hash_process: forall b,
+      write_step (CONSUMING b)
+                 (word.of_Z (TOP_EARLGREY_HMAC_BASE_ADDR + HMAC_CMD_REG_OFFSET))
+                 (word.of_Z (Z.shiftl 1 HMAC_CMD_HASH_PROCESS_BIT))
+                 (PROCESSING b max_negative_done_polls).
+
+  (* TODO: Can we make these register conventions more explicit in our spec?
```
jadephilipoom, line 182: I think this is a good idea; it would make our specifications clearer and easier to match to OpenTitan's.

jadephilipoom (approved):

investigations/bedrock2/Hmac/HmacSemantics.v
samuelgruetter: I don't quite know how to do that, so I opened an issue: https://github.com/project-oak/silveroak/issues/837

investigations/bedrock2/Hmac/HmacSemantics.v
samuelgruetter: I was worrying about the mode and key length part of the control register (and indeed missed that you're already using the decryption/encryption flag).
I removed the part of the comment about Aes so that it doesn't get out of sync.
I think there would be a way to soundly model only a subset of the features of a hardware module, in such a way that the following condition holds:

*If a program can be proven correct against our less feature-complete model, it is also correct when run on the more feature-complete actual hardware device.*

It seems that since the current Aes model ignores the mode and key length part of the control register, it does not satisfy this condition, but also, that it would be easy to change so that it satisfies the condition. I'll open an issue for this after some more thought.

jadephilipoom: Maybe we could even model the more complex behavior and prove this property, just with the two models?

PR #838, opened 2021-06-23
Deep embedding experiments

jadephilipoom:
This provides some examples of a Cava representation built on a PHOAS-style expression, with AST `Bind` and `Ret` operations replacing monad instances. The Coq semantics and the netlist are different interpretations of the same expression. The examples and minimal circuit structure are based on Satnam's "acorn" experiments in #786 (the SystemVerilog translation is actually pretty much completely unchanged). Additionally, I included a definition of the [fibonacci](https://project-oak.github.io/silveroak/demo/tutorial#example-7-fibonacci-sequence) circuit from our tutorial and a proof of its correctness.

Some cool stuff:
- **Binds work across sequential components**: Like the acorn experiments, you can name wires that come out of delays and loops (for example, `x <- LoopDelay 0 foo @ y ;; And2 (x, y)`)
- **No unit types in concrete circuit states** : Unit types get automatically simplified away if all the types are concrete; for instance, the state of the `fibonacci` circuit is `Nat * Nat` instead of the ` unit * (unit * combType (Vec Bit sz)) * unit * combType (Vec Bit sz))` type in the tutorial.
- **Proofs are reasonable**: The proof for the fibonacci sequence is a little simpler than in the tutorial, but part of that is replacing `Vec Bit sz` with `Nat`, so I'd say the proof complexity was about the same as what we have now. If we want to do compiler proofs in the future, however, an AST representation has a lot of advantages.

Some meh stuff:
- **Casts for state types of circuits with abstract state** : The magic that gets rid of unit types for concrete state types doesn't work quite as well for circuits with _abstract_ state types (for instance, the `fsT` combinator, whose state is whatever the passed-in circuit's state is). For this use case, the expression has a `Cast` operator that essentially declares that if you can prove you can convert to and from two types, you can change the state type of a circuit from one to the other. You can pass these in manually, but by default some typeclass instances try to figure the conversions out for you (and succeed for all of my small examples). I think the result is something that _will work_ for any state types that are actually convertible, but maybe not without some manually-programmed plumbing.

Some not-so-cool stuff:
- **No destructing binders**: I haven't been able to get binders that bind multiple names (e.g. `'(x, y) <- foo ;; ...`) to work consistently, because although the new type system does produce Gallina pairs, the `denote_type` definition is not always unfolded enough. Similarly, sometimes `Ret (x, y)` doesn't typecheck, and you have to write `Ret (t:=_*_) (x, y)` to make it extremely clear that you're writing a pair.
- **Circuit state types are declared explicitly**: Rather than being computed based on the circuit structure, circuit state types are part of the type declaration. You can let the state type be auto-computed by leaving an underscore, e.g. `foo : Circuit Nat _ (Bit * Nat)`, but that might leave you annoyingly verbose terms like `tzero ** Nat ** (tzero ** tzero)` (which simplifies to `Nat` because the `**` operation removes extra `tzero`s; `foo : Circuit Nat Nat (Bit * Nat)` would also typecheck). One thing I did while writing these circuits was to leave an underscore at first, then copy over the term and figure out what it simplifies to, then replace the underscore with the simplified version, but it's a bit of an awkward workflow.


blaxill (approved):
Do you have examples of the retiming too? If possible i'd like to experiment with how different representations interact with your retiming work.

jadephilipoom:
> Do you have examples of the retiming too? If possible i'd like to experiment with how different representations interact with your retiming work.

Working on it, should have something by EOD today.

PR #839, opened 2021-06-23
Interactive circuit retiming

jadephilipoom:
This code is a big mess still -- I've redefined weak equivalence and retiming about a million times, most recently about two hours ago. But the core of the idea is in `RetimingProperties.v`; I want to be able to use lemmas to interactively push delays around inside a proof. More details about some of the thought process here are in the notes document I've shared. Retiming inside of loops is the hard part -- you have to reason about keeping the loop state and input in sync.

This adds a lightweight type system on top of the existing Circuit structure to make reasoning about adding delays to arbitrary circuits' states possible.

### DIFF
--- a/cava/Cava/Core/CavaClass.v
+++ b/cava/Cava/Core/CavaClass.v
@@ -36,7 +36,6 @@ Class Cava {signal : sdenote} := {
   constantV : forall {A} {n : nat}, Vector.t (signal A) n -> signal (Vec A n);
   (* Default values. *)
   defaultSignal : forall {t: SignalType}, signal t;
-  defaultValue : forall {t : type}, value t := fun t => default_value (@defaultSignal) t;
   (* SystemVerilog primitive gates *)
   inv : signal Bit -> cava (signal Bit);
   and2 : signal Bit * signal Bit -> cava (signal Bit);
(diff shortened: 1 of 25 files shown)


jadephilipoom:
Updated this with my latest attempts. Current status: mostly working, but has some admits that might be tricky (although probably not impossible). I think overall I'm happy with the _interface_ here, but not necessarily with this specific implementation (which might not be relevant if we change the core library anyway).

The end goal is this kind of setup, where a user can interactively insert delays wherever they want in a circuit. Here, we can insert two delays to pipeline a simple nand circuit:
retiming_nand

For a more complex example, here's a simplified version of the AES cipher getting retimed to match a new, retimed `sub_bytes`. Note that everything can be left abstract! We don't have to know exactly what the retimed circuit, or indeed the original circuit, actually looks like in order to figure out where the new delays should go:
retiming_cipher

This was an interesting investigation, and it seems like a really cool and potentially useful feature to be able to manipulate circuit expressions in this way (not just for retiming; you can also simplify or re-group components). However, there's significant technical challenge involved, so I'm not sure whether it's worth the effort to support it.

jadephilipoom:
Since this work will have to be redone for the cava2 rewrite, I'm closing the PR for now. The latest version has been pushed to the `retiming-interactive` branch on the main repo so we can reference it later if we decide to pursue this feature.

Issue #840, opened 2021-06-23
Omitting features safely


samuelgruetter:
As we model hardware devices using state machines in Coq, we often only focus on the subset of features that the software is using, rather than modeling all features of the hardware device.

In order for our proofs to be meaningful, the following condition should be true:

*If a program can be proven correct against our less feature-complete model, it is also correct when run on the more feature-complete actual hardware device.*

I'm opening this issue because it seems to me that the current model in `AesSemantics.v` does not satisfy this condition.

For instance, imagine a program that intends to perform AES-256 encryption, but erroneously writes `kAes128` instead of `kAes256` to the `KEY_LEN` bits in the `AES_CTRL` register of the AES module. If I read `AesSemantics.v` correctly, it claims that the hardware module always performs AES-256, no matter what value was written to the `KEY_LEN` bits, so I believe it would be possible to prove correctness for a wrong program, which means that the spec is broken.

One way to solve this issue would simply be to completely model the hardware devices, but I think there could/should be a safe way of leaving out certain features in the model. The intuition is that if the hardware module has some features about which we don't tell the software correctness proof anything, the software correctness proof should still hold on a more feature-rich hardware module.

PR #841, opened 2021-06-23
When is it safe to omit features in a state machine modeling a hardware device?

samuelgruetter:
Here's an attempt (not successful yet) to formalize what I meant in https://github.com/project-oak/silveroak/issues/840 by "omitting features safely": I define a predicate `state_machine_extends SM1 SM2`, intended to capture the fact that state machine `SM1` is an extended (ie, having more features) version of state machine `SM2`, and that any program proven correct against `SM2` would also run correctly on a device implementing `SM1`.

If we do end-to-end proofs, we will not need this: We can just prove the software and hardware against the same state machine and have the state machine, which is an intermediate specification, cancel out.

The reason I still started this experiment is for the case where we prove a bedrock2 program against some state machine, but omit some features in this state machine, and we need to argue why it is safe to omit these features. In any case, this argument will not be carried out in Coq (because if we had actually modeled all features, we could just prove the bedrock2 program against the complete state machine), but I still wanted to try to formally state the notion of "a state machine extends another state machine", and to see if it preserves bedrock2 program logic proofs, so that part of the non-Coq argument could be supported by a Coq proof.

However, I got stuck on details for which I don't quite know whether they are technical details or more fundamental, and since this is not needed for end-to-end proofs, and also not needed if we model hardware devices fully, I decided to stop working on it and just post it here FYI.

### DIFF
--- a/investigations/bedrock2/ExtSpecExtends.v
+++ b/investigations/bedrock2/ExtSpecExtends.v
@@ -1,0 +1,40 @@
+Require Import Coq.ZArith.ZArith.
+Require Import bedrock2.Syntax bedrock2.Semantics.
+Require Import coqutil.Map.Interface.
+Require Import coqutil.Word.Interface.
+Local Open Scope Z_scope.
+
+Section WithParameters.
+  Context {width: Z}
+          {word: word width}
+          {mem: map.map word Byte.byte}
+          {locals: map.map String.string word}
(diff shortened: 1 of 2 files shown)


samuelgruetter:
Closing this as it's not needed in the end-to-end proofs that we're currently focusing on, and it's already tracked as a TODO item in https://github.com/project-oak/silveroak/issues/840.

PR #844, opened 2021-06-30
Bedrock2 Transcription of  Keymgr and add README

dayeol:


investigations/bedrock2/Keymgr/Keymgr.v
```diff
@@ -0,0 +1,217 @@
...
+      // Write and lock (rw0c) the max key version.
+      abs_mmio_write32(kBase + KEYMGR_MAX_CREATOR_KEY_VER_REG_OFFSET, max_key_ver);
+      abs_mmio_write32(kBase + KEYMGR_MAX_CREATOR_KEY_VER_REGWEN_REG_OFFSET, 0);
+
+      // Advance to CREATOR_ROOT_KEY state.
+      advance_state();
+      return kErrorOk;
+    }
+   ***)
+  (* Note: unrolling the loop/array might be easier to reason *)
```
jadephilipoom, line 165: Was the loop unrolled in this transcription because it was easier to transcribe that way, or because of the expectation that it will be easier to prove the function correct that way? If the former, that's fine, but if the latter, I think the optimization might be premature. The `unroll_while` lemma allows you to completely unroll constant-size loops in proofs, so I think the proof overhead might not be so bad. And all other things being equal, it's nice to match up with the source as closely as we can.

investigations/bedrock2/README.md
```diff
@@ -0,0 +1,96 @@
...
+First, setup OpenTitan and SilverOak repositories and build them:
+
+**OpenTitan**
+
+Follow the [Getting started with Verilator](https://docs.opentitan.org/doc/ug/getting_started_verilator/) to setup test environment for OpenTitan with Verilator.
+You may need to install prerequisites and software tools before starting the simulation with Verilator.
+
+**SilverOak**
+
+Follow the [SilverOak README](https://github.com/project-oak/silveroak) to setup silveroak repository.
```
jadephilipoom, line 20:
```suggestion
Follow the [SilverOak README](https://github.com/project-oak/silveroak) to setup the silveroak repository.
```

investigations/bedrock2/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+
+### Testing
+
+To run the unit tests, run following command in the OpenTitan's root directory, where `<silveroak>` is the SilverOak repo top and `<device>` is the device source file name without `.c` extension.
+For example, to test `uart.c`, replace `<device>` with `uart`.
+
+```
+# in opentitan repo top
+<silveroak>/investigations/bedrock2/run_silicon_creator_tests_in_opentitan.sh <device>
```
jadephilipoom, line 78: Nice to have this interface smoothed out and documented!

jadephilipoom (approved):
LGTM! The README is a very good idea, nice to have everything written down and discoverable.

investigations/bedrock2/Keymgr/Keymgr.v
dayeol: It was the former, if that makes sense.
The input was a small array which are basically sequential MMIO registers, and the loop was accessing all of them.

PR #845, opened 2021-07-06
Alternative PHOAS 

blaxill:
Just experimenting with some different choices for a PHOAS representation:

- Provides partial application without the standard way of making functions part of the
type representation.
- Provides a notation that uses the embedded apply consistently to
reduce require type annotations.


jadephilipoom (approved):
LGTM. The let/delay construct is cool!

PR #847, opened 2021-07-16
Example of using first order representation to get state

blaxill:
Just a quick example of casting to a first order representation to get the state. Alternatively doing it directly with the PHOAS rep. is also possible i think by selecting the right `var` type. The benefit of an explicit first order representation is that it might be useful for Jade's retiming work that benefits from a binderless inductive structure.

### DIFF
--- a/investigations/ast/AST2.v
+++ b/investigations/ast/AST2.v
@@ -302,3 +302,65 @@ Proof.
     erewrite <-list_unit_equiv. reflexivity. }
 Qed.
 
+Inductive FCircuit : nat -> type -> type -> Type :=
+| FVar : forall {s x}, nat -> FCircuit s [] x
+| FAbs : forall {s x y z}, FCircuit (S s) y z -> FCircuit s (x ** y) z
+| FApp : forall {s x y z}, FCircuit s (x ** y) z -> FCircuit s [] x -> FCircuit s y z
+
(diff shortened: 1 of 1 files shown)


investigations/ast/AST2.v
```diff
@@ -302,3 +302,65 @@ Proof.
...
+  end.
+
+Definition fib_fo {sz} := to_first_order 0 (fibonacci (sz:=sz)).
+
+Definition get_phoas_state {i o s var} (_: Circuit (var:=var) s i o) := s.
+
+Compute (fib_fo (sz:=12)).
+Compute fo_state (fib_fo (sz:=12)).
+
+Goal forall sz, get_phoas_state (var:=fun _ => nat) (fibonacci (sz:=sz)) = fo_state (fib_fo (sz:=sz)).
```
jadephilipoom, line 362: Going a bit further here, I think it should be possible to define a function `to_fo_state : forall {i s o} (c : Circuit s i o), denote_type s -> denote_type (fo_state (to_first_order c)))` and its inverse. Since we'd pretty much only need to use the type equivalence to coerce _values_ of one type to the other, this might be easier than having to use `rew` with the type equivalence.

jadephilipoom (approved):
LGTM, thanks for writing this out!

PR #850, opened 2021-07-21
Add untested TLUL register adapter 

blaxill:
Primitive operations are currently axiomatized as they may change


satnam6502:
I'll stamp this for now to allow progress, but I've also added Sam as a reviewer in case he wants to chime in.

satnam6502 (approved):

PR #856, opened 2021-07-27
simulate risc-v machine interacting with Cava Hmac device

samuelgruetter:
This PR adds an end-to-end simulator in Coq like the one we already have for IncrementWait, but this time for Sha256-Hmac.

The line `Compute snd (trace 50 initial)` in `RunSha256SoftwareOnCava.v` starts with a risc-v machine `initial` whose instruction memory contains a main function that invokes sha256 on a simple test vector, and runs it instruction by instruction, and after each instruction, it logs a triple of the form `(ok_flag, pc, digest)`, where `ok_flag` tells if an error occurred, `pc` indicates the current PC, and `digest` is the memory region into which the program is supposed to store the SHA-256 hash resulting from the test vector.

Currently, it does not yet compute the correct hash, because instead of invoking Ben's Cava Hmac device, I invoke a [dummy device](https://github.com/samuelgruetter/silveroak/blob/5030dd2431dd371b9f2e23fa6c416b526ddb7d6a/investigations/bedrock2/Hmac/CavaHmacDevice.v#L16-L34) that does nothing.

As a first milestone towards end-to-end integration verification of the Hmac driver software and the Hmac Cava code, I'd suggest that we (i.e, @blaxill, more concretely :wink:) replace this dummy device by the actual Cava Hmac device, and see what the simulation in `RunSha256SoftwareOnCava.v` outputs. Even if it does not output the expected digest (I'm actually quite sure it won't because the driver software and all the wrapper code in `RunSha256SoftwareOnCava.v` are unverified), this exercise should enable us to find out quickly if at least the signatures of the definitions we're developing are compatible, a first step towards integration.


blaxill (approved):
Lgtm! Sorry for the delay, I missed this for a while somehow.

PR #857, opened 2021-07-28
Link to OpenTitan installation instructions in README

### LINKED ISSUE #852: The link to install OpenTitan has no instructions how to install OpenTitan.
The last line in the Pre-requisites section, regarding installing OpenTitan, the OpenTitan link (to https://github.com/lowRISC/opentitan/blob/master/README.md) is not helpful, there are no instructions there.

Also, if this step is optional it should be made more clear.

jadephilipoom:
Fixes #852; previously we just linked to OpenTitan's README, but their "getting started with Verilator" page is more useful. Thanks to @fshaked for flagging this!


fshaked (approved):
LGTM

PR #858, opened 2021-07-30
Fix a bug in UartProperties proof

dayeol:
The proof was failing due to the change in AbsMMIO spec

### DIFF
--- a/investigations/bedrock2/Uart/UartProperties.v
+++ b/investigations/bedrock2/Uart/UartProperties.v
@@ -166,6 +166,9 @@ Section Proofs.
   Local Ltac infer_reg_using_addr :=
     let base := TOP_EARLGREY_UART0_BASE_ADDR in
     lazymatch goal with
+      | H: reg_addr _ = _ |- _ => rewrite <- word.ring_morph_add in H
+      end;
+    lazymatch goal with
     | H: reg_addr _ = word.of_Z (base + UART_STATUS_REG_OFFSET) |- _ => apply reg_is_status in H; subst
     | H: reg_addr _ = word.of_Z (base + UART_CTRL_REG_OFFSET) |- _ => apply reg_is_ctrl in H; subst
     end.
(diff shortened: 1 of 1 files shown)


investigations/bedrock2/Uart/UartProperties.v
```diff
@@ -199,31 +202,35 @@ Section Proofs.
...
     { repeat straightline.
       straightline_call; eauto.
       repeat straightline.
-      (* keep "exection a x" for later eassumption *)
-      pose proof H5 as HH.
-      simpl in H5.
-      destruct H5. destruct H3.
-      replace s with x1 in *.
+      (* keep "execution a x" for later eassumption *)
+      repeat lazymatch goal with
```
samuelgruetter, line 215: do you really need `repeat` here?

investigations/bedrock2/Uart/UartProperties.v
```diff
@@ -199,31 +202,35 @@ Section Proofs.
...
+      | H0 : execution tr _, H1: execution ?a _ |-  context [?tr] => rename H0 into Hexec; rename H1 into Hexec'
+      end.
+      pose proof Hexec' as HH.
+      simpl in Hexec'.
+      destruct Hexec'.
+      lazymatch goal with
+      | H: execution tr ?x /\ step _ _ _ _ _ |- _ => destruct H; replace s with x in *
+      end.
```
samuelgruetter, line 223: can you replace the above 4 lines by something like `destruct Hexec' as (s' & ? & ?). replace s with s' in *`?

investigations/bedrock2/Uart/UartProperties.v
dayeol: no i don't

investigations/bedrock2/Uart/UartProperties.v
dayeol: fixed.

samuelgruetter (approved):

PR #859, opened 2021-08-02
Computable primitives 

blaxill:
Adds computable semantics for the currently implemented primitives. 

The denotation for `Vec` is slightly unusual but works quite well for the fibonacci benchmark, not to difficult to change in the future if its an issue though.


investigations/cava2/_CoqProject
```diff
@@ -0,0 +1,15 @@
+INSTALLDEFAULTROOT = ./
+-R ../../third_party/coq-ext-lib/theories ExtLib
+-R ../../third_party/bedrock2/deps/coqutil/src/coqutil coqutil
+-R . Cava
+-R ../../cava/Cava Cava
```
jadephilipoom, line 5: Two paths for `Cava`?

investigations/cava2/Types.v
```diff
@@ -14,24 +14,54 @@
...
+Defined.
+
+(* simple_denote_type denotes all 'Vec's as lists *)
+Fixpoint simple_denote_type (t: type) :=
+  match t with
+  | Unit => unit
+  | Bit => bool
+  | Vec t n => list (simple_denote_type t)
+  | Pair x y => (simple_denote_type x * simple_denote_type y)%type
+  end.
```
jadephilipoom, line 49: Why do we need to have two denotations?

investigations/cava2/Expr.v
```diff
@@ -116,47 +119,152 @@ Module ExprNotations.
...
-    }}.
-    (* Notation "f >=> g" := (compose f g) (at level 61, right associativity) : expr_scope. *)
-
-  End Var.
 End ExprNotations.
+
+Section Var.
+  Context {var : tvar}.
+
+  Local Open Scope N.
```
jadephilipoom, line 127: Same comment about scopes.

investigations/cava2/Expr.v
```diff
@@ -116,47 +119,152 @@ Module ExprNotations.
...
+  Definition _1 {sz} := Constant (val_of (BitVec sz) 1).
+  Definition _2 {sz} := Constant (val_of (BitVec sz) 2).
+
+  Class bitlike x :=
+  { eq : var x -> var x -> Circuit [] [] Bit
+  ; not : var x -> Circuit [] [] x
+  (* ; xor : var x -> var x -> Circuit [] [] x *)
+  (* ; and : var x -> var x -> Circuit [] [] x *)
+  (* ; add : var x -> var x -> Circuit [] [] x *)
+  }.
```
jadephilipoom, line 143: Shouldn't everything be `bitlike`? The types are unit, vec, bit, and pair; it seems like all of those have a straightforward implementation of bitwise operations.

investigations/cava2/Expr.v
```diff
@@ -116,47 +119,152 @@ Module ExprNotations.
...
+    {{ fun vec => `UnaryOp UnVecToTuple vec` }}.
+  Definition slice {var t n} (start length: nat): Circuit (var:=var) [] [Vec t n] (Vec t length) :=
+    {{ fun vec => `UnaryOp (UnVecSlice start length) vec` }}.
+  Definition replace {var t n i}: Circuit (var:=var) [] [Vec t n; BitVec i; t] (Vec t n) :=
+    {{ fun vec index val => `TernaryOp TernVecReplace vec index val` }}.
+
+End PrimitiveNotations.
+
+Axiom value_hole : forall {t}, t.
+Axiom circuit_hole : forall {t var}, Circuit (var:=var) [] [] t.
```
jadephilipoom, line 270: What are these for?

jadephilipoom (approved):
There's some cleanup still needed, but I'm happy with the general idea here.

investigations/cava2/Fib.v
```diff
@@ -58,68 +61,74 @@ Fixpoint fibonacci_nat (n : nat) :=
...
   end.
 
-Definition spec_of_fibonacci (sz : nat) (input : list unit) : list nat
-  := map (fun n => fibonacci_nat n mod (2 ^ sz)) (seq 0 (List.length input)).
+Definition spec_of_fibonacci (sz : nat) (input : list unit) : list N
+  := map (fun n => N.of_nat (fibonacci_nat n) mod (2 ^ N.of_nat sz))%N (seq 0 (List.length input)).
 
 Lemma fork2_step A state input : step (fork2 (A:=A)) state (input, tt) = (tt, (input, input)).
 Proof. reflexivity. Qed.
+Open Scope N.
```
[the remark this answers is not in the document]
blaxill: Done thanks

investigations/cava2/Hmac.v
```diff
@@ -36,9 +36,13 @@ Section Var.
   Definition hmac_register_index := 5.
   Definition hmac_register := BitVec hmac_register_index.
 
+  Local Open Scope N.
+
   Definition REG_INTR_STATE := Constant (0: denote_type hmac_register).
   Definition REG_CMD := Constant (5: denote_type hmac_register).
 
+  Unset Printing Notations.
```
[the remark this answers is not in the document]
blaxill: Done thanks

investigations/cava2/_CoqProject
blaxill: This is temporary so I can call the local module Cava but also access the old Cava at the other path, mostly to access Cava Lemmas.

investigations/cava2/Types.v
blaxill: I'm not 100% confident of the `denote_type` version, although I like it it might turn out worse with lots of bit operations. I'll keep two for now just so its easier to switch if needed.

investigations/cava2/Expr.v
blaxill: This is currently not too thought out, feel free to modify! Some unanswered questions include, should shift/rotate be part of bitlike (its not clear if e.g. bit shifting a tuple is useful?)

investigations/cava2/Expr.v
blaxill: Just temporary values I can quickly use in the circuits that still have parts TODO, but satisfy type checking in slightly different ways.

investigations/cava2/Expr.v
jadephilipoom: I think it makes sense to say that all our signals are either bit-vectors (including bits, which are bit-vectors of size 1, and units, which are bit-vectors of size 0) or collections of bit-vectors (either as vectors or pairs). I think it'd be consistent to define a bit-shift on a tuple or multi-dimensional vector as bit-shifting every vector in that collection of bit-vectors.

investigations/cava2/Sha256.v
```diff
@@ -143,3 +145,9 @@ Section Var.
 
 End Var.
 
+(* Section SanityCheck. *)
+(*   Require Import Cava.Semantics. *)
+(*   Time Compute step sha256_message_schedule_update tt (0,(3,(6,(9,tt)))). *)
+(*   Open Scope list_scope. *)
+(*   Time Compute step sha256_compress tt ([1;2;3;4;5;7;8;9],(1,(2,tt))). *)
+(* End SanityCheck. *)
```
samuelgruetter, line 153: These `Compute` commands look exciting! How long does it take to run them?
If it doesn't take too long, they should run on CI to prevent regressions. But without spamming the build log nor the name space, so could you do something like this?
```coq
Goal step sha256_message_schedule_update tt (0,(3,(6,(9,tt)))) = ....
Proof. vm_compute. reflexivity. Abort.
```

samuelgruetter (approved):

investigations/cava2/Expr.v
blaxill: I'm completely open to changing the embedded type system btw. I've also been thinking about dropping Bit to just use bitvectors length=1, making units just bitvectors length=0 also makes sense to me. 
Re:shifting, at a syntactic level i'd expect shifting a vector to shift its contents rather than bitshift internally. Perhaps we could have a sugared vs desugared syntax, and the mixing intent vs available operations could be made clearer. I wouldn't suggest this until after I've completed the hmac though

investigations/cava2/Sha256.v
blaxill: They are quick so far, <0.01s. Hopefully it will remain fast for larger circuits. I'll look at adding them to CI once I've checked they are computing the correct values :D 

I've opened #863 to track this

PR #860, opened 2021-08-02
Move firmware

### LINKED ISSUE #855: run CI on investigations/bedrock2
Now that we work quite a bit in `investigations/bedrock2`, it would be good to run CI on this directory, to check whether (seemingly) independent PRs still work after merging another PR.

jadephilipoom:
Resolves #855 

- Moves the entire directory `investigations/bedrock2` into a new top-level directory called `firmware/ (the Coq library name is still "Bedrock2Experiments", which we might want to change separately)
- Adds a top-level Makefile rule and a CI job for the `firmware/` directory

This should help us make sure changes don't break the bedrock2 proofs, and upgrades the firmware work outside of `investigations` given it's not really an experiment any more! :slightly_smiling_face: I didn't add the OpenTitan tests to the CI, because they use different versions of OpenTitan at the moment, but if that's resolved at some point it might be good to add.


blaxill (approved):
Lgtm! thanks

satnam6502 (approved):

PR #864, opened 2021-08-03
Replace parameter records by individual parameters

samuelgruetter:
This PR updates bedrock2 to a new version that uses individual parameters instead of parameter records, and also replaces the parameter records in `firmware` by individual parameters.

While this results in slightly longer lists of `Context` variables at `Section` heads, it saves us from writing parameter record transformer instances, and solves all implicit parameter mismatch problems.

A few examples (from the diff of this PR) of how things become simpler:

```
execution (p := state_machine_parameters) (getLog initial) IDLE
```

becomes just

```
execution (getLog initial) IDLE
```

Beasts like eg this one

```coq
  (* change the implicit arguments of scalar to match hypothesis *)
  lazymatch goal with
  | H : sep _ _ ?m |- sep _ _ ?m =>
    lazymatch type of H with
    | context [(@scalar ?a1 ?b1 ?c1)] =>
      lazymatch goal with
      | |- context [(@scalar ?a2 ?b2 ?c2)] =>
        change a2 with a1; change b2 with b1; change c2 with c1
      end
    end
  end.
  subst a2.

  use_sep_assumption.
  cancel.
  cancel_seps_at_indices 0%nat 0%nat. 1: reflexivity.
  reflexivity.
```

now can be replaced by just

```coq
  ecancel_assumption.
```

and this one

```coq
    (* store result in memory *)
    simplify_implicits.
    ring_simplify_store_addr.
    (* the following line is in [straightline] but needs simplify_implicits for
       it to work *)
    eapply store_four_of_sep;
      [ simplify_implicits; solve [ ecancel_assumption ] |  ].
    repeat straightline.
```

now is just

```coq
    (* store result in memory *)
    ring_simplify_store_addr.
    repeat straightline.
```

### DIFF
--- a/firmware/StateMachineMMIO.v
+++ b/firmware/StateMachineMMIO.v
@@ -51,7 +51,7 @@ Definition compile_read(action: string): Z -> Z -> Z -> Instruction :=
      whether a 32-bit load is signed or unsigned makes no difference on a 32-bit machine. *)
   if String.eqb action READ8 then Lbu else if String.eqb action READ16 then Lhu else Lw.
 
-Definition compile_ext_call(results: list Z) a (args: list Z):
+Definition compile_interact(results: list Z) a (args: list Z):
   list Instruction :=
   if String.prefix WRITE_PREFIX a then
     match results, args with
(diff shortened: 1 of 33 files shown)


firmware/StateMachineMMIO.v
```diff
@@ -585,12 +540,6 @@ Section MMIO1.
         typeclasses eauto.
       }
       eauto 10.
-  Time Qed. (* takes ~30s *)
+  Qed.
```
jadephilipoom, line 543: Did this get faster?

jadephilipoom (approved):
This is fantastic! The mismatched implicit arguments were a major source of friction; behind all of those ugly tactics simplifying n layers of implicits is also significant time spent manually comparing massive `Set Printing All` expressions, and a lot of frustration. This should make future work go much more smoothly.

firmware/StateMachineMMIO.v
samuelgruetter: It already was "fast" (~ 1.5s) before this PR, and this PR doesn't change the Qed time significantly. I just removed an outdated comment here. In order to know which commit resulted in the speedup, we'd have to do some git archeology...

jadephilipoom: No need to dig it up, I think -- thanks for updating!

PR #865, opened 2021-08-05
Fix primitives

blaxill:
Also added a few new ones

### DIFF
--- a/investigations/cava2/Expr.v
+++ b/investigations/cava2/Expr.v
@@ -53,7 +53,7 @@ Section Vars.
     -> Circuit s [] z
 
   | Constant: forall {x}, denote_type x -> Circuit [] [] x
-  | MakeTuple: forall {s1 s2 x y}, Circuit s1 [] x
+  | MakePair: forall {s1 s2 x y}, Circuit s1 [] x
     -> Circuit s2 [] y
     -> Circuit (s1++s2) [] (x**y)
 
@@ -115,7 +115,7 @@ Module ExprNotations.
     (in custom expr at level 1, x at level 4, v constr at level 7, only parsing)  : expr_scope.
 
   Notation "( x , .. , y , z )" := (
-      MakeTuple x .. (MakeTuple y z) ..
+      MakePair x .. (MakePair y z) ..
     ) (in custom expr, x at level 1, y at level 1, z at level 1) : expr_scope.
 
   Notation "'if' i 'then' t 'else' e" := ((ElimBool i t e))
--- a/investigations/cava2/Expr.v
+++ b/investigations/cava2/Expr.v
@@ -126,16 +126,10 @@ End ExprNotations.
 Section Var.
   Context {var : tvar}.
 
-  Definition val_of t : denote_type t -> denote_type t := id.
-  Definition val_N {sz}: N -> denote_type (BitVec sz) := id.
-
   Definition True := Constant (true: denote_type Bit).
   Definition False := Constant (false: denote_type Bit).
-  Definition _0 {sz} := Constant (val_of (BitVec sz) 0).
-  Definition _1 {sz} := Constant (val_of (BitVec sz) 1).
-  Definition _2 {sz} := Constant (val_of (BitVec sz) 2).
-  Definition _3 {sz} := Constant (val_of (BitVec sz) 2).
-  Definition _4 {sz} := Constant (val_of (BitVec sz) 2).
+  Definition K {sz}: N -> Circuit [] [] (BitVec sz) :=
+    fun x : denote_type (BitVec sz) => Constant x.
 
   Class bitlike x :=
   { eq : var x -> var x -> Circuit [] [] Bit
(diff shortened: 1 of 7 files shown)


investigations/cava2/Expr.v
```diff
@@ -129,10 +129,13 @@ Section Var.
   Definition val_of t : denote_type t -> denote_type t := id.
   Definition val_N {sz}: N -> denote_type (BitVec sz) := id.
 
+  Definition True := Constant (true: denote_type Bit).
```
samuelgruetter, line 132: Why do you sometimes use `val_of` and sometimes use casts? Or asked differently: Is there anything `val_of` can achieve that casts can't achieve?

investigations/cava2/Expr.v
```diff
@@ -129,10 +129,13 @@ Section Var.
...
   Definition False := Constant (false: denote_type Bit).
   Definition _0 {sz} := Constant (val_of (BitVec sz) 0).
   Definition _1 {sz} := Constant (val_of (BitVec sz) 1).
   Definition _2 {sz} := Constant (val_of (BitVec sz) 2).
+  Definition _3 {sz} := Constant (val_of (BitVec sz) 2).
+  Definition _4 {sz} := Constant (val_of (BitVec sz) 2).
```
samuelgruetter, line 138: Can't you make a general definition or notation that works for all numbers? This would also prevent typos like the above, where you define 3 and 4 to be 2 :wink:

investigations/cava2/Expr.v
blaxill: No advantage to `val_of` vs being shorter. These constants annoyingly always need our embedded `type` annotation. Can you think of a better + less verbose way to do this?

samuelgruetter: So why don't you replace all usages of `val_of` with casts?
And I don't know of a better way than defining one `Definition` or `Notation` with a cast per constant (but all numbers could share the same one, I think).

blaxill: I was previously using casts everywhere, they are more verbose but not by much. Perhaps it is better to just revert to using the casts completely for now.

investigations/cava2/Expr.v
blaxill: I've removed these now

investigations/cava2/Expr.v
blaxill: I've removed `val_of`

investigations/cava2/Fib.v
```diff
@@ -35,13 +35,13 @@ Section Var.
...
 
-  Definition fib_init sz := val_of (BitVec sz) (2^(N.of_nat sz)-1).
+  Definition fib_init sz := 2^(N.of_nat sz)-1 : denote_type (BitVec sz).
 
   Definition fibonacci {sz: nat}: Circuit (BitVec sz ** BitVec sz) [] (BitVec sz) := {{
     let/delay r1 :=
       let r2 := delay r1 initially (fib_init sz) in
       r1 + r2
-      initially (val_of (BitVec sz) 1) in
+      initially (1: denote_type (BitVec sz)) in
```
samuelgruetter, line 44:
```suggestion
      initially (K 1) in
```
does that work?

investigations/cava2/Fib.v
```diff
@@ -35,13 +35,13 @@ Section Var.
     fun a => (a, a)
   }}.
 
-  Definition fib_init sz := val_of (BitVec sz) (2^(N.of_nat sz)-1).
+  Definition fib_init sz := 2^(N.of_nat sz)-1 : denote_type (BitVec sz).
```
samuelgruetter, line 38:
```suggestion
  Definition fib_init sz := K (2^(N.of_nat sz)-1).
```
does that work?

investigations/cava2/Expr.v
```diff
@@ -179,7 +176,7 @@ Section RegressionTests.
       if `silly_id` flag then (a) else a
   }}.
 
-  Definition inital_state {sz} := val_of (BitVec sz ** BitVec sz) (0,1)%N.
+  Definition inital_state {sz} := (0,1)%N : denote_type (BitVec sz ** BitVec sz).
```
samuelgruetter, line 179:
```suggestion
  Definition inital_state {sz}: denote_type (BitVec sz ** BitVec sz) := (K 0, K 1).
```
does that work?

investigations/cava2/Fib.v
blaxill: No- `K x` is `Constant x : Circuit [] [] (BitVec _)` a circuit. The "initially" part is a constant value (which cannot be a circuit).

investigations/cava2/Fib.v
blaxill: Passing the `sz` through this would work (`Definition fib_init sz := K (sz:=sz) (2^(N.of_nat sz)-1).`)

investigations/cava2/Expr.v
blaxill: No, similar to the first one, I'm using `K` as a very uninformative circuit constructor, different to how I was using `val_of`.

investigations/cava2/Fib.v
blaxill: ... or at least it can be defined like this, but the type is different from its current use case (as the others)

PR #868, opened 2021-08-06
Better support for typechecking Cava constants

samuelgruetter:
Currently,

```coq
Check (Constant 1 : Circuit [] [] (BitVec 32)).
```

yields

```
Error:
In environment
var : tvar
The term "1" has type "N" while it is expected to have type "denote_type ?x".
```

This PR adds a trick that makes the above `Check` succeed, as well as typechecking problems involving `LetDelay` and `Delay`, which suffer from the same problem.

`Constant`, `LetDelay` and `Delay` all expect an argument of type `denote_type ?x`, and even if the expression being typechecked has a known expected type like eg `Circuit [] [] (BitVec 32)`, Coq's typechecker does not propagate `(BitVec 32)` into the `?x` argument of `Constant`.

The trick is to temporarily give the typechecker an additional degree of freedom by introducing a variable type variable `X` for the type of the constant, and by requiring a constraint `denote_rel x X` whose purpose is to ensure that `denote_type x = X`, in a way which can be inferred by typeclass search, which, contrary to ordinary typechecking, does propagate `(BitVec 32)` from the expected type into the `?x` argument of `Constant`.

### DIFF
--- a/investigations/cava2/Expr.v
+++ b/investigations/cava2/Expr.v
@@ -35,12 +35,12 @@ Section Vars.
 
   | Let: forall {x y z s1 s2}, Circuit s1 [] x -> (var x -> Circuit s2 y z) -> Circuit (s1++s2) y z
   (* slightly different fomualtion, but equivalent to loop delay *)
-  | LetDelay : forall {x y z s1 s2}, denote_type x
+  | LetDelayRaw : forall {x y z s1 s2}, denote_type x
     -> (var x -> Circuit s1 [] x)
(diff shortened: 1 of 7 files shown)


blaxill:
Nice! Are there any times this breaks down? 
For naming I think I would keep the `LetDelay` names in `Circuit`, and use `Definition let_delay ...` style names for the new methods.

samuelgruetter:
Closing in favor of https://github.com/project-oak/silveroak/pull/871

Issue #869, opened 2021-08-06
Create an HMAC specification


jadephilipoom:
Just like we have a specification for AES, we should have one for HMAC. I was unable to find an existing Coq specification for HMAC that didn't depend on CompCert libraries (which we can't use because of licensing issues). The first step will be writing a specification for SHA-256.

PR #870, opened 2021-08-06
Initial setup for SHA-256 spec

jadephilipoom:
For #869 

- Sets up a folder for the HMAC spec to live in (modelled of AES code structure)
- Start the specification for SHA-256 (a subcomponent of HMAC) by implementing a few helper functions

Potentially controversial choice: I've written this spec using `N` instead of `Vector` or `list`, because I know we have some interpretations of datatypes in cava2 that convert bit-vectors to numbers and I think the algorithm might be clearer/easier to reason about in this way. However, happy to change it to lists/vectors/words if anyone has strong feelings.


blaxill (approved):
Lgtm

PR #871, opened 2021-08-06
make type argument of `Constant` explicit

samuelgruetter:
This allows us to write constants without a cast, ie instead of `Constant (literal : denote_type T)` we simply can write `Constant T literal`, thus saving us a `:`, a pair of parentheses, and a `denote_type`.

In the rare cases where the value passed to `Constant` has already a type of the form `denote_type _`, this approach costs us one `_`: Instead of `Constant value`, we now have to write `Constant _ value`.

Compared to https://github.com/project-oak/silveroak/pull/868, this is still a bit more verbose, but has about one fewer bell and whistle :wink:

### DIFF
--- a/investigations/cava2/Expr.v
+++ b/investigations/cava2/Expr.v
@@ -52,7 +52,7 @@ Section Vars.
     -> var (x**y)
     -> Circuit s [] z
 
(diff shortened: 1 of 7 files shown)


blaxill (approved):
Lgtm thanks!

PR #872, opened 2021-08-06
Full SHA-256 spec

jadephilipoom:
For #869 

I tested this on one example vector and it works, more rigorous testing to come later! I defined some local, only-parsing notations to keep as close to the FIPS document as possible. Overall it's pretty concise, around 200 lines, with imports only from the standard library.

### DIFF
--- a/silveroak-opentitan/hmac/Spec/SHA256.v
+++ b/silveroak-opentitan/hmac/Spec/SHA256.v
@@ -14,16 +14,14 @@
 (* limitations under the License.                                           *)
 (****************************************************************************)
 
+Require Import Coq.Lists.List.
 Require Import Coq.NArith.NArith.
+Require Import Coq.ZArith.ZArith.
+Import ListNotations.
 Local Open Scope N_scope.
 
 (* Specification of SHA-256 as described by FIPS 180-4:
    https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf  *)
-
-(* Notations for bitwise operations, all left-associative *)
-Local Infix "<<" := N.shiftl (at level 32, left associativity, only parsing).
-Local Infix ">>" := N.shiftr (at level 32, left associativity, only parsing).
-Local Infix "⊕" := N.lxor (at level 32, left associativity, only parsing).
 
 (* Word size for SHA-256 *)
 Definition w := 32.
@@ -31,13 +29,28 @@ Definition w := 32.
 (* Addition of w-bit words *)
 Definition add_mod (x y : N) := (x + y) mod (2 ^ w).
 
+(* Truncating shift-left *)
+Definition truncating_shiftl (x n : N) := (N.shiftl x n) mod (2 ^ w).
+
+(* Notations for bitwise operations as defined in section 2.2.2, all
+   left-associative *)
+Local Infix "<<" := truncating_shiftl (at level 32, left associativity, only parsing).
+Local Infix ">>" := N.shiftr (at level 32, left associativity, only parsing).
+Local Infix "|" := N.lor (at level 32, left associativity, only parsing).
+Local Infix "&" := N.land (at level 32, left associativity, only parsing).
+Local Infix "⊕" := N.lxor (at level 32, left associativity, only parsing).
+Local Notation "¬ x" := (N.lnot x w) (at level 32, only parsing).
+
+(* All addition henceforth is add_mod *)
+Local Infix "+" := add_mod (at level 50, left associativity, only parsing).
+
 (* From section 2.2.2:
 
    The rotate right (circular right shift) operation, where x is a w-bit word
    and n is an integer with 0 <= n < w, is defined by
        ROTR n x = (x >> n) ∨ (x << w - n).
  *)
-Definition ROTR n x := N.lor (x >> n) (x << (w - n)).
+Definition ROTR n x := (x >> n) | (x << (w - n)).
 
 (* From section 2.2.2:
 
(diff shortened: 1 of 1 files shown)


silveroak-opentitan/hmac/Spec/SHA256.v
```diff
@@ -64,3 +75,167 @@ Definition sigma0 (x : N) := (ROTR 7 x) ⊕ (ROTR 18 x) ⊕ (SHR 3 x).
...
+   d807aa98 12835b01 243185be 550c7dc3 72be5d74 80deb1fe 9bdc06a7 c19bf174
+   e49b69c1 efbe4786 0fc19dc6 240ca1cc 2de92c6f 4a7484aa 5cb0a9dc 76f988da
+   983e5152 a831c66d b00327c8 bf597fc7 c6e00bf3 d5a79147 06ca6351 14292967
+   27b70a85 2e1b2138 4d2c6dfc 53380d13 650a7354 766a0abb 81c2c92e 92722c85
+   a2bfe8a1 a81a664b c24b8b70 c76c51a3 d192e819 d6990624 f40e3585 106aa070
+   19a4c116 1e376c08 2748774c 34b0bcb5 391c0cb3 4ed8aa4a 5b9cca4f 682e6ff3
+   748f82ee 78a5636f 84c87814 8cc70208 90befffa a4506ceb bef9a3f7 c67178f2
+ *)
+Definition K : list N :=
+  [ 0x428a2f98; 0x71374491; 0xb5c0fbcf; 0xe9b5dba5; 0x3956c25b; 0x59f111f1; 0x923f82a4; 0xab1c5ed5;
```
blaxill, line 97: 👀  Am I crazy or is hexadecimal parsing a recent addition to Coq

blaxill (commented):
Lgtm!

silveroak-opentitan/hmac/Spec/SHA256.v
jadephilipoom: A recent addition! I realized it now existed when Sam used it :slightly_smiling_face:

samuelgruetter:
The signature of the SHA function in this PR is `sha256(l msg : N): N`, whereas in specs of bedrock2 functions, we'll need `sha256(msg: list byte): list byte` (or maybe `sha256(msg: list byte): list word`).
I was going to write a conversion adapter to obtain the desired signature, but when I looked at the definitions of `M`, `padded_msg`, and `concat_digest`, it seemed to me that they're a bit unhappy about the operators they're using: The high-level concepts they're trying to express are concatenation and indexing, but instead of using these high-level concepts, they emulate them using combination of `<<`, `>>`, `|` and `mod`, which seems a bit unnatural.
So maybe if these functions were expressed over `list byte` instead of `N`, they would look more natural, *and* we would not have to write a signature adapter and prove properties about it.
@jadephilipoom could you give this a try? Or did you already try and found reasons for preferring `N` over `list byte`?

jadephilipoom:
> @jadephilipoom could you give this a try? Or did you already try and found reasons for preferring `N` over `list byte`?

I did try to write the spec in terms of bytes and found that, while it makes bit concatenation less awkward, it makes other operations (addition modulo `2^w`, for instance, and shifting) so much more awkward that my evaluation was that wasn't worth it. However, if you'd like, I can resurrect that code and make an alternate PR so you can compare them yourself; if we need `list byte` for the interface, maybe the tradeoffs shift enough to make it worthwhile.

samuelgruetter:
> it makes other operations (addition modulo `2^w`, for instance, and shifting) so much more awkward

Ooh I'm sorry I think I was not clear: I meant to express only `padded_msg`, `concat_digest`, `sha256`, and the body of `M` in terms of `list byte`, but to still use `N` to represent "4 bytes", because indeed operations like eg `mod`, `>>`, `&`, `⊕` are much simpler on `N`, I was actually really happy to see how nice and concise your original spec using `N` everywhere was.
So my suggestion would be to use a hybrid approach, where the toplevel signature uses `list byte`, and the low-level operations are on `N`s representing 4 bytes, and at one point in the middle, conversion between `list byte` and `N` takes place. The question is then, at which level should this conversion take place? In https://github.com/project-oak/silveroak/pull/882, the conversion takes place at a very low level, but what about somewhere at a sweet spot in the middle? For instance, what if `M` takes a `padded_msg` that's a list of bytes, but returns an `N` representing 4 bytes?

jadephilipoom:
I think this could work. In a nutshell, we could do the following steps for `sha256`:
1. Compute `padded_msg` based on input data
2. Convert `padded_msg` from `list byte` to `list N`, calculate `M` based on this (since the `N`s would represent words, `M` would just index into the padded message)
2. Run the inner body (compression function loops)
3. Convert the final digest from `list N` to `list byte` and return

Sound like a plan?

samuelgruetter:
Yeah, that sounds good! 

Also, maybe [`coqutil.Word.LittleEndianList`](https://github.com/mit-plv/coqutil/blob/6eaedfbe93a8131d022ee7f20bafc1825361f8fe/src/coqutil/Word/LittleEndianList.v) could be useful, or a (yet-to-be-written) analog for big-endian, to convert between `Z` and list of bytes.

Another question would be whether to use `N` or `Z`: In bedrock2, we decided to use `Z` everywhere, even for nonnegative numbers, just to avoid conversions between `N` and `Z` needed when subtractions could become negative, eg conversions like in `Definition k := Z.to_N ((448 - (Z.of_N l + 1)) mod 512)%Z.` would not be needed any more if everything is on `Z`.

jadephilipoom:
I've also wondered about `N` vs `Z`, but since our bit-vectors are (at least at the moment) always treated as unsigned, it's so far made sense to me to always use `N`. The `k` definition is actually the first place I've ever had to use `Z`. It's sort of a tradeoff between reasoning about subtraction underflow and having lots of `0 <= x` preconditions, and my personal view is that right now the former is better than the latter. I think it would change my mind if we decided we wanted to represent signed bit-vectors in our DSL, though.

PR #873, opened 2021-08-06
Fix inner SHA256 circuit

blaxill:
SHA padded-input-to-digest implementation now "correct" (correct output on test vector), everything else still untested (read buggy).

### DIFF
--- a/investigations/cava2/Primitives.v
+++ b/investigations/cava2/Primitives.v
@@ -28,6 +28,15 @@ Inductive UnaryPrim : type -> type -> Type :=
 | UnVecSlice: forall {t n} (start len: nat), UnaryPrim (Vec t n) (Vec t len)
 | UnVecResize: forall {t n m}, UnaryPrim (Vec t n) (Vec t m)
 | UnVecReverse: forall {t n}, UnaryPrim (Vec t n) (Vec t n)
+| UnVecUncons: forall {t n}, UnaryPrim (Vec t (S n)) (t ** Vec t n)
+
+(* Our numbers are little endian but bitshifting is usually lexicographic
+ * ordered, that is for X = 0x3 = 0b110, X >> 1 = 0x1 = 0b100
+ * Whereas vector shifting is opposite X = [a,b,c], shift_right X 1 = [_,a,b]
+ * TODO(blaxill): simplify this situation?
+ *)
+| UnBitVecRotateRight: forall {t n}, nat -> UnaryPrim (Vec t n) (Vec t n)
+| UnBitVecShiftRight: forall {t n}, nat -> UnaryPrim (Vec t n) (Vec t n)
 
 | UnVecRotateRight: forall {t n}, nat -> UnaryPrim (Vec t n) (Vec t n)
 | UnVecShiftRight: forall {t n}, nat -> UnaryPrim (Vec t n) (Vec t n)
@@ -86,12 +95,27 @@ Fixpoint rotate_left {A} n (ls: list A): list A :=
     | x :: xs => rotate_left n' (xs ++ [x])
     end
   end.
+Fixpoint shift_left {A} n (ls: list (simple_denote_type A)): list (simple_denote_type A) :=
+  match n with
+  | 0 => ls
+  | S n' =>
+    match ls with
+    | nil => []
+    | x :: xs => shift_left n' (xs ++ [simple_default])
+    end
+  end.
 
 Fixpoint rotate_right {A} n (ls: list (simple_denote_type A)): list (simple_denote_type A) :=
   match n with
   | 0 => ls
   | S n' =>
     rotate_right n' (last ls simple_default :: removelast ls)
+  end.
+Fixpoint shift_right {A} n (ls: list (simple_denote_type A)): list (simple_denote_type A) :=
+  match n with
+  | 0 => ls
+  | S n' =>
+    shift_right n' (simple_default :: removelast ls)
   end.
 
 Definition unary_semantics {x r} (prim: UnaryPrim x r)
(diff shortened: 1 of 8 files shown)


investigations/cava2/Primitives.v
```diff
@@ -107,10 +131,20 @@ Definition unary_semantics {x r} (prim: UnaryPrim x r)
...
   | @UnVecReverse t n =>
     via_simple (b:=Vec t n) (a:= Vec t _) (fun x => rev x)%list
+
+  | @UnVecUncons t n =>
+    via_simple (b:=t ** Vec t n) (a:= Vec t _) (fun x => (hd simple_default x, tl x) )%list
+
+  | UnBitVecRotateRight n =>
+    via_simple (b:=Vec _ _) (a:= Vec _ _) (fun x => rotate_left n x)
+  | UnBitVecShiftRight n =>
+    via_simple (b:=Vec _ _) (a:= Vec _ _) (fun x => shift_left n x)
```
samuelgruetter, line 141: If I understood correctly your comment in `Primitives.v`, here you intentionally implement an operation called "shift right" with a function called "shift left", right? 😅 
It might indeed be good to simplify this so that we don't need separate shift operators for vectors and numbers. Maybe by modifying the direction in `simple_denote_to_denote` and `denote_to_simple_denote`?

investigations/cava2/Sha256.v
```diff
@@ -127,27 +201,89 @@ Section Var.
...
+  Goal
+    fst (snd (snd (simulate' sha256_inner (repeat ([], (false, ([], tt))) 64) (fst init)))) =
+    [0x506E3058; 0xD39A2165; 0x04D24D6C; 0xB85E2CE9; 0x5EF50F24; 0xFB121210; 0x948D25B6; 0x961F4894].
+  Proof. vm_compute. reflexivity. Qed.
+
+  (* Output is the correct message digest *)
+  Goal
+    fst (last (fst (simulate' sha256_inner (repeat ([], (false, ([], tt))) 64) (fst init))) default) =
+    [0xBA7816BF; 0x8F01CFEA; 0x414140DE; 0x5DAE2223; 0xB00361A3; 0x96177A9C; 0xB410FF61; 0xF20015AD ].
+  Proof. vm_compute. reflexivity. Qed.
```
samuelgruetter, line 287: Cool to see that this works!
On the other hand, what's the status on the netlist generation side? Can you generate code that Verilator and other hardware tools understand and can simulate and synthesize?

samuelgruetter (approved):

investigations/cava2/Sha256.v
blaxill: Not done yet, the most complex part will be marshalling the types, but having been through it a few times I don't expect it to take too long or be too difficult.

investigations/cava2/Primitives.v
blaxill: Hmm, I tried this out but it of course flips the indexing operations. That is `x=0xF0`, `x[0]` is `false`, whereas its expected to be little endian i.e. `x[0]` should be `true`. 

The confusion comes from in that SystemVerilog, little endian and big endian are both "represented" in lexical order, it is only the indexing that changes:
```verilog
wire [7:0] wires_little_endian;
wire [0:7] wires_big_endian;
/* Representation
wires_little_endian = [7][6][5][4][3][2][1][0]
wires_big_endian = [0][1][2][3][4][5][6][7]
*/
```

Whereas we usually thing about cons lists/vectors `[0]([1]([2] ...` for both. 

```
[

Comment (samuelgruetter) [reply investigations/cava2/Primitives.v:141]: Aaah so it seems that if we want to keep the standard way of representing lists, ie index 0 of `a :: b :: c :: d :: nil` is `a`, little and big endian order both are unintuitive:

- In little endian, `8` would be represented as `false :: false :: false :: true :: nil`, and in order to do a numeric right shift, we have to do a left shift on the list.
- In big endian, `8` would be represented as `true :: false :: false :: false :: nil`, but if you ask "give me bit 3 of this number", you'd expect to get the bit of weight `2^3`, but you'd get the last one in the list, which is the one of weight `2^0`.

So we can only choose between bad solutions 😅

File: investigations/cava2/Sha256.v [hunk 3]
```diff
@@ -23,57 +23,133 @@ Require Import Cava.Types.
...
+        else if state == `padder_waiting` && final then length + `resize` final_byte_length
+        else if state == `padder_writing_length` && current_offset == `K 62`
+          then `K 0`
+        else length
+        in
+
+      let next_out :=
+        if state == `padder_waiting` && final then
+          if final_byte_length == `K 0`
+          then `Constant (HexString.to_N ("0x00000001"): denote_type (BitVec 32))`
```
jadephilipoom, line 88: I think this can just be:
```suggestion
          then `Constant (0x00000001): denote_type (BitVec 32))`
```
now that Coq has hexadecimal parsing!

investigations/cava2/Sha256.v
```diff
@@ -127,27 +201,89 @@ Section Var.
...
+      : denote_type (sha_digest ** sha_digest ** sha_block ** Bit ** sha_round )) in
 
-    (current_digest, done)
+    let updated_digest := `map2 {{fun a b => ( a + b ) }} ` stored_digest current_digest in
+    (updated_digest, done)
 
   }}.
 
+  (* SHA-256 core *)
+  (* TODO(FIXME): *)
```
jadephilipoom, line 228: Should this be removed?

jadephilipoom (approved):

investigations/cava2/Primitives.v
jadephilipoom: Could we resolve this conflict by breaking the connection between Coq boolean vectors and our bit-vectors? Let's say we changed the type system such that there's a `BitVec` type directly, and it gets interpreted as `N`. The `Bit` type is replaced by `BitVec 1`, so now `Vec Bit` doesn't exist. We provide library functions to construct `BitVec`s from `N`, but not from a vector of booleans, and vector operations like `append` or `map2` are defined only on higher-level vectors (you can append a two vectors of bit-vectors, but not append two bit-vectors).

Then, we could define our shifting on bit-vectors to match what you'd expect from numeric right/left shifts, and indexing could be defined _separately_ for higher-level vectors and bit-vectors. For bit-vectors we could give the bit with weight `2^i`, but for higher-level vectors we could give the standard functional-programming interpretation.

One downside of this approach is that if what you want is logically a vector of booleans instead of a numeric value, you'll have to reason about it like a number. I suppose we _could_ have both `BitVec` and `Vec Bit`, where numeric operations are defined on the former and vector operations are defined on the latter, but that might also be confusing.

PR #878, opened 2021-08-09
Robust testing for HMAC and SHA-256

### LINKED ISSUE #876: Tests for HMAC
Create some more robust tests for the HMAC specification and circuits, such that the same test vectors can be used for both. This will provide quick feedback during development. Ideally, there should also be some step-by-step tests that compute intermediate values, to help with debugging.

jadephilipoom:
Resolves #876 

- Transcribes test vectors from NIST documents for HMAC and SHA-256, in such a way that intermediate values can also be checked for debugging purposes
- Runs the tests on SHA-256 and HMAC specifications
- Adjusts the HMAC specification to handle keys with a length greater than `B`

### DIFF
--- a/silveroak-opentitan/hmac/Spec/HMAC.v
+++ b/silveroak-opentitan/hmac/Spec/HMAC.v
@@ -14,6 +14,7 @@
 (* limitations under the License.                                           *)
 (****************************************************************************)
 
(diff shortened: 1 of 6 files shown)


blaxill (approved):
Lgtm! #880

PR #881, opened 2021-08-10
Proofs about SHA-256 spec

jadephilipoom:
The SHA-256 spec stipulates that the `k` parameter be the smallest possible solution to a particular equation, but doesn't give an explicit formula. I've proven here that the formula I used in the spec produces the smallest possible solution. The proofs are pretty ugly because of a lack of automation for `Z.to_N` and `Z.of_N`, but since we don't really do other proofs with `Z` I'm not sure it's worth the added effort to change that.

### DIFF
--- a/silveroak-opentitan/hmac/Spec/SHA256Properties.v
+++ b/silveroak-opentitan/hmac/Spec/SHA256Properties.v
@@ -18,58 +18,22 @@ Require Import Coq.Lists.List.
 Require Import Coq.NArith.NArith.
 Require Import Coq.micromega.Lia.
 Require Import Coq.ZArith.ZArith.
-Require Import coqutil.Tactics.Tactics.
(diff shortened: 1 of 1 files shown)


blaxill (approved):

samuelgruetter:
In the past few years, `lia` became much more powerful: it does not choke any more on goals produced by `Z.to_euclidean_division_equations`, and also, `Z.to_euclidean_division_equations` made it from "fiat-crypto wizardry" to "part of Coq's standard library". This allows us to turn all three proofs into one-liners. I took the liberty to push directly to the branch in @jadephilipoom's fork, so that my change appears as part of this PR, hope thats fine.

samuelgruetter:
@googlebot I consent.

jadephilipoom:
Oh awesome, thanks! :+1:

PR #888, opened 2021-08-12
Uart semantics and proofs (cont'd)

dayeol:
This completes the spec & proof of the following functions (#889)
- `uart_putchar`
- `uart_tx_full`

The function has 2 while loops: at the beginning to make sure the tx FIFO is not full, and at the end to ensure the device is back to IDLE state.
For now, it has been proven with the precondition where the device state is always IDLE when we enter the function.
This makes the first loop iterate exactly once.
It makes sense in a single-threaded environment because the only way to make the device busy is via `uart_putchar`, 
but the function waits the device to be IDLE state before it returns.
However, we could also generalize it by relaxing the precondition to be `execution tr s` for any state `s`,
and it would be pretty straightforward to prove that too -- `s` can be destructed, and if `s` is `IDLE` we just do the same thing, and if `s` is `BUSY` then we can prove that at some point the loop should break leaving the device in `IDLE` state.

### DIFF
--- a/firmware/IncrementWait/IncrementWait.v
+++ b/firmware/IncrementWait/IncrementWait.v
@@ -8,6 +8,7 @@ Require Import bedrock2.ToCString.
 Require Import coqutil.Word.Interface.
 Require Import Bedrock2Experiments.IncrementWait.Constants.
 Require Import Bedrock2Experiments.LibBase.MMIOLabels.
+Require Import Bedrock2Experiments.LibBase.AbsMMIO.
 Import Syntax.Coercions List.ListNotations.
 Local Open Scope string_scope.
 Local Open Scope Z_scope.
@@ -31,12 +32,16 @@ Section Impl.
     ("put_wait_get",
      ([input], [out], bedrock_func_body:(
         (* write input value *)
-        output! WRITE32 (VALUE_ADDR, input) ;
+        (* IDLE *)
+        abs_mmio_write32 (VALUE_ADDR, input) ;
         (* initialize status to 0 *)
         status = 0 ;
+        (* BUSY input ncycles_processing *)
         (* wait for status to be DONE *)
         while ((status & (1 << STATUS_DONE)) == 0) {
-          io! status = READ32 ( STATUS_ADDR )
+          (* BUSY O *)
+          unpack! status = abs_mmio_read32 ( STATUS_ADDR )
+          (*  *)
         };
         (* read the value and exit *)
         io! out = READ32 ( VALUE_ADDR )
--- a/firmware/IncrementWait/IncrementWait.v
+++ b/firmware/IncrementWait/IncrementWait.v
@@ -8,7 +8,6 @@ Require Import bedrock2.ToCString.
 Require Import coqutil.Word.Interface.
 Require Import Bedrock2Experiments.IncrementWait.Constants.
 Require Import Bedrock2Experiments.LibBase.MMIOLabels.
-Require Import Bedrock2Experiments.LibBase.AbsMMIO.
 Import Syntax.Coercions List.ListNotations.
 Local Open Scope string_scope.
 Local Open Scope Z_scope.
@@ -32,16 +31,12 @@ Section Impl.
     ("put_wait_get",
      ([input], [out], bedrock_func_body:(
         (* write input value *)
-        (* IDLE *)
-        abs_mmio_write32 (VALUE_ADDR, input) ;
+        output! WRITE32 (VALUE_ADDR, input) ;
         (* initialize status to 0 *)
         status = 0 ;
-        (* BUSY input ncycles_processing *)
         (* wait for status to be DONE *)
         while ((status & (1 << STATUS_DONE)) == 0) {
-          (* BUSY O *)
-          unpack! status = abs_mmio_read32 ( STATUS_ADDR )
-          (*  *)
+          io! status = READ32 ( STATUS_ADDR )
         };
         (* read the value and exit *)
         io! out = READ32 ( VALUE_ADDR )
--- a/firmware/Uart/Uart.v
+++ b/firmware/Uart/Uart.v
@@ -160,15 +160,19 @@ Section Impl.
     let cond := "cond" in
     ("b2_uart_putchar", ([byte], [],
     bedrock_func_body:(
-      unpack! cond = uart_tx_full();
+      cond = 1;
       while (cond == 1) {
         unpack! cond = uart_tx_full()
       };
+
       unpack! reg = bitfield_field32_write(0, UART_WDATA_WDATA_MASK, UART_WDATA_WDATA_OFFSET, byte);
       abs_mmio_write32(TOP_EARLGREY_UART0_BASE_ADDR + UART_WDATA_REG_OFFSET, reg);
-      unpack! cond = uart_tx_idle();
+      cond = 0;
+      (* BUSY ncycles_processing *)
       while (cond == 0) {
-        unpack! cond = uart_tx_idle()
+        (* s = BUSY O s = IDLE*)
+        unpack! cond = uart_tx_idle() (* cond = 0 *)
+        (* s = IDLE *)
       }
     ))).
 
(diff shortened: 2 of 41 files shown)


firmware/Uart/UartSemantics.v
```diff
@@ -48,6 +51,13 @@ Section WithParameters.
   | TIMEOUT_CTRL
   .
 
+  Axiom ncycles_not_zero :
+    Nat.eqb ncycles_processing 0 = false.
+  Axiom fifo_depth_not_zero :
+    Nat.eqb fifo_depth 0 = false.
+  Axiom ncycles_lt_max_ncycles :
+    Nat.ltb (ncycles_processing) (fifo_depth * ncycles_processing) = true.
```
jadephilipoom, line 59: These `Axiom`s will be impossible to instantiate later; if I were to prove that my `ncycles_processing` is nonzero, I can't then use that proof to make the UART programs' correctness proofs no longer depend on the axiom. If you instead introduce these with `Context`, though, the proofs will just show up with preconditions that can be elimated later.

jadephilipoom (changes requested):
The UART proofs look good, great work! However, I want to resolve the point about `IncrementWait` before merging.

firmware/Uart/UartProperties.v
```diff
@@ -217,4 +339,224 @@ Section Proofs.
...
+        end
+    end.
+
+
+  Global Instance spec_of_uart_putchar : spec_of "b2_uart_putchar" :=
+    fun function_env =>
+      forall (tr : trace) (m : mem) (R : _ -> Prop) (s : state) byte,
+      R m ->
+      (* uart must be in IDLE state before putchar *)
+      execution tr IDLE ->
```
samuelgruetter, line 360: Do you think it would be too hard to allow any `s` here?

firmware/IncrementWait/IncrementWait.v
```diff
@@ -31,12 +32,16 @@ Section Impl.
     ("put_wait_get",
      ([input], [out], bedrock_func_body:(
         (* write input value *)
-        output! WRITE32 (VALUE_ADDR, input) ;
+        (* IDLE *)
+        abs_mmio_write32 (VALUE_ADDR, input) ;
```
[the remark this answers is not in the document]
samuelgruetter: I agree that it would be good to keep `IncrementWait` small, and also without admitted lemmas like `status_read_always_ok` in `IncrementWaitProperties.v`. Is there much insight in the changes you did to `IncrementWait` that have not yet made it into `Uart`? If yes, we should probably copy the directory, but if no, it would be better to avoid the duplication, because every future change/refactoring will have to be applied to two instead of one directory. Alternatively, we could also keep it around on a separate branch, maybe?

samuelgruetter (changes requested):

firmware/IncrementWait/IncrementWait.v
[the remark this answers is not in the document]
dayeol: Good point, I think I can just omit `IncrementWait` from this PR as it was mostly practice.
Also I was kinda rushing when I submit this PR (cuz it's my second from the last day), so I did not spend much time to make `IncrementWait` look good. It'll be better to have a separate branch or directory and polish a bit if we are gonna include it.

firmware/Uart/UartSemantics.v
```diff
@@ -132,9 +140,12 @@ Section WithParameters.
     match r with
     | WDATA =>
         match s with
-        | IDLE => s' = BUSY 1
+        | IDLE => s' = BUSY (ncycles_processing)
             (* writing to 4-byte WDATA reg fills the FIFO with the last byte *)
-        | BUSY tx => if Nat.leb tx 32 then s' = BUSY (tx + 1) else False
+        | BUSY tx => if Nat.leb tx (fifo_depth * ncycles_processing)
+            then s' = BUSY (tx + (ncycles_processing))
```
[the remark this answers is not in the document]
dayeol: I think it needs to be `(fifo_depth - 1) * ncycles_processing` so that the maximum number of busy cycles is `fifo_depth * ncycles_processing`. I fixed it and also I'll try to write the invariant.

firmware/Uart/UartSemantics.v
dayeol: Thanks for the comments! But it turns out that we didn't even need the axioms, so I eradicated those.

firmware/Uart/UartSemantics.v
```diff
@@ -97,32 +107,30 @@ Section WithParameters.
...
-        | IDLE => status_matches_state s val = true /\ s' = IDLE
-        | BUSY O =>
-            (status_matches_state s val = true /\ s' = IDLE)
-        | BUSY txlvl =>
-            (status_matches_state s val = true /\ s' = BUSY (txlvl - 1))
+        | IDLE => status_matches_state s' val = true /\ s' = IDLE
+        | BUSY n =>
+            (* tx can non-deterministically finish sending *)
+            (status_matches_state s' val = true /\ s' = IDLE)
+            \/ (exists n', n = S n' /\ status_matches_state s' val = true /\ s' = BUSY n')
```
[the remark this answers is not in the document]
dayeol: fixed

firmware/Uart/UartProperties.v
dayeol: This is possible but I think it could be another huge change so can we make this a new PR?

firmware/Uart/UartSemantics.v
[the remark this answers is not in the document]
dayeol: Can we make this a new PR too? seems like it will need some time for me

dayeol:
@googlebot I signed it!

firmware/Uart/Uart.v
```diff
@@ -160,15 +160,19 @@ Section Impl.
...
       unpack! reg = bitfield_field32_write(0, UART_WDATA_WDATA_MASK, UART_WDATA_WDATA_OFFSET, byte);
       abs_mmio_write32(TOP_EARLGREY_UART0_BASE_ADDR + UART_WDATA_REG_OFFSET, reg);
-      unpack! cond = uart_tx_idle();
+      cond = 0;
+      (* BUSY ncycles_processing *)
       while (cond == 0) {
-        unpack! cond = uart_tx_idle()
+        (* s = BUSY O s = IDLE*)
+        unpack! cond = uart_tx_idle() (* cond = 0 *)
+        (* s = IDLE *)
```
samuelgruetter, line 175: these comments don't immediately make sense to me, are they just some sample run?

firmware/Uart/UartSemantics.v
```diff
@@ -97,32 +100,31 @@ Section WithParameters.
...
+        (if Nat.ltb t (fifo_depth * ncycles_processing)  then
+          negb (is_flag_set status UART_STATUS_TXFULL_BIT)
+        else
+          (* FIFO is full *)
+          is_flag_set status UART_STATUS_TXFULL_BIT))
     end.
 
   Definition read_step
     (s : state) (r : Register) (val : word) (s' : state) : Prop :=
+    status_matches_state s' val = true /\
```
samuelgruetter, line 119: I think the line `status_matches_state s' val = true` should be 2 lines further down, because it only needs to hold if we want to read the `STATUS` register. (But it doesn't matter (yet) because currently the `STATUS` register is the only readable register).

samuelgruetter (approved):

firmware/Uart/Uart.v
dayeol: removed!

samuelgruetter: did you also push it? (I can't see it here)

dayeol: no, not yet, I usually resolve the comments when I fix & test on my local. I'll push it soon.

firmware/Uart/UartSemantics.v
dayeol: fixed.

firmware/Uart/UartProperties.v
dayeol: #897

firmware/Uart/UartSemantics.v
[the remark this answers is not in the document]
dayeol: #898

samuelgruetter (approved):

samuelgruetter:
It seems that @jadephilipoom's as well as my requested changes have been addressed, or turned into new issues, so I'm going to merge this now, so that I can make my next PR on top of it.

samuelgruetter:
Thanks @dayeol for dealing with all my nitpicks 😉

PR #890, opened 2021-08-12
Prove sha256_compress correct

jadephilipoom:
As the first major circuit proofs using cava2, this also involved making some infrastructure improvements:
- Circuit type notations were globally `Open`ed in the `Types.v` file, meaning that anyone importing that file (even transitively) couldn't disable them. They directly conflict with list notations, so this was problematic. Instead, I've used `Bind Scope` to make Coq automatically interpret things it knows are `type`s in `circuit_type_scope`. The previous behavior can be recreated using `Local Open Scope circuit_type_scope`, but in practice the `Bind Scope` was enough in most places.
- Moved the various primitive circuits out of `PrimitiveNotations` so that it's possible to not import `PrimitiveNotations` (in proofs, I find circuit notations are unhelpful because I just translate it all into Gallina immediately anyway) and not have everything prefixed by `PrimitiveNotations.`

### DIFF
--- a/investigations/cava2/Expr.v
+++ b/investigations/cava2/Expr.v
@@ -49,7 +49,7 @@ Section Vars.
     -> Circuit (s1++s2) [] x
   | ElimPair: forall {x y z s},
     (var x -> var y -> Circuit s [] z)
-    -> var (x**y)
+    -> var (x**y)%circuit_type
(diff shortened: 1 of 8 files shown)


blaxill (approved):
Lgtm!

PR #891, opened 2021-08-12
Fix bug in SHA-256 spec

jadephilipoom:
#890 introduced a bug (`N.add` instead of `add_mod`). It wasn't caught by the SHA-256 test vectors (I guess none of them had an overflow there) but was caught by the HMAC tests, so I've added an extra test case to the SHA-256 tests that tests with the particular input that triggered the bug so we catch it earlier.

This wasn't caught by CI because it appears CI is not actually running the things under `silveroak-opentitan/hmac`. We should fix that. Thanks to Ben for pointing out to me that the HMAC tests were failing!

### DIFF
--- a/cava/Cava/Util/String.v
+++ b/cava/Cava/Util/String.v
@@ -25,3 +25,10 @@ Definition string_to_bytes (s : string) : list Byte.byte :=
 
 Definition string_to_N (s : string) : N :=
   BigEndianBytes.concat_bytes (string_to_bytes s).
+
+Definition bytes_to_string (bs : list Byte.byte) : string :=
+  string_of_list_ascii (map ascii_of_byte bs).
+
+(* n is the number of bytes expected *)
+Definition N_to_string (n : nat) (s : N) : string :=
+  bytes_to_string (BigEndianBytes.N_to_bytes n s).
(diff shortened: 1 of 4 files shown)


blaxill (approved):

samuelgruetter:
This raises an interesting question: Why did using `N.add` instead of `add_mod` lead to a bug at all? Shouldn't it be the case that all operations that treat `N`s as 4-byte values ignore the bits of index >= 32? (in the sense that presence or absence of bits above index 32 does not affect the bits of index < 32, but might still affect what kind of "garbage" is found at indices >= 32). And I think the answer is (without testing it, though), that the only operation that violates this is `>>`, so my guess is that if you defined `>>` as

```coq
Definition fixed_width_right_shift (x n : N) := N.shiftr (x mod (2 ^ w)) n.
```

then using `N.add` instead of `add_mod` would not have mattered.

It might be useful to use this definition of `>>` instead, because stating and proving properties could be simpler if bits at indices >=32 are always ignored, what do you think?

blaxill:
Sam makes an interesting point, in the meantime I will merge this so `hmac/` build succeeds so I can perform CI updates.

jadephilipoom:
Definitely an interesting idea. If we just ignore the high bits, proof statements definitely become cleaner because we won't have preconditions on the bounds (instead, everything will just be in terms of `x mod 2^w`). I'm halfway through some proofs based on the current version, so maybe after I finish those I'll try switching to ignoring high bits and we can directly compare the effort :slightly_smiling_face:

PR #892, opened 2021-08-12
Add hmac directory to CI

blaxill:
Also:
- Cache cava/
- Simplify CI script

### DIFF
--- a/.github/workflows/build_all.yaml
+++ b/.github/workflows/build_all.yaml
@@ -85,5 +85,11 @@ jobs:
       if: steps.third_party-cache.outputs.cache-hit != 'true' || steps.cava-cache.outputs.cache-hit != 'true'
       run: exit 1
 
-    - run: nix-shell --run "cd $BUILD_DIR; make -j1"
+    - run: |
+        ls
+        echo $BUILD_DIR
+        ls $BUILD_DIR
+        cd $BUILD_DIR
+        nix-shell --run 'make -j1'
+        # nix-shell --run "cd $BUILD_DIR; make -j1"
 
--- a/.github/workflows/build_all.yaml
+++ b/.github/workflows/build_all.yaml
@@ -20,28 +20,41 @@ jobs:
       run: |
         git submodule status > ./third_party/submodules.lock
       shell: bash
+
     - name: Cache third_party
       id: third_party-cache
       uses: actions/cache@v2
       with:
         path: third_party
         key: ${{ runner.os }}-third_party-${{ hashFiles('third_party/submodules.lock') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_all.yaml
```diff
@@ -130,177 +75,15 @@ jobs:
...
+      id: cava-cache
       uses: actions/cache@v2
       with:
-        path: third_party
-        key: ${{ runner.os }}-third_party-${{ hashFiles('third_party/submodules.lock') }}
+        path: cava
+        key: ${{ runner.os }}-cava-${{ hashFiles('third_party/submodules.lock', 'cava/**/*.v', 'cava/Makefile' ) }}
+    - name: Fail on cache miss
+      if: steps.third_party-cache.outputs.cache-hit != 'true' || steps.cava-cache.outputs.cache-hit != 'true'
+      run: exit 1
```
samuelgruetter, line 86: Do I understand correctly that this `Fail on cache miss` step is just a safety check that should really always succeed, like an assert?

.github/workflows/build_all.yaml
```diff
@@ -20,95 +20,41 @@ jobs:
...
+    strategy:
+      matrix:
+        build_dir: ["demos", "examples", "tests"
+        , "silveroak-opentitan/pinmux"
+        , "silveroak-opentitan/aes"
+        , "silveroak-opentitan/hmac"
+        , "examples/xilinx"
+        , "tests/xilinx"
+        , "firmware"
+        ]
```
samuelgruetter, line 54: If I understand correctly, `matrix` might run all its task in parallel, in an unspecified order. At the moment, that's fine, but soon, the `firmware` directory will reference `silveroak-opentitan/hmac/Spec/SHA256.v`, so that we can state a theorem saying that the driver function outputs a digest computed according to the definition of `sha256` in that file. At that point, we'll need to make sure that the `firmware` directory is only built once the `silveroak-opentitan/hmac` has finished building.
But since that dependency is not yet present, I'd be also ok with merging this PR as-is.

samuelgruetter (approved):

.github/workflows/build_all.yaml
samuelgruetter: Also, the same problem already exists with `investigations/cava2/Sha256Properties.v`, but since that directory is not run on CI, it's "fine" as well at the moment :wink:

.github/workflows/build_all.yaml
blaxill: Yes exactly, its probably unneeded but no harm in having it.

Issue #893, opened 2021-08-12
Missing directory dependencies on CI


samuelgruetter:
The end-to-end theorem that I'd like to sketch by the end of next week will introduce the following three dependencies which are currently not taken into account by the CI (where `<--` means "depends on"):

* `firmware/Hmac/HmacProperties.v` `<--` `silveroak-opentitan/hmac/Spec/SHA256.v`
* `investigations/cava2/Sha256Properties.v` `<--` `silveroak-opentitan/hmac/Spec/SHA256.v`
* A file with an end-to-end theorem about Sha256, either living in `firmware/Hmac` (similarly to `firmware/IncrementWait/IncrementWaitToCava.v`), or in a separate directory `<--` `investigations/cava2/`

To start, I can of course just build the right subdirectories locally & manually, but it would be nice if this was also supported on CI.

jadephilipoom:
I suggest:
- Move `investigations/cava2` out of `investigations` and make it a top-level directory that is run by CI
- Move the HMAC-specific circuit implementations out of `cava2` and move it into a new directory `silveroak-opentitan/hmac/hw` (since we eventually want `cava2` to be just the core library, it's probably good not to have a bunch of specific code there)
- Move the HMAC-specific firmware out of `firmware` and into a new directory `silveroak-opentitan/hmac/sw` (less sure of this one, but it might make sense to organize things by device for bigger projects like HMAC -- an unfortunate side effect is that the `run_silicon_creator_tests_in_opentitan` script will have to be copied)
- Move the end-to-end proofs into a new directory `silveroak-opentitan/hmac/end2end`
- Introduce the necessary CI dependencies:
  -  `silveroak-opentitan/hmac` `<--` `cava2`
  -  `silveroak-opentitan/hmac` `<--` `firmware`
- Remaining dependencies will be handled in `silveroak-opentitan/hmac/Makefile`:
  -  `silveroak-opentitan/hmac/hw` `<--` `silveroak-opentitan/hmac/Spec`
  -  `silveroak-opentitan/hmac/sw` `<--` `silveroak-opentitan/hmac/Spec`
  -  `silveroak-opentitan/hmac/end2end` `<--` `silveroak-opentitan/hmac/hw`
  -  `silveroak-opentitan/hmac/end2end` `<--` `silveroak-opentitan/hmac/sw`

[...]

PR #894, opened 2021-08-13
sha256 firmware proofs

samuelgruetter:
Here are my sha256 driver correctness proofs! They are almost admit-free, except for some uninteresting bitfiddling, `execution_unique`, the sha256 spec, and the assumption that the byte list returned by sha256 spec is always of length 32, which we'll be able to instantiate as soon as the `firmware` can depend on `silveroak-opentitan/hmac/Spec/SHA256.v` (see https://github.com/project-oak/silveroak/issues/893).

Given that the remaining admits look uninteresting and unlikely to reveal interesting incompatibilities, I'd like switch to working on the end-to-end theorem now (where interesting incompatibilities are more likely), and to merge the sha256 driver correctness, even though there are still some admits.


firmware/Hmac/HmacProperties.v
```diff
@@ -0,0 +1,1040 @@
...
+    cancel_seps_at_indices 0%nat 0%nat. 1: reflexivity.
+    cancel_seps_at_indices 1%nat 0%nat. {
+      f_equal.
+      rewrite List.app_length. cbn [List.nth List.length].
+      ZnWords.
+    }
+    ecancel_done.
+  Qed.
+
+  Lemma while_zero_iterations {functions} {e c t l} {m : mem} {post : _->_->_-> Prop}
```
jadephilipoom, line 222: Move to `firmware/WhileProperties.v`?

firmware/Hmac/HmacProperties.v
```diff
@@ -0,0 +1,1040 @@
...
+    - clear dependent m. intros m Hsep. eapply HPost.
+      SeparationLogic.seprewrite isolate_scalar32_of_bytarray. {
+        rewrite List.upds_length. subst n. exact HL.
+      }
+      use_sep_assumption.
+      cancel.
+      unfold List.upds. change (Z.to_nat \[addr' ^- addr]) with n.
+      cancel_seps_at_indices 0%nat 1%nat. {
+        f_equal.
+        (* TODO this kind of list simplification should be automated *)
```
jadephilipoom, line 343: If you import `Cava.Util.List` I believe this will go by `autorewrite with listsimpl push_skipn push_firstn push_length`.

jadephilipoom (approved):

firmware/Hmac/HmacProperties.v
samuelgruetter: It didn't work out-of-the-box, but after adding three more hints, it worked for the first list goal. However, in the second goal, I got stuck on [`autorewrite`'s limitation](https://github.com/coq/coq/issues/7672) that it does not backtrack to the next rewritable pattern if the sidecondition solver fails for the first one.

Here's an isolated example from the second list goal in this lemma:

```coq
Require Import Coq.Lists.List. Import ListNotations.
Require Import Coq.micromega.Lia.

Lemma firstn_eq_0 {A : Type} (n : nat) (l : list A) (HL : n = 0%nat) : List.firstn n l = [].
Proof. subst n. apply List.firstn_O. Qed.
Hint Rewrite @firstn_eq_0 using lia : push_firstn.

Goal forall {A: Type} (n: nat) (values: list A),
    firstn n values = firstn n values ++ firstn (n - n) values.
Proof.
  intros.

  (* doesn't do anything because after failing to prove n=0,
     it does not backtrack to the next pattern that matches the LHS of firtn_eq_0 *)
  autorewrite with push_firstn.

  (* workaround: *)
  remember (firstn n values) as dontTryToRewriteHere.
  autorewrite with push_firstn.
```

Once I had discovered https://github.com/coq/coq/issues/7672, I decided to stop using `autorewrite`, and instead use tactics of the form `repeat simplification_step`, where `Ltac simplification_step := match goal with |- context[...] => rewrite ... end`, which are also easier to debug because I can run one `simplification_step` at a time to see step-by-step what's happening.

jadephilipoom: What an annoying bug. We should probably rewrite the list hints to use manual matches instead of `autorewrite`, as you suggest.

firmware/Hmac/HmacProperties.v
samuelgruetter: Actually, I just moved it to `bedrock2.Loops`

PR #900, opened 2021-08-17
Fix multiblock SHA circuit

blaxill:
- Fixed incorrect padding for multiple blocks
- Switched padding done signal to latch (from single pulse)


jadephilipoom (approved):
This probably demonstrates a good reason to prioritize #880; it's far better to discover bugs like this in tests rather than in proofs, when it's harder to tease out whether there's actually an issue or whether there's something missing in your reasoning. Luckily, I hadn't tried to prove the padder yet!

jadephilipoom:
Merging this so I can build on top of it.

PR #901, opened 2021-08-17
Hmac update

blaxill:
HMAC appears to be working - we can set the key registers and stream data to the fifo and it produces the correct value in the digest registers.

It took me longer than intended to get the timing right :( 

Also
- Small simplification to sha256, that I can remove if annoying for current proof work
- Changed the order of realignment and fifo in comparison to OpenTitan as it was simpler to encode in Cava2
- There are some implicit assumptions such as not writing to fifo after CMD_PROCESS or when fifo is full (this should be obeyed by a correct operating host)


cava2/Fifo.v
```diff
@@ -35,67 +35,131 @@ Section Var.
...
+      else if data_mask == `K 0x1` then (data << 24, `K 1`)
+      else if data_mask == `K 0x2` then (data << 16, `K 1`)
+      else if data_mask == `K 0x4` then (data << 8,  `K 1`)
+      else if data_mask == `K 0x8` then (data     ,  `K 1`)
+      else if data_mask == `K 0x3` then (data << 16, `K 2`)
+      else if data_mask == `K 0x6` then (data << 8,  `K 2`)
+      else if data_mask == `K 0xC` then (data     ,  `K 2`)
+      else if data_mask == `K 0x9` then (data << 16, `K 3`)
+      else if data_mask == `K 0xE` then (data      , `K 3`)
+      (* else if data_mask == `K 0xF` then *) else (data, `K 4`)
```
jadephilipoom, line 85: It looks like the mask is a `BitVec 32`; isn't it possible that the mask could be larger than a byte? If not, why allocate 32 bits for it?

silveroak-opentitan/hmac/hw/Hmac.v
```diff
@@ -46,39 +183,110 @@ Section Var.
...
+      (tl_o, registers')
 
-      initially value_hole
+      initially default
     in tl_o
   }}.
 
 End Var.
+
+Section SanityCheck.
```
jadephilipoom, line 201: Would you mind using the HMAC test vectors instead of hardcoded examples? I realize it's a bit more work at the start, but it's going to be a lot easier to update, and they cover a wider range of cases so we can hopefully find bugs earlier than we did with SHA-256.

jadephilipoom (approved):

silveroak-opentitan/hmac/hw/Hmac.v
blaxill: I agree the test vectors are better for catching bugs, but these hardcoded versions allow me to inspect internal state values more easily. I'd like to keep this for now, but with the intention of the test vector version superseding in the future when I'm more sure there won't be changes.

cava2/Fifo.v
blaxill: In a previous version I had `32` bit wide masks, but the mask is only 4 bits here- `data` is the `BitVec 32`.

silveroak-opentitan/hmac/hw/Hmac.v
jadephilipoom: Okay, sounds fine with me as long as we run the more extensive tests before attempting proofs.

cava2/Fifo.v
jadephilipoom: Looks like I was off by one when counting arguments -- never mind!

PR #902, opened 2021-08-17
Generate cava2 _CoqProject automatically

jadephilipoom:
This matches the build setup for the rest of the repo, and will make it easier not to accidentally leave files out of the build as we add more things.


investigations/cava2/Makefile
```diff
@@ -14,14 +14,26 @@
 # limitations under the License.
 #
 
-.PHONY: all clean
+.PHONY: all clean coq
```
samuelgruetter, line 17: Currently, some makefiles have `_CoqProject` marked as `.PHONY`, while others don't. I think all of them should have it marked as `.PHONY`, because this will cause `_CoqProject` to be regenerated on each run, which is desirable because otherwise you have to delete (or `touch`) it each time you add, delete, or rename files.

samuelgruetter:
```
grep -R . --include='Makefile' --exclude-dir=third_party -l -e _CoqProject | xargs grep -e '.PHONY' | grep -v _CoqProject
```

lists all makefiles that mention `_CoqProject`, but don't mention it in `.PHONY`:

```
./demos/Makefile:.PHONY: all coq miminize-requires clean html
./investigations/acorn/Makefile:.PHONY: all clean
./investigations/kami/counter/Makefile:.PHONY: all clean configure
./investigations/kami/icestick-counter/Makefile:.PHONY: all clean configure
./investigations/modeling/Makefile:.PHONY: all coq clean
./investigations/acornish/Makefile:.PHONY: all clean
./investigations/cava2/Makefile:.PHONY: all clean
./firmware/Makefile:.PHONY: all coq clean aes.c
```

At some point (maybe with this PR, or separately), we should fix it everywhere?

blaxill (approved):

jadephilipoom:
> At some point (maybe with this PR, or separately), we should fix it everywhere?

Yes, at least for the ones that aren't under `investigations` -- I'll make an issue to address this.

PR #903, opened 2021-08-17
Proofs for sha256_inner circuit

jadephilipoom:
On top of #902; should not be merged until that one is.

This proof was pretty involved because it was the first one that involved reasoning about a nontrivial control flow; `sha256_inner` relies on a combination of flags to do essentially a few different computations, only some of which are meaningful. I settled on a 
 proof structure here that might be useful for future circuits:
- Define an invariant for the circuit state that produces a `Prop` given a state and ghost variables (e.g. `i`, `msg`)
- Define a precondition for the circuit input that produces a `Prop` given an input and ghost variables
- Define a "spec" with the exact output given a state and input
- Prove that, if the invariant holds on the state and the precondition holds on the input, the invariant holds on the new state from `step`
- Prove that, if the invariant holds on the state and the precondition holds on the input, the output from `step` matches the spec

One key feature here is that we don't have to specify exactly what garbage fills unused state variables during cycles when those variables are unused; trying to specify all the state exactly essentially required re-implementing the circuit in Gallina, which seems undesirable.

I've defined a bunch of stuff in `Util` that I needed for the proof; now that we're using lists instead of vectors, we have to beef up the list library a bit. ~I also added an `N.push_pull_mod` based on coqutil's `Z.push_pull_mod`, which might be worth upstreaming at some point.~ (edit: removed, this now works with `Z.to_euclidean_division_equations` and `lia` thanks to input from @samuelgruetter)


samuelgruetter:
> I also added an `N.push_pull_mod` based on coqutil's `Z.push_pull_mod`, which might be worth upstreaming at some point.

Did you try using `Z.to_euclidean_division_equations; lia` instead? (requires that the modulus is a constant, which I think is the case here, and might also need to be preceded by `Zify.zify`). I see that your `N.push_pull_mod` usage is not at the end of a goal, so it might require some restructuring, but I can't emphasize enough that `lia` recently became much more powerful at dealing with goals produced by `Z.to_euclidean_division_equations`, and we should take advantage of this to improve automation whenever we can.

blaxill (approved):
Great job, this actually came out cleaner so far than I'd hoped

jadephilipoom:
> Did you try using `Z.to_euclidean_division_equations; lia` instead? (requires that the modulus is a constant, which I think is the case here, and might also need to be preceded by `Zify.zify`). I see that your `N.push_pull_mod` usage is not at the end of a goal, so it might require some restructuring, but I can't emphasize enough that `lia` recently became much more powerful at dealing with goals produced by `Z.to_euclidean_division_equations`, and we should take advantage of this to improve automation whenever we can.

Got it working with `zify` and a little computation!

investigations/cava2/Sha256Properties.v
```diff
@@ -57,13 +46,226 @@ Hint Rewrite @step_rotr using solve [eauto] : stepsimpl.
...
+                         rewrite <-H end.
+  cbv [SHA256.add_mod SHA256.w]. apply f_equal.
+  cbv [SHA256.sigma1 SHA256.sigma0 SHA256.SHR].
+  cbv [N.of_nat Pos.of_succ_nat Pos.succ]. clear.
+  (* fully compute moduli *)
+  repeat match goal with |- context [(_ mod ?m)%N] =>
+                         progress compute_expr m end.
+  (* convert to Z, solve with Z.div_mod_to_equations *)
+  zify.
+  repeat rewrite Z.rem_mod_nonneg; Z.div_mod_to_equations; lia.
```
samuelgruetter, line 82: I was wondering why you did `repeat rewrite Z.rem_mod_nonneg; Z.div_mod_to_equations; lia` instead of just `Z.to_euclidean_division_equations; lia`, and when I tried it out, I noted that it's probably because of performance: the former only takes `0.236 secs`, whereas the latter takes `0.822 secs`.

jadephilipoom: Actually, the latter fails for me (Coq 8.13.2); `lia` says it can't solve the goal.

samuelgruetter: Interesting, I also have 8.13.2... Did you keep the `zify`?

jadephilipoom: Yep. From current main, with diff:
```diff
diff --git a/investigations/cava2/Sha256Properties.v b/investigations/cava2/Sha256Properties.v
index 19205708..37898b4b 100644
--- a/investigations/cava2/Sha256Properties.v
+++ b/investigations/cava2/Sha256Properties.v
@@ -79,7 +79,8 @@ Proof.
                          progress compute_expr m end.
   (* convert to Z, solve with Z.div_mod_to_equations *)
   zify.
-  repeat rewrite Z.rem_mod_nonneg; Z.div_mod_to_equations; lia.
+  (*
+  repeat rewrite Z.rem_mod_nonneg; *) Z.div_mod_to_equations; lia.
 Qed.
 Hint Rewrite @step_sha256_message_schedule_update using solve [eauto] : stepsimpl.
```
I get:
```
File "./Sha256Properties.v", line 83, characters 62-65:
Error: Tactic failure:  Cannot find witness.
```

samuelgruetter: Did you notice that `Z.div_mod_to_equations ≠ Z.to_euclidean_division_equations`? The latter deals not only with `div` and `mod`, but also with `quot` and `rem`.

jadephilipoom: Aha, no, my eye skipped over the difference! Mystery solved. I'll remember that for the future, since it seems like for this particular PR the performance gain might be worth it.

PR #904, opened 2021-08-17
Redo the increment device in Cava2.

fshaked:


firmware/IncrementWait/CavaIncrementDevice.v
```diff
@@ -1,113 +1,127 @@
...
+  Section Cava.
+    Definition circuit_state {i s o} (c : Circuit s i o) := denote_type s.
+  End Cava.
+
+  Definition incr
+    : Circuit _
+              (* Input *)
+              [ Bit          (* is_read_req *)
+                ; Bit        (* is_write_req *)
+                ; BitVec 32  (* req_addr (only relevant if is_reaq_req or is_write_req) *)
```
samuelgruetter, line 33:
```suggestion
                ; BitVec 32  (* req_addr (only relevant if is_read_req or is_write_req) *)
```
(originally my typo)

firmware/IncrementWait/CavaIncrementDevice.v
```diff
@@ -1,113 +1,127 @@
...
+                             ; (1%N,  (0%N, (  4%N, (111%N, tt)))) (* read STATUS *)
+                             ; (1%N,  (0%N, (  4%N, (111%N, tt)))) (* read STATUS *)
+                             ; (1%N,  (0%N, (  0%N, (111%N, tt)))) (* read VALUE *)
+                           ].
+
+
+Definition bool_to_dbit (b : bool) : N :=
+  if b then 1%N else 0%N.
+Definition dbit_to_bool (b : N) : bool :=
+  if N.eqb b 0%N then false else true.
```
samuelgruetter, line 81: Why not use `NArith.BinNat.N.b2n` and `Cava.Util.BitArithmetic.n2bool`?

firmware/IncrementWait/CavaIncrementDevice.v
```diff
@@ -1,113 +1,127 @@
...
+
+Definition incr_device_step s (input : bool * bool * N * N) : (_ * (bool * N)) :=
+  let '(is_read_req, is_write_req, req_addr, req_value) := input in
+
+  let '(s', (is_resp, resp)) := step incr s (bool_to_dbit is_read_req,
+                                             (bool_to_dbit is_write_req,
+                                              (req_addr, (req_value, tt)))) in
+  (s', (dbit_to_bool is_resp, resp)).
```
samuelgruetter, line 89: Seeing all these `bool_to_dbit` and `dbit_to_bool` conversions makes me wonder whether `Notation Bit := (BitVec 1)` in Cava2 was really a good idea... Why not add back a separate `Bit` type that gets denoted to `bool`? (/cc @jadephilipoom and @blaxill)

firmware/IncrementWait/CavaIncrementDevice.v
```diff
@@ -1,113 +1,127 @@
...
+(*   (* : list (circuit_state c * o) := *) *)
+(*   := *)
+(*   match inps with *)
+(*   | [] => [] *)
+(*   | inp :: rest => let r := step c initial (denote_type inp) in *)
+(*                  r :: simulate_with_state_rec c (fst r) rest *)
+(*   end *)
+(* . *)
+(* Definition simulate_with_state{i o}(c: Circuit i o): list i -> list (circuit_state c * o) := *)
+(*   simulate_with_state_rec c (reset_state c). *)
```
samuelgruetter, line 118: I don't think commented-out code is a good idea, because as more changes are made, it will quickly rot and soon not compile any more. I'd fix it & uncomment it (or delete it).

firmware/IncrementWait/CavaIncrementDevice.v
```diff
@@ -137,28 +156,33 @@ Section WithParameters.
...
   | BUSY_done_related: forall val ncycles,
       counter_related (BUSY val ncycles)
-                      (mk_counter_state false false true (word_to_bv (word.add (word.of_Z 1) val)))
+                      (mk_counter_state false 0 true (word_to_N (word.add (word.of_Z 1) val)))
   | DONE_related: forall val,
-      counter_related (DONE val) (mk_counter_state false false true (word_to_bv val)).
+      counter_related (DONE val) (mk_counter_state false 0 true (word_to_N val)).
 
-  Axiom TODO: False.
+  Axiom TODO: Coq.Init.Logic.False.
```
samuelgruetter, line 179: Does `False` here still refer to Cava's `False`? That doesn't sound right, can we put it into a `Notations` module that gets only imported inside sections where we really want it?

firmware/IncrementWait/CavaIncrementDevice.v
```diff
@@ -177,29 +201,26 @@ Section WithParameters.
...
+        apply BUSY_done_related.
       + (* BUSY_done_related: *)
-        cbn. rewrite andb_false_r. cbn. assumption.
+        change (0%N) at 1 with (bool_to_dbit false).
+        change (1%N) with (bool_to_dbit true).
+        apply BUSY_done_related.
       + (* DONE_related: *)
-        cbn. rewrite andb_false_r. cbn. assumption.
+        change (0%N) at 1 with (bool_to_dbit false).
+        change (1%N) with (bool_to_dbit true).
```
samuelgruetter, line 222: More code suggesting that we should consider again a separate `Bit` type that gets denoted to `bool`, /cc @jadephilipoom @blaxill

samuelgruetter (approved):

samuelgruetter:
Looks great! (And I think the CI failure is due to https://github.com/project-oak/silveroak/issues/893)

firmware/IncrementWait/CavaIncrementDevice.v
```diff
@@ -1,113 +1,127 @@
...
+  Import ExprNotations.
+  Import PrimitiveNotations.
+
+  Context {var : tvar}.
+
+  Local Open Scope N.
+
+  (* Things that should probably be in cava *)
+  Section Cava.
+    Definition circuit_state {i s o} (c : Circuit s i o) := denote_type s.
```
jadephilipoom, line 25: #905 will add something like this to `Expr.v` when it lands (called `state_of`)

firmware/IncrementWait/CavaIncrementDevice.v
fshaked: `bool_to_dbit` and `dbit_to_bool` are not really necessary, I just didn't want to change the device class while cava2 was changing. The bool/number distinction doesn't affect the proof below much (though it might for other proofs). That said, it feels weird to use N instead of bool.

firmware/IncrementWait/CavaIncrementDevice.v
fshaked: Done

firmware/IncrementWait/CavaIncrementDevice.v
fshaked: Removed

firmware/IncrementWait/CavaIncrementDevice.v
fshaked: Renamed to make the transition easier

PR #905, opened 2021-08-17
Run SHA-256 test vectors on circuit

jadephilipoom:
For #880 

Attempting to run the SHA-256 test vectors on the circuit. Two problems I'd like help on (@blaxill ?):
- I can't figure out how long the circuit is expected to run; test1 ("abc", 1 block, 3 bytes) takes `64*1+17` cycles, but test2 (2 blocks, 14 bytes) takes `64*2+21`. For bigger tests I can't tell, because...
- When I try to run test3, the circuit doesn't ever switch the "done" flag to 1, even when run for 500 cycles, and as far as I can tell doesn't ever get to the correct answer. I also tried running a 4-block test (the 4-byte word derived from "abcd" repeated 48 times) with the same result; it didn't ever seem to finish the computation.

### DIFF
--- a/investigations/cava2/Expr.v
+++ b/investigations/cava2/Expr.v
@@ -63,6 +63,10 @@ Section Vars.
   .
 End Vars.
 
+Definition state_of {var s i o} (c : @Circuit var s i o) : type := s.
+Definition input_of {var s i o} (c : @Circuit var s i o) : type := i.
+Definition output_of {var s i o} (c : @Circuit var s i o) : type := o.
+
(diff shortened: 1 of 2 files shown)


jadephilipoom:
I am marking this "ready for review" because it will be affected by the reorganization discussed in #893 and the failing test is currently commented out, so it's probably easiest to just merge it now and then uncomment the test once the bug is fixed. Also, fixing the bug might be easier on top of this PR!

blaxill (approved):

PR #908, opened 2021-08-18
Fix and comment SHA256 circuit

blaxill:


investigations/cava2/Sha256.v
```diff
@@ -440,4 +467,60 @@ Section SanityCheck.
...
+       ; 0x1A1B1819
+       ; 0x06070405
+       ; 0x02030001
+       ; 0x0E0F0C0D
+       ; 0x0A0B0809
+       ])
+
+       (* Note we are not observing the sha block outputs, but if we were
+          we would see that it is not immediately ready to receive input here *)
+       ++ repeat (0, (0, (0, (0, (0, tt))))) 80
```
blaxill, line 496: cc @jadephilipoom

jadephilipoom: Is this what was causing the failure on the test3 test vector? Do you know why doesn't it happen with the other 2-block test?

jadephilipoom (approved):

blaxill:
> Is this what was causing the failure on the test3 test vector?

test3 would have failed anyway as there was a bug in the core which is now fixed.

> Do you know why doesn't it happen with the other 2-block test?
 
It doesn't happen in other 2-block test, because the provided _input_ sequence fits in the first block, the second block contains only the padding&length from the padder which schedules itself correctly.

PR #913, opened 2021-08-19
Complete SHA-256 circuit tests

jadephilipoom:
For #880 

All SHA-256 tests, including the ones that test intermediate digests and only-padding, are now run on the cava2 SHA-256 implementation. I removed the "sanity check", since it's a bit annoying to update when something about the signalling changes and is now superseded by the tests file.

Part of the code here is a boolean list equality decider that I've submitted upstream to coqutil (see https://github.com/mit-plv/coqutil/pull/37), so hopefully that will get merged and the code will no longer live in our test file.


blaxill (approved):

silveroak-opentitan/hmac/hw/Sha256Tests.v
```diff
@@ -26,10 +30,55 @@ Require Import HmacSpec.Tests.SHA256TestVectors.
 Require Import HmacHardware.Sha256.
 Import ListNotations.
 
+(**** list boolean equality decider ****)
+
+(* TODO: this may be upstreamed, remove if
+   https://github.com/mit-plv/coqutil/pull/37 gets merged *)
```
samuelgruetter, line 36: the "gets merged" has happened, happy recompiling :wink:

silveroak-opentitan/hmac/hw/Sha256Tests.v
```diff
@@ -69,8 +133,65 @@ Definition from_sha256_output
...
+           else [data]) out in
+  concat_words (N.to_nat w) valid_out.
+
+(* extract all intermediate digests from circuit output signals *)
+Definition intermediate_digests_from_sha256_output
+           (out : list (denote_type (output_of (var:=denote_type) sha256)))
+  : list (list N) :=
+  let digests := List.map (fun '(done, (digest, accept_input)) => digest) out in
+  (* remove repetitions of the same digest *)
+  dedup (list_eqb N.eqb) digests.
```
samuelgruetter, line 168: Could it be possible to get the same digest twice, which would make this code wrong? I'd estimate that the probability of this is roughly in the same order of magnitude as the probability of finding hash collisions, ie in the range of don't care, but my formal correctness mindset just couldn't refrain from commenting here... :wink:

silveroak-opentitan/hmac/hw/Sha256Tests.v
jadephilipoom: Thanks! On it :slightly_smiling_face:

silveroak-opentitan/hmac/hw/Sha256Tests.v
jadephilipoom: Yes, my formal correctness brain also considered this but decided it was probably not likely enough to work around!

PR #915, opened 2021-08-19
Re-add Bit type and adapt proofs/circuits

jadephilipoom:
See discussion https://github.com/project-oak/silveroak/pull/904#discussion_r690325205

Earlier I removed `Bit` and replaced it with `BitVec 1`, with the rationale that this would make our core type system simpler. It does, but it also makes individual proofs harder because you end up doing case analysis on an `N` which really you only expect to be a 0 or 1. I tried out adding `Bit` back as a separate type that is denoted as `bool`, and I think overall the wins for simplicity of invariants etc are worth it, so I suggest we add it back!

### DIFF
--- a/cava2/Expr.v
+++ b/cava2/Expr.v
@@ -135,8 +135,8 @@ End ExprNotations.
 Section Var.
   Context {var : tvar}.
 
-  Definition True := Constant Bit 1.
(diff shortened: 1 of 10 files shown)


blaxill (approved):
Lgtm!

PR #919, opened 2021-08-25
Fifo spec

blaxill:
@jadephilipoom let me know if i'm on the right track. It's currently missing any state invariant which we would likely want to say something about the correctness of the fifo vector.

### DIFF
--- a/cava2/Fifo.v
+++ b/cava2/Fifo.v
@@ -18,6 +18,7 @@ Require Import Coq.Strings.String.
 Require Import Coq.Strings.HexString.
 Require Import Coq.Lists.List.
 Require Import Coq.NArith.NArith.
+Require Import Coq.Arith.PeanoNat.
 Require Import Cava.Util.List.
 
 Require Import Cava.Types.
@@ -38,7 +39,7 @@ Section Var.
   Definition fifo {T} fifo_size: Circuit _ [Bit; T; Bit]
     (* out_valid, out, full *)
     (Bit ** T ** Bit ** Bit) :=
-    let fifo_bits := BitVec (Nat.log2 (fifo_size) + 1) in
+    let fifo_bits := BitVec (Nat.log2_up (fifo_size + 1)) in
     {{
     fun data_valid data accepted_output =>
 
@@ -57,7 +58,7 @@ Section Var.
       initially (false,(default,(@default (Vec T fifo_size), 0)))
       : denote_type (Bit ** T ** Vec T fifo_size ** fifo_bits) in
 
-    ( valid, output
+    ( valid, if valid then output else `Constant T default`
     (* Will be empty if accepted_output is asserted next cycle *)
     , (count == `Constant fifo_bits 0` )
     (* We are full, does not assume output this cycle has been accepted yet *)
@@ -160,6 +161,4 @@ Section Var.
     (out_valid, out_data, out_length, is_last, fifo_full)
   }}.
 
-
 End Var.
-
--- a/cava/Cava/Util/List.v
+++ b/cava/Cava/Util/List.v
@@ -136,6 +136,23 @@ Section Misc.
     rewrite app_comm_cons, !removelast_last. reflexivity.
   Qed.
 
+  Lemma removelast_cons {A} (x0 x1:A) xs : removelast (x0 :: x1 :: xs) = x0 :: removelast (x1 :: xs).
+  Proof. reflexivity. Qed.
+
+  Lemma removelast_cons1 {A} (x:A) xs : xs <> [] -> removelast (x :: xs) = x :: removelast xs.
+  Proof. destruct xs; [contradiction|]. intros; apply removelast_cons. Qed.
+
+  Lemma length_removelast_cons {A} xs (x:A) : length (removelast (x :: xs)) = length xs.
+  Proof.
+    revert x.
+    induction xs; [reflexivity|].
+
+    intros x.
+    rewrite removelast_cons.
+    cbn [length].
+    now rewrite (IHxs _).
+  Qed.
+
   Lemma list_unit_equiv (l : list unit) : l = repeat tt (length l).
   Proof.
     induction l; [ reflexivity | ].
--- a/cava2/FifoProperties.v
+++ b/cava2/FifoProperties.v
@@ -81,7 +81,7 @@ Section FifoSpec.
     let '(valid, (data, (accepted_output,_))) := input in
     let '(_, (_, (fifo, count))) := state in
     (* if count is fifo_size we are full and we shouldn't be receiving  *)
-    (if valid then N.to_nat count < fifo_size else Coq.Init.Logic.True).
+    (if valid then N.to_nat count < fifo_size else True).
 
   Definition fifo_invariant
     (contents: list (denote_type T))
--- a/cava2/FifoProperties.v
+++ b/cava2/FifoProperties.v
@@ -1,0 +1,397 @@
+(****************************************************************************)
+(* Copyright 2021 The Project Oak Authors                                   *)
+(*                                                                          *)
+(* Licensed under the Apache License, Version 2.0 (the "License")           *)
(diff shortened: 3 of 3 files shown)


cava2/Fifo.v
```diff
@@ -160,6 +160,122 @@ Section Var.
...
+Check fifo.
+
+Section FifoSpec.
+  Context (T: type).
+  Context (fifo_size: nat).
+
+  Definition fifo_bits_size := Nat.log2_up (fifo_size + 1).
+
+  Notation input_ty := [Bit; T; Bit]%circuit_type.
+  Notation state_ty := (Bit ** T ** Vec T fifo_size ** BitVec fifo_bits_size)%circuit_type.
```
jadephilipoom, line 196: You can also use `input_of` and `state_of` if you want shorter type descriptions without using notations (notations can be harder to control because you can't unfold them when you want to see the contents in the proof).

If you do want to keep the notations, they should definitely be `Local` so that files that import this don't end up e.g. having random parts of their types displayed as `state_ty`!

cava2/Fifo.v
```diff
@@ -160,6 +160,122 @@ Section Var.
...
+
+  Definition fifo_pre
+    (input : denote_type input_ty)
+    (state: denote_type state_ty)
+    :=
+    let '(valid, (data, (accepted_output,_))) := input in
+    let '(_, (_, (fifo, count))) := state in
+    (* fifo is represented as list that should be of size fifo_size *)
+    fifo_size = Datatypes.length fifo /\
```
jadephilipoom, line 205: I think this one probably should be moved to the state invariant (once it's added) rather than the input precondition, since it only involves the state.

cava2/Fifo.v
```diff
@@ -160,6 +160,122 @@ Section Var.
...
+    let out_data := nth (N.to_nat (wrapped_pred count)) fifo default in
+
+    let decrement := accepted_output && (1 <=? count) in
+    let next_count :=
+      if valid && negb decrement then count + 1
+      else if negb valid && decrement then (wrapped_pred count)
+      else count
+    in
+
+    (out_valid, (out_data, (next_count =? 0, (next_count =? N.of_nat fifo_size)))).
```
jadephilipoom, line 228: This looks like a re-implementation of the circuit in Gallina, which I recommend against; when you're trying to use this lemma in other proofs, you'll transform occurrences of `step fifo` into this expression. And there might be a lot of them, so you really want to keep this as small and conceptually simple as possible to avoid expression-size blowup, performance issues, and conceptual overload. You can use ghost variables to keep track of state that's not tracked in the circuit or not efficient to represent, and simplify the logic using those.

For example, you could use a list of occupants of the fifo as a ghost variable (where the state invariant requires that the list length matches `count`). Then, you can easily express the expected outputs in terms of this ghost variable; the entire spec would be :
```coq
let new_contents := tl contents ++ if valid then [data] else [] in
 (length contents =? 0, (hd default contents,
 (length new_contents =? 0, length new_contents =? fifo_size)))
```

cava2/Fifo.v
```diff
@@ -160,6 +160,122 @@ Section Var.
...
+    :=
+    let '(valid, (data, (accepted_output,_))) := input in
+    let '(_, (_, (fifo, count))) := state in
+    (* fifo is represented as list that should be of size fifo_size *)
+    fifo_size = Datatypes.length fifo /\
+    (* if count is fifo_size we are full and we shouldn't be receiving  *)
+    (if count <? N.of_nat fifo_size then Coq.Init.Logic.True else valid = false).
+
+  Definition wrapped_pred x :=
+    (x + 2 ^ N.of_nat fifo_bits_size - 1) mod 2 ^ N.of_nat fifo_bits_size.
```
jadephilipoom, line 210: I think if you write a state invariant that requires `count <= fifo_size`, you won't need this wrapping.

cava2/Fifo.v
blaxill: Makes sense, let me have another go...

blaxill (commented):
I have the clean `fifo_spec` here, but for `step_fifo` i still have one existential in the case of stale data (when the fifo signals invalid). I think this should be alright as `step_fifo_as_valid` can be used if valid is known, which doesn't mention an existential. @jadephilipoom what do you think? 

(This still needs a cleanup pass)

jadephilipoom:
Hmm, it's a bit tricky to formulate. Thinking about it from the perspective of a proof that needs to use this lemma, I have a few thoughts:
- The existential means that `rewrite` won't work; I'd have to `pose proof`, fill in all preconditions, and then destruct to get the existential, even if all I care about is the other half of the tuple. This would be a really awkward process.
- Circuits that use a fifo probably don't only step it when it's non-empty, so in those proofs you'd have to rewrite in the "invalid" case somehow.

One deep change that could fix the issue here and in future places is to encode optional types in our type system. In netlists, they could be implemented as a bit and a value, but in the semantics they'd get interpreted as `option`. Then, in this lemma, it would be completely fine to just write `None` for the invalid case. We could define operations on options in the style of `option_bind`; it's fine to do a computation on an optional value, but your result will also be optional (in circuitry, it would just be wiring through the valid bit).

Alternatively, it's possible to write the `step_fifo` lemma with multiple clauses so that it specifies only the other parts of the result, without mentioning the stale data at all in the invalid case. Something like:
```coq
Lemma step_fifo :
  ...
  (fst (snd (step fifo input state)) = fst (fifo_spec ...)
   /\ if fst (fifo_spec ...) then fst (snd (snd (step fifo input state)) = fst (snd (fifo spec ...)) else True
   /\ snd (snd (snd (step fifo input state)))) = snd (snd (fifo_spec ...)))).
```
However, this is not very readable and the strategy might not scale well if there are many optional outputs.

blaxill:
> * Circuits that use a fifo probably don't only step it when it's non-empty, so in those proofs you'd have to rewrite in the "invalid" case somehow.

Although, the invalid value should never be used so when `valid=false` the existential should also disappear. I agree its awkward though.

> One deep change that could fix the issue here and in future places is to encode optional types in our type system. In netlists, they could be implemented as a bit and a value, but in the semantics they'd get interpreted as `option`. Then, in this lemma, it would be completely fine to just write `None` for the invalid case. We could define operations on options in the style of `option_bind`; it's fine to do a computation on an optional value, but your result will also be optional (in circuitry, it would just be wiring through the valid bit).

Yeah I've been thinking about this but held off in the interest of time, perhaps it would require less work than I've been thinking...

I've just realised I can of course change the circuit to just always zero when invalid, I think in the interest of time I might just do this consistently.

jadephilipoom:
> > * Circuits that use a fifo probably don't only step it when it's non-empty, so in those proofs you'd have to rewrite in the "invalid" case somehow.

> Although, the invalid value should never be used so when `valid=false` the existential should also disappear.

The problem is that even if the existential is unused and will disappear after the lemma is used, the tactic `rewrite` itself won't work. Concrete example:
```coq
Definition foo (x : N) := (x, 1).
Lemma foo_eq x : exists y, foo x = (y, 1).
Proof. exists x. reflexivity. Qed.
Lemma foo2_eq x : snd (foo x) = 1.
Proof.
  Fail rewrite foo_eq.
  (* Error: Cannot find an homogeneous relation to rewrite. *)
```
You can work around it, e.g.:
```coq
  destruct (foo_eq x) as [? H].
  rewrite H. reflexivity.
```
But plugging in all of the arguments for your lemma is going to get really annoying when the lemmas are complex and have many arguments, and when those arguments might be big expressions you have to hunt for in your goal.

> I've just realised I can of course change the circuit to just always zero when invalid, I think in the interest of time I might just do this consistently.

For the sake of short term progress, I think that sounds like a good way to go!

blaxill:
Ok should be good to go... The proofs are a bit ugly but I'm not sure it matters.

jadephilipoom (approved):
LGTM!

PR #922, opened 2021-09-01
Fix bug in SHA-256 padder

jadephilipoom:
While trying to prove the circuit correct, I found that there was a bug in the case where the circuit needs to switch from the "emit_bit" state directory to "writing_length" (occurs if there's exactly one word between the message and length). This PR fixes the bug and adds a regression test.


blaxill (approved):
Argh thanks

PR #929, opened 2021-09-13
Make use of tvar typeclass

jadephilipoom:
Currently, we have a `tvar` typeclass that should help us infer type denotation definitions, but no instances of the typeclass are ever constructed (although they do exist inside sections). Making `denote_type` an instance of `tvar` means we can remove a lot of `(var:=denote_type)` clauses, which seems like an improvement.

### DIFF
--- a/cava2/Expr.v
+++ b/cava2/Expr.v
@@ -24,9 +24,6 @@ Require Import Cava.Primitives.
 
 Local Open Scope N.
 
-Definition tvar : Type := type -> Type.
-Existing Class tvar.
-
 Section Vars.
   Inductive Circuit {var: tvar}: type -> type -> type -> Type :=
   | Var : forall {x},     var x -> Circuit [] [] x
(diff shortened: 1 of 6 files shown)


samuelgruetter:
Does that mean that in a `Section` with a `Context {var: tvar}` where `cava2.Types` is also imported, there are two available `tvar` instances (`denote_type` and `var`)? IIUC, it's not specified which of them Coq's typeclass search will pick. Is that a "risk" you're willing to take? (It might well be worth it, just asking because I'd like to have some understanding of the codebase).

jadephilipoom:
Another option would be to put the `denote_type` instance in a file like `Semantics.v` or something that you would only import when doing proofs, not defining circuits. In the previous version of Cava we had two files that exported various core-library components: one for defining circuits, and the other for proofs, so you didn't have to import a bunch of (slow) proof files in order to just write down a circuit. We could introduce a similar structure and have `denote_type` only declared as an instance for the proofs version.

samuelgruetter:
> I thought in this case the more "local" instance was always preferred.

That seems to be true in my experience, I'm just not sure how much one should rely on it.

> We could introduce a similar structure and have `denote_type` only declared as an instance for the proofs version.

That would be nice, but I'd only do it if it requires very little additional effort. I think this PR could also be merged as is.

jadephilipoom:
> That would be nice, but I'd only do it if it requires very little additional effort. I think this PR could also be merged as is.

It's probably a good idea to re-introduce that structure at some point anyway, so I'll maybe make an issue for it and note in the issue that we should properly handle `tvar` instances (which will be especially important once we also have a netlist-generation instance).

PR #930, opened 2021-09-13
Realign spec and proofs

blaxill:
A few admits still left to go.


cava2/FifoProperties.v
```diff
@@ -395,3 +395,712 @@ Section FifoSpec.
...
+
+Require Import Cava.Util.If.
+Lemma tup_if {A B} (b : bool) (x y: A) (z w: B) : (if b then x else y, if b then z else w) = if b then (x,z) else (y, w).
+Proof. destruct b; reflexivity. Qed.
+
+Lemma length_if {A} (b : bool) (x y: list A) : length (if b then x else y) = if b then length x else length y.
+Proof. destruct b; reflexivity. Qed.
+
+Hint Rewrite @tup_if using solve [eauto] : tuple_if.
+Hint Rewrite @length_if using solve [eauto] : push_length.
```
blaxill, line 903: @jadephilipoom can I add these globally or do you think its a bad idea?

jadephilipoom: I think it's fine to add them, but in the `If.v` file rather than this one.

samuelgruetter:
I see a lot of proofs here that start with `N.bits_inj` and then prove that for each bit index, `N.testbit` at that index returns the same on the left and right of the equality to prove.
I have already developed automation for this proof strategy twice (but for `Z`, not for `N`):
* A [first approach](https://github.com/mit-plv/coqutil/blob/c4265ae220b6ce8dcb0d0f1a90fbc38169ce788b/src/coqutil/Z/prove_Zeq_bitwise.v), mostly used in riscv-coq, was getting a bit complex, and ran into performance problems in `lia`, `btauto`, and rewriting (though `rewrite_strat` helped), and still fails to prove some goals I think it should prove, and also runs forever if I add all the rewrite lemmas I think it should use.
* A simpler [second approach](https://github.com/mit-plv/coqutil/blob/c4265ae220b6ce8dcb0d0f1a90fbc38169ce788b/src/coqutil/Z/bitblast.v), which is easier to deal with, but less powerful than the first one.

I don't claim that there's anything ready to use for you there, but depending on how many more bitwise goals you need to prove, maybe you could harvest some reusable snippets from these files.
Or if you expect really many bitwise goals, it might be worth considering developing a tactic that's actually good and works for all goals in cava & riscv-coq & bedrock2, but that would be a bit of an investment...

blaxill:
> I see a lot of proofs here that start with `N.bits_inj` and then prove that for each bit index, `N.testbit` at that index returns the same on the left and right of the equality to prove.
> I have already developed automation for this proof strategy twice (but for `Z`, not for `N`):
> 
> * A [first approach](https://github.com/mit-plv/coqutil/blob/c4265ae220b6ce8dcb0d0f1a90fbc38169ce788b/src/coqutil/Z/prove_Zeq_bitwise.v), mostly used in riscv-coq, was getting a bit complex, and ran into performance problems in `lia`, `btauto`, and rewriting (though `rewrite_strat` helped), and still fails to prove some goals I think it should prove, and also runs forever if I add all the rewrite lemmas I think it should use.
> * A simpler [second approach](https://github.com/mit-plv/coqutil/blob/c4265ae220b6ce8dcb0d0f1a90fbc38169ce788b/src/coqutil/Z/bitblast.v), which is easier to deal with, but less powerful than the first one.
> 
> I don't claim that there's anything ready to use for you there, but depending on how many more bitwise goals you need to prove, maybe you could harvest some reusable snippets from these files.
> Or if you expect really many bitwise goals, it might be worth considering developing a tactic that's actually good and works for all goals in cava & riscv-coq & bedrock2, but that would be a bit of an investment...

Thanks. I had previously approached those proofs with normal rewriting, but it was taking a huge number of rewrites to massage the goal and was taking too long. The bitwise specialised-for-my-specific goal tactics didn't actually take too long, but I'll definitely use your machinery in the future

jadephilipoom:
I've already got a tactic for bitwise reasoning in `N` in our repo called `push_Ntestbit`. Just apply `N.bits_inj`, then `intro`, then `push_Ntestbit` will take you as far as it can go. Might need further `destruct`s to rewrite with lemmas that have arithmetic side conditions (e.g. for `N.testbit (N,ones x) i` you will need to destruct `i <? x`), so there's room for more automation there, but should help with the majority of the manual steps.

blaxill:
The proof is reduced to some admitted lemmas about `concat_bytes` which all look to be true- I'd like to address them in a follow up PR.

I should have broken down the main proofs further/better but oh well I've learnt my lesson for the future.

@jadephilipoom could you take a look please?

cava2/FifoProperties.v
```diff
@@ -395,3 +401,1015 @@ Section FifoSpec.
...
+    all:
+      rewrite N.testbit_lt with (x:=input) (n:=32) by lia;
+      rewrite Bool.orb_false_r;
+      match goal with
+      | |- N.testbit ?i ?X = N.testbit ?i ?Y => f_equal; lia
+      end.
+  Qed.
+
+  Definition realign_spec
+    latched_valid
```
jadephilipoom, line 865: Looking at the invariant and spec, it seems like this value isn't really used; the spec determines whether the output is valid just based on the state data, and it's not mentioned in the invariant except in a clause that says it matches `out_valid`, which is also not referenced elsewhere. I think you could probably just remove it, which would simplify the reasoning a bit in higher-level proofs.

cava2/FifoProperties.v
```diff
@@ -395,3 +401,1015 @@ Section FifoSpec.
...
+  Admitted.
+
+  Lemma concat_bytes_8_is_64bit xs n:
+        64 <= n -> 
+        N.testbit
+          (BigEndianBytes.concat_bytes (List.resize Byte.x00 8 xs))
+          n = false.
+  Admitted.
+
+  Lemma length_N_to_byte n xs:
```
jadephilipoom, line 713:
```suggestion
  Lemma length_N_to_bytes n xs:
```
I actually have this proven in a local branch that I'll merge once its dependencies are merged, so no need to worry about it.

cava2/FifoProperties.v
```diff
@@ -395,3 +401,1015 @@ Section FifoSpec.
...
+  Admitted.
+
+  Lemma concat_bytes_skip_lower_is_upper xs n:
+        32 <= n -> n < 64 ->
+        N.testbit (BigEndianBytes.concat_bytes (List.resize Byte.x00 8 xs))
+          (n - 32) =
+        N.testbit
+          (BigEndianBytes.concat_bytes (List.resize Byte.x00 8 (skipn 4 xs)))
+          n.
+  Admitted.
```
jadephilipoom, line 704: I think these could be reduced to a single lemma about `N.testbit (BigEndianBytes.concat_bytes _)`, something like:
```coq
N.testbit (BigEndianBytes.concat_bytes bs) n = N.testbit (Byte.to_N (nth (n / 8) bs 0)) (n mod 8)
```
You could then use lemmas about `nth` to do the rest.

cava2/FifoProperties.v
```diff
@@ -395,3 +401,1015 @@ Section FifoSpec.
...
+    }
+    rewrite H1.
+    rewrite N.lor_0_r.
+    rewrite N.shiftl_0_r.
+    reflexivity.
+  Qed.
+
+  Lemma bytes_to_N_id_4 xs:
+    BigEndianBytes.N_to_bytes 4 (BigEndianBytes.concat_bytes (List.resize Byte.x00 4 xs)) = xs.
+  Admitted.
```
jadephilipoom, line 673: I think this one should be true in general:
```suggestion
  Lemma bytes_to_N_id xs:
    BigEndianBytes.N_to_bytes n (BigEndianBytes.concat_bytes (List.resize Byte.x00 n xs)) = xs.
  Admitted.
```

jadephilipoom (approved):
A few comments for small improvements/cleanup, but I think it's OK to merge without blocking on them.

cava2/FifoProperties.v
```diff
@@ -395,3 +401,1015 @@ Section FifoSpec.
...
+    remember (N.to_nat n) as n'.
+    clear Heqn' n.
+
+    { easy_bit_bang n'.
+      rewrite N.shiftl_spec_high' by lia.
+      rewrite N.shiftr_spec by lia.
+      easy_bit_bang n'.
+      rewrite N.bits_above_log2.
+      { rewrite Bool.andb_false_l.
+        rewrite N.bits_above_log2.
```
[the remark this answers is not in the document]
blaxill: The testbit of the constant doesn't reduce with push_Ntestbit alone, but I've reduced it to 

```
450     destr (n' <? 32)%nat.
451     { do 32 (destruct n'; [push_Ntestbit; now boolsimpl| ]); lia. } 
452     {  
453       push_Ntestbit; boolsimpl. 
454       rewrite N.sub_add by lia. 
455       repeat rewrite (N.testbit_high_lt _ _ _ H) by lia.  
456       now boolsimpl.   
457     }
```

It doesn't matter of course, but its a noticeably slower, presumably because of the rewriting order of push_Ntestbit.

cava2/FifoProperties.v
blaxill: It's used in the first branch of `if negb consumer_ready then (latched_valid,..`. Unfortunately in that case it can't be recovered from `state_data`.

PR #932, opened 2021-09-14
More structured invariant-reasoning infrastructure with examples

jadephilipoom:
I've been thinking that we should probably introduce some more structure and ergonomics to the invariant-style reasoning we've been loosely using. So I created some small examples that test the kind of things we want invariant reasoning to be able to do, including handle abstract types and subcircuits.

I've got a certain typeclass-based setup working on the examples, and added some automation. But I think before we commit to this kind of setup, it's worth having a little bit of design discussion about it. The two examples should let us iterate more quickly on ideas in this area.

### DIFF
--- a/cava2/Invariant.v
+++ b/cava2/Invariant.v
@@ -69,6 +69,18 @@ Class correctness_for {s i o} (c : Circuit s i o)
     invariant_preserved_pf : invariant_preserved c;
     output_correct_pf : output_correct c }.
 
+(* Switch between higher-level representations *)
+Definition invariant_logic_isomorphism {s i o} {c : Circuit s i o}
+           {s1 s2} (phi : s1 -> s2) (inv : s2 -> s1)
+           (invariant_logic : invariant_logic_for c s1)
+  : invariant_logic_for c s2 :=
+  {| reset_hl_state := phi reset_hl_state;
+     update_hl_state := fun input x => phi (update_hl_state input (inv x));
+     invariant_for := fun (state : denote_type s) x => invariant_for c state (inv x);
+     precondition_for := fun input x => precondition_for c input (inv x);
+     spec_for := fun input x => spec_for c input (inv x)
+  |}.
+
 (* Succeeds if an instance is found for circuit correctness *)
 Ltac find_correctness c :=
   let x := constr:(_:correctness_for c) in
@@ -133,11 +145,27 @@ Section StateLogicProofs.
         reflexivity. } }
     { intros; logical_simplify; subst. reflexivity. }
   Qed.
+
+  (* if there's an isomorphism between two higher-level states, then invariant
+     logic and correctness proofs can be ported between them. *)
+  Lemma invariant_logic_isomorphism_correct
+        (hl_state' : Type) (phi : hl_state -> hl_state') (inv : hl_state' -> hl_state) :
+    (forall x, inv (phi x) = x) ->
+    correctness_for (invariant_logic:=invariant_logic_isomorphism phi inv _) c.
+  Proof.
+    intros Hphi.
+    constructor; cbv [invariant_at_reset invariant_preserved output_correct
+                                         invariant_logic_isomorphism].
+    all:cbn [reset_hl_state invariant_for precondition_for spec_for].
+    all:rewrite ?Hphi; intros; subst.
+    all:eauto using invariant_preserved_pf, output_correct_pf.
+    all:eapply invariant_at_reset_pf.
+  Qed.
 End StateLogicProofs.
 
 
-(**** Example usage of invariant logic ****)
-Module Example.
+(**** Example usage of invariant logic for counters ****)
+Module DoubleCounterExample.
   Section CircuitDefinitions.
     Context {var : tvar}.
     Import ExprNotations.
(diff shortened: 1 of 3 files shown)


cava2/Invariant.v
```diff
@@ -0,0 +1,605 @@
...
+             if enable
+             then ((value + 1) mod (2 ^ 8))%N
+             else value;
+         precondition := fun _ _ => True;
+         postcondition :=
+           fun input value (output : denote_type (BitVec 8 ** Bit)) =>
+             let '(enable,_) := input in
+             let new_val := if enable then (value + 1)%N else value in
+             output = (new_val mod (2 ^ 8), 2 ^ 8 <=? new_val)
+      |}.
```
jadephilipoom, line 245: This is the specification for a counter with `N` as the higher-level representation:
- `reset_repr` is the higher-level state after a reset (0 here)
- `update_repr` updates the `N` value according to the input
- `precondition` is the precondition based on current higher-level state; here there are no restrictions
- `postcondition` is the postcondition on the output (note that it's a `Prop`, not a function; this will allow us to deal with invalid/stale data in cases like the fifo, and automation takes care of the `pose proof` awkwardness)

cava2/Invariant.v
```diff
@@ -0,0 +1,605 @@
...
+
+    Global Instance counter_invariant : invariant_for counter N :=
+      fun (state : denote_type (BitVec 8 ** _)) value =>
+        let '(data,_) := state in
+        (* value is exactly equivalent to [data] *)
+        data = value
+        (* ...and value < 2 ^ 8 *)
+        /\ value < 2 ^ 8.
```
jadephilipoom, line 253: This is the counter invariant.

cava2/Invariant.v
```diff
@@ -0,0 +1,605 @@
...
+
+    Lemma counter_invariant_preserved : invariant_preserved counter.
+    Proof.
+      intros (enable,[]) (data, ?) value. intros; subst.
+      simplify_invariant counter. simplify_spec counter.
+      cbv [counter]. stepsimpl. logical_simplify; subst.
+      destruct enable; cbn [negb fst snd].
+      all:ssplit; first [ lia | reflexivity
+                          | apply N.mod_bound_pos; lia ].
+    Qed.
```
jadephilipoom, line 305: This is the proof that the counter invariant is preserved. Note that now you don't have to write everything out; the lemma statement is just `invariant_preserved counter`, and typeclasses plug in the right precondition/invariant/high-level representation.

cava2/Invariant.v
```diff
@@ -0,0 +1,605 @@
...
+      ssplit; [ eapply counter_invariant_at_reset .. | ].
+      cbn; lia.
+    Qed.
+
+    Lemma double_counter_invariant_preserved : invariant_preserved double_counter.
+    Proof.
+      cbv [invariant_preserved]. intros (enable,[]) (data,?) value.
+      intros; subst. simplify_invariant double_counter.
+      cbv [double_counter]. stepsimpl. logical_simplify; subst.
+      repeat use_correctness. stepsimpl.
```
jadephilipoom, line 344: This line (the `use_correctness` tactic specifically) automatically finds the counter correctness proofs and does the `pose proof` acrobatics to apply the postcondition. Before this line, we have:
```coq
 let
  '(counter1_state, counter2_state) :=
   fst
     (let
      '(nsx, x) := let '(nsf, o) := step counter data (enable, tt) in (nsf, o) in
       let
       '(nsf, o) :=
        (let (x0, y) := x in
         fun (s : N * bool) (i : unit) =>
         let
         '(ns, o) :=
          let
          '(nsx0, x1) := let '(nsf, o) := step counter s (y, tt) in (nsf, o) in
           let
           '(_, o) :=
            (let (x2, y0) := x1 in
             fun _ _ : unit =>
             let
             '(ns, o) :=
              let
              '(_, x3) :=
               let
               '(_, x3) :=
                let '(nsf, o) := let '(nsf, o) := step bvconcat tt (x2, (x0, tt)) in (nsf, o) in (nsf, o) in
                let '(nsf, o) := step (bvresize 16) tt (x3, tt) in (nsf, o) in (tt, (x3, y0)) in 
              (ns, o)) tt i in (nsx0, o) in (ns, o)) d tt in (nsx, nsf, o)) in
   counter_invariant counter1_state (update_repr (enable, tt) value mod 2 ^ 8) /\
   counter_invariant counter2_state (update_repr (enable, tt) value / 2 ^ 8) /\
   update_repr (enable, tt) value < 2 ^ 16
```
And after:
```coq 
counter_invariant (fst (step counter data (enable, tt))) (update_repr (enable, tt) value mod 2 ^ 8) /\
  counter_invariant
    (fst (step counter d (2 ^ 8 <=? (if enable then value mod 2 ^ 8 + 1 else value mod 2 ^ 8), tt)))
    (update_repr (enable, tt) value / 2 ^ 8) /\ update_repr (enable, tt) value < 2 ^ 16
```

cava2/Invariant.v
```diff
@@ -0,0 +1,605 @@
...
+    Definition double_minimum
+               {T minimum_state} (cmp : Circuit [] [T;T] Bit)
+               (minimum : Circuit minimum_state [T] T)
+      : Circuit _ [T;T] T :=
+      {{ fun input1 input2 =>
+           let min1 := `minimum` input1 in
+           let min2 := `minimum` input2 in
+           let min1_le_min2 := `cmp` min1 min2 in
+           if min1_le_min2 then min1 else min2
+      }}.
```
jadephilipoom, line 425: This example tests that we can leave subcircuits abstract and reason about them based on their specifications. In this case, the `double_minimum` invariant preservation lemma has the type signature:
```coq
double_minimum_invariant_preserved
     : forall (rankT : denote_type ?T -> N) (cmp : Circuit [] [?T; ?T] Bit) (minimum_state : type)
         (minimum : Circuit minimum_state [?T] ?T)
         (minimum_invariant : invariant_for minimum (list (denote_type ?T))),
       correctness_for minimum -> invariant_preserved (double_minimum cmp minimum)
```
You can see that it requires some circuit `minimum`, some invariant for the circuit, and then a proof that that invariant is sufficient to prove the `minimum_specification` defined here. Given that, we have a proof that the `double_minimum` invariant is preserved.

blaxill (approved):
Looks like a great approach to me, is it worth porting existing circuit-spec proofs to this framework?

jadephilipoom:
> Looks like a great approach to me, is it worth porting existing circuit-spec proofs to this framework?

I think if we decide to use it at all, we should port everything to it for consistency.

jadephilipoom:
My main concern with this approach is that it's potentially erring on the side of too much typeclass magic and automation, which could make it difficult to use. For that reason and because I'm leaving soon, I think it would be a good idea for you or Shaked to try porting something or writing a new example, and see if it works well for someone who didn't write it :slightly_smiling_face:

PR #937, opened 2021-09-17
Prove and reorganize all padder helper lemmas

jadephilipoom:
I needed some of the same helper lemmas to adjust some of the other proofs and start with the outer `sha256` circuit, so it made sense to just prove them and move them to the appropriate files. Padder proofs are now closed under the global context :tada:


blaxill (approved):
Lgtm! I'll rebase my PR on main after you merge yours

PR #938, opened 2021-09-17
Adjust sha256_inner proofs

jadephilipoom:
In preparation for proving `sha256`, I realized it would be helpful to change the phrasing of `sha256_inner`'s specification. In particular:
1. The abstract state should reference the message _seen so far_, not the whole message, because `sha256` won't know the whole message if it's only gotten part of the data
2. The postcondition should be in terms of `sha256_step`, not `sha256_compress`, just to make the logic a little smoother

Change (2) was pretty straightforward, but change (1) was more complicated. Basically, the `sha256_inner` circuit really should only reason about the _padded_ message, which is its input, not whatever original message the padded message came from. But this was problematic because the specs require the original message; I couldn't describe what `sha256_step` should return without passing it the original bytes of the message, which were complicated to derive from the circuit input.

To resolve this problem, I created some shadow definitions for some of the spec functions (`M`, `W`, `sha256_step`, and `sha256`) that take the padded message as their argument instead of the original message, and proved they're equivalent to the originals. Then I could prove that `sha256_inner` matches this alternate representation with respect to the padded-message input it gets, without actually requiring that it's a valid padded message. In `sha256`, when I have the context to know that what I'm passing `sha256_inner` is indeed a valid padded message, I can use the equivalence lemma to rewrite back to the original spec.

### DIFF
--- a/silveroak-opentitan/hmac/Spec/SHA256.v
+++ b/silveroak-opentitan/hmac/Spec/SHA256.v
@@ -187,7 +187,7 @@ Section WithMessage.
               (seq 0 64) [].
 
   (* See steps in section 6.2.2. *)
-  Definition sha256_compress (i : nat) (H : list N) (t : nat) : list N :=
+  Definition sha256_compress (W : list N) (H : list N) (t : nat) : list N :=
     (* initialize working variables *)
     let a := nth 0 H 0 in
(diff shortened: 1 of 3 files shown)


blaxill (approved):
Lgtm! thanks

PR #939, opened 2021-09-21
Increment device with TLUL

fshaked:
At the end I didn't use any spec. I suspect both [tlul_adapter_reg] and [incr] are just too simple to justify the overhead of a spec. Also, I can't think of a spec that is any different than the implementation.

### DIFF
--- a/firmware/IncrementWait/CavaIncrementDevice.v
+++ b/firmware/IncrementWait/CavaIncrementDevice.v
@@ -161,11 +161,12 @@ Example sample_trace :=
                ; nop
                ; read_reg 4 (* status *)
              ]%N.
-Print sample_trace.
+(* Print sample_trace. *)
 
 (* Require Import Coq.ZArith.ZArith. Open Scope Z_scope. *)
 
-From Coq Require Import Lia.
+(* From Coq Require Import Lia. *)
+Require Import Coq.micromega.Lia.
 
 Require Import riscv.Utility.Utility.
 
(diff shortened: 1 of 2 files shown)


firmware/IncrementWait/CavaIncrementDevice.v
```diff
@@ -9,93 +8,180 @@ From Cava Require Import
...
+               ; nop
+               ; read_reg 4 (* status *)
+               ; nop
+               ; read_reg 0 (* value *)
+               ; nop
+               ; read_reg 4 (* status *)
+             ]%N.
+Print sample_trace.
+
+(* Require Import Coq.ZArith.ZArith. Open Scope Z_scope. *)
```
samuelgruetter, line 166: I guess commenting out this `Open Scope Z_scope.` lead to the build error
```
File "./IncrementWait/RunIncrementWaitSoftwareOnCava.v", line 28, characters 12-17:
Error: Cannot interpret this number as a value of type nat
```

I sometimes get tired of doing a local `Open Scope Z_scope` in each file and let it leak from one file into other files, which might not be best practice, sorry for that, up to you to decide how to fix it.

samuelgruetter (approved):
> At the end I didn't use any spec.

You did prove a spec, namely

```coq
  Global Instance cava_counter_satisfies_state_machine:
    device_implements_state_machine counter_device increment_wait_state_machine.
```

But yeah you didn't use an *intermediate* spec, and I guess that's fine because the counter is simple enough to prove it directly against the state machine used in the bedrock2 proofs.

Exciting & LGTM, and it will be even cooler with the new `device.run1` signature!

blaxill (approved):
Lgtm, great to see this working!

This has shown that the current definition of `tlul_adapter_reg` is leaky- I think a good next step would be to move `is_value_read_req` here into a field of `TLUL.io_req` and also have `tlul_adapter_reg` handle the masking on `a_ready`. At that point I think all the TLUL code would be contained in `tlul_adapter_reg` which _might_ make it cleaner for the HMAC component.

PR #940, opened 2021-09-21
Add realign_fifo spec and proof

blaxill:
This splits `fifo` `realign` and `realign_fifo` into separate files for easier manipulation (as some of the proof terms are slow), and also adds a spec and proof for `realign_fifo` in terms of the `fifo` and `realign` specs.


fshaked (approved):
LGTM

PR #942, opened 2021-09-23
Change device.run1 to use TL interface

fshaked:


blaxill (approved):
Awesome! Exciting to see things start to fit together

firmware/RiscvMachineWithCavaDevice/MMIOToCava.v
```diff
@@ -257,6 +276,13 @@ Section WithParams.
     assumption.
   Qed.
 
+    Ltac destruct_tl_d2h :=
+      match goal with
+      | v : tl_d2h |- _ =>
+        unfold tl_d2h in v; simpl in v;
+        destruct v as [d_valid [d_opcode [d_param [d_size [d_source [d_sink [d_data [d_user [d_error a_ready]]]]]]]]]
+      end.
```
samuelgruetter, line 284: `destruct_tl_d2h` is duplicated between `MMIOToCava.v` and  `CavaIncrementDevice.v` (but the first one generates fresh names). It would be better to put them right next to the definition of `tl_d2h` (together with getters and setters and tactics to `cbn` them).

samuelgruetter (commented):
Looks great! You might notice though that I was a picky reviewer this time :wink: The reason I care about readability a lot here is that all of `InternalMMIOMachine.v` is part of the statement of our toplevel theorem, so we should work really hard on making that as readable as possible, so that people can actually understand what we're proving.
In summary, I think we just need getters and setters for the tl tuples, which could either be added on top of this PR, or as a new one, whatever works better for you.

fshaked:
@samuelgruetter, I fixed everything you mentioned. Please take another look at the changes in InternalMMIOMachine.v, I used tactics to generate the get/set functions. The generated Galina looks reasonable (no crazy 'match'/'return').

firmware/IncrementWait/CavaIncrementDevice.v
```diff
@@ -237,7 +207,7 @@ Section WithParameters.
     device.is_ready_state s := exists val tl_d2h tlul_state,
         consistent_states tlul_state tl_d2h
         /\ s = mk_counter_state 0 val tl_d2h tlul_state;
-    device.run1 := incr_device_step;
+    device.run1 s i := Semantics.step incr s (i, tt);
```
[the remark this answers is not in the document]
blaxill: It has smartness to avoid units in the state, there is always an extra unit in the right most position of the input which is used by our app constructor https://github.com/project-oak/silveroak/blob/f6cac6d66a45097d8042a5b294d0067d16d72660/cava2/Expr.v#L31

This allows us partial application without HOFs which don't have a direct circuit representation.

firmware/RiscvMachineWithCavaDevice/MMIOToCava.v
```diff
@@ -62,11 +63,16 @@ Class device_implements_state_machine{word: word.word 32}{mem: map.map word Byte
...
       (exists v sH', state_machine.read_step n sH r v sH') ->
       device_state_related sH sL ->
-      exists v sL' sH',
-        device.runUntilResp true false (word_to_N (state_machine.reg_addr r)) (word_to_N (word.of_Z 0))
-                            device.maxRespDelay sL = (Some (word_to_N v), sL') /\
+      exists d2h sL' sH',
+        device.runUntilResp
+          (set_a_valid true
+          (set_a_opcode 4%N (* Get *)
+          (set_a_size 2%N   (* 2^2 bytes *)
```
samuelgruetter, line 70: This value should not just default to 2, but it should match `n`, so you could remove the universal quantification over `n`, and instead quantify over a variable named `log2_nbytes`, pass that to `set_a_size`, and instead of passing `n` to `state_machine.read_step`, you'd pass `2 ^ log2_nbytes`.

firmware/RiscvMachineWithCavaDevice/MMIOToCava.v
```diff
@@ -62,11 +63,16 @@ Class device_implements_state_machine{word: word.word 32}{mem: map.map word Byte
   state_machine_read_to_device_read: forall n r sH sL,
       (exists v sH', state_machine.read_step n sH r v sH') ->
       device_state_related sH sL ->
-      exists v sL' sH',
-        device.runUntilResp true false (word_to_N (state_machine.reg_addr r)) (word_to_N (word.of_Z 0))
-                            device.maxRespDelay sL = (Some (word_to_N v), sL') /\
+      exists d2h sL' sH',
+        device.runUntilResp
+          (set_a_valid true
+          (set_a_opcode 4%N (* Get *)
```
samuelgruetter, line 69: Instead of hardcoding `4` here, you should reuse the [definition of `Get` in `cava2/TLUL.v`](https://github.com/project-oak/silveroak/blob/f6cac6d66a45097d8042a5b294d0067d16d72660/cava2/TLUL.v#L107).
The types won't match, though (it's a Cava `Constant` instead of `N`), but I guess you could change the types of `PutFullData`/`PutPartialData`/`Get` to `N` and prefix their usages in that file by `K`? /cc @blaxill

silveroak-opentitan/hmac/end2end/CavaHmacDevice.v
```diff
@@ -18,11 +18,15 @@ Section WithParameters.
...
     device.is_ready_state := (eq tt);
-    device.run1 s '(is_read_req, is_write_req, req_addr, req_value) :=
+    device.run1 s i :=
       let s' := s in
-      let is_resp := true in
-      let resp := req_value in
-      (s', (is_resp, resp));
+      let res :=
+          set_d_valid true
+          (set_d_opcode 1%N (* AccessAckData *)
```
samuelgruetter, line 25: Here, too, it would be better to refer to a constant defined in `TLUL.v`

samuelgruetter (commented):
These changes look great, but I'd like more of them :wink: So I ended up writing more comments...

firmware/RiscvMachineWithCavaDevice/InternalMMIOMachine.v
```diff
@@ -43,6 +43,161 @@ Module free.
   End WithParams.
 End free.
 
+Require Import Cava.TLUL.
+Require Import Cava.Types.
+
+Definition tl_h2d := denote_type TLUL.tl_h2d_t.
+
+Definition tl_h2d_def :=
```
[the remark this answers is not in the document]
fshaked: "_default" is more consistent with Cava2. In fact, I'm going to change the definition to `default (t := tl_h2d_t)` which evaluates to the same thing.

firmware/RiscvMachineWithCavaDevice/MMIOToCava.v
blaxill: I think that makes sense, and have the update the current circuit definitions to use the constant 'Get := Constant KGet' or similar.

fshaked:
@samuelgruetter, I fixed everything. You can have another go before I merge.

firmware/RiscvMachineWithCavaDevice/InternalMMIOMachine.v
```diff
@@ -144,44 +145,57 @@ Section WithParams.
...
+    Return (LittleEndian.split (2 ^ log2_nbytes) (word.unsigned v)).
+
+  Definition mmioStore(log2_nbytes: nat)(addr: word)(v: HList.tuple byte (2 ^ log2_nbytes))
+    : OState (ExtraRiscvMachine D) unit :=
+    let h2d : tl_h2d :=
+        set_a_valid true
+        (set_a_opcode PutFullData
+        (set_a_size (N.of_nat log2_nbytes)
+        (set_a_address (word_to_N addr)
+        (set_a_data (word_to_N (word.of_Z (LittleEndian.combine (2 ^ log2_nbytes) v)))
```
samuelgruetter, line 175: `(word_to_N (word.of_Z x))` might make sense when you want to make sure that `x` is truncated to the right bitwidth (even though you could also do that with a modulo), but here, the type of `v` already guarantees that it's within bounds, so no truncation is needed, so you could just use `Z.to_N` instead.

fshaked: The combination of word_to_N and word.of_Z makes the stateMachine_primitive_to_cava proof easier. I changed it to Z.to_N anyway, and added a lemma to fix the proof.

PR #943, opened 2021-09-27
Add `concat_bytes` & misc. proofs

blaxill:


cava/Cava/Util/List.v
```diff
@@ -866,9 +870,43 @@ Section Resize.
...
+    intros; cbv [resize].
+    now repeat (natsimpl || listsimpl || push_firstn || push_length).
+  Qed.
+
   Lemma resize_cons {A} a n x (xs: list A):
     resize a (S n) (x::xs) = x :: resize a n xs.
   Proof. now cbn. Qed.
+
+  Lemma resize_resize {A} (x:A) xs n m:
+    n <= m -> resize x n (resize x m xs) = resize x n xs.
```
jadephilipoom, line 887: I think it shouldn't be too hard to make this proof accommodate different default values, e.g.:
```suggestion
  Lemma resize_resize {A} (x1 x2:A) xs n m:
    n <= m -> resize x1 n (resize x2 m xs) = resize x2 n xs.
```

jadephilipoom (approved):
LGTM but slightly confused by the title; what do you mean by "add `concat_bytes`", given the `concat_bytes` definition is already defined?

blaxill:
> LGTM but slightly confused by the title; what do you mean by "add `concat_bytes`", given the `concat_bytes` definition is already defined?

Ah I meant "concat_bytes-and-misc proofs", ill make the commit title clearer.

PR #944, opened 2021-09-28
Use invariant classes for components

blaxill:


cava2/Components/FifoProperties.v
```diff
@@ -41,362 +43,205 @@ Local Open Scope bool.
...
+     postcondition :=
+      fun input contents (output: denote_type (output_of (fifo fifo_size))) =>
+        let '(valid, (data, (accepted_output,_))) := input in
+
+        let pempty := length contents =? 0 in
+        let pfull := length contents =? fifo_size in
+
+        let new_contents :=
+          (if accepted_output then if negb pempty then tl contents else contents else contents)
+          ++ if valid && negb pfull then [data] else [] in
```
jadephilipoom, line 94:
```suggestion
        let pfull := length contents =? fifo_size in

        let new_contents :=
          (if accepted_output then tl contents else contents)
          ++ if valid && negb pfull then [data] else [] in
```
Since `tl [] = []`, the case analysis here can be simplified. If you wanted to make it even simpler, you could also make use of the fact that the precondition requires the fifo not be full if valid data is passed:
```suggestion
        let new_contents :=
          if` accepted_output then tl contents else contents ++ if valid then [data] else [] in
```

cava2/Components/FifoProperties.v
```diff
@@ -41,362 +43,205 @@ Local Open Scope bool.
...
+
+        if pempty then
+        exists unknown, output = (false, (unknown, (length new_contents =? 0, length new_contents =? fifo_size)))
+        else
+        output = (true, (hd default contents, (length new_contents =? 0, length new_contents =? fifo_size)));
```
jadephilipoom, line 99: Automation might have an easier time handling the existential if it's not buried inside a case statement, e.g.
```suggestion
exists valid value,
        output = (valid (value, (length new_contents =? 0, length new_contents =? fifo_size)))
        /\ valid = negb pempty
        /\ (if valid then value = hd default contents else True).
```
When you try to use this lemma later, you'll end up with a proof state something like:
```coq
H : exists valid value,
      output = (valid (value, (length new_contents =? 0, length new_contents =? fifo_size)))
      /\ valid = negb pempty
      /\ (if valid then value = hd default contents else True).
===================================================
P output
```
Then if you just do `logical_simplify; subst` you'll have:
```coq
valid : bool
value : N
H : if valid then value = hd default contents else True
===================================================
P (negb pempty, (value, (length new_contents =? 0, length new_contents =? fifo_size)))
```
Essentially, you won't have to destruct the cases in order to extract the existential or the tuple structure, meaning you'll be able to simplify your goal more easily.

cava2/Components/FifoProperties.v
```diff
@@ -41,362 +43,205 @@ Local Open Scope bool.
...
+      match goal with
+      | |- context [ length _ =? fifo_size ] => destr ( length contents =? fifo_size )
+      | |- _ => idtac
+      end;
+      boolsimpl; natsimpl.
+
+    all: try rewrite E in *; ssplit; try lia; natsimpl.
+    all: try rewrite Nat.sub_add by lia; try rewrite H4;
+         push_firstn; listsimpl; push_firstn;
+         repeat rewrite <- H4;
```
jadephilipoom, line 141: These autogenerated names might lead to later headaches if more hypotheses are inserted, but it's up to you. Since it looks like you use `H4` a few times, you can always rename it, e.g.:
```coq
lazymatch goal with
| H : <whatever H4 should look like> |- _ => rename H into foo
end.
```

cava2/Components/FifoProperties.v
```diff
@@ -41,362 +43,205 @@ Local Open Scope bool.
...
-      rewrite tl_length.
-      reflexivity.
+    fold denote_type in *.
+    rewrite Bool.negb_andb;
+      destr (length contents =? 0);
+      destr (length contents =? fifo_size);
+      revert H; subst; boolsimpl; ssplit; [lia|..]; intros.
+    {
+      eexists.
+      rewrite E.
```
jadephilipoom, line 201: Based on what's above I'm guessing this could be `subst` instead, which will be less fragile and naming-sensitive.

cava2/Components/FifoProperties.v
```diff
@@ -41,362 +43,205 @@ Local Open Scope bool.
...
-      assert (Datatypes.length (contents) <> Datatypes.length fifo) by lia;
-      destruct (Nat.eqb_spec (Datatypes.length (contents)) (Datatypes.length fifo)); try tauto;
-      try rewrite app_length;
-      try rewrite tl_length;
-      try now (cbn; lia).
+      rewrite Nat.add_sub_swap by lia;
+      rewrite Nat.add_mod by (try apply Nat.pow_nonzero; lia);
+      rewrite Nat.mod_same by (try apply Nat.pow_nonzero; lia);
+      rewrite Nat.add_0_r;
+      rewrite Nat.mod_mod by (try apply Nat.pow_nonzero; lia);
```
jadephilipoom, line 212: Sometimes when I have a lot of `(2 ^ n <> 0)` side conditions I like to just `assert (2 ^ n <> 0) by (apply Nat.pow_nonzero; lia)` once, and then everything else will go by just `lia`.

cava2/Components/RealignMaskedFifoProperties.v
```diff
@@ -0,0 +1,382 @@
...
+
+      let realign_input :=
+        (data_valid, (data, (data_mask, (drain && fifo_empty && consumer_ready, (negb fifo_full, tt))))) in
+      let '(realign_valid, realign_data, new_realign_contents) :=
+        update_repr (c:=realign) realign_input realign_contents in
+      let fifo_input :=
+        (realign_valid && negb drain && negb fifo_full, (BigEndianBytes.concat_bytes realign_data, (consumer_ready, tt))) in
+
+      precondition realign realign_input realign_contents
+      /\ precondition (fifo (T:=BitVec 32) fifo_size) fifo_input fifo_contents
```
jadephilipoom, line 109: I advise against this strategy of reproducing the input for subcircuits then referencing their specifications in the specification for this circuit (referencing invariants is OK). That strategy appears here in the postcondition and update_repr as well, but I'll explain my reasoning with reference to the precondition case so I can be specific.

When a circuit uses `realign_masked` as a subcircuit, it'll have to prove this precondition. Does that circuit really want to be unfolding and reasoning directly about the `fifo` precondition with a complicated input expression? It's a very leaky abstraction. I think usability will be improved if you write as concise a precondition as you can that doesn't reference the subcircuits but is strong enough to imply their preconditions, even if it means repeating clauses from the subcircuits' preconditions; then the next layer can think about "the realigner's preconditions" instead of "all the realigners' subcircuits' preconditions", which doesn't scale.

In this specific case, the fifo precondition says only that if new data is passed, the fifo must not be full. That's guaranteed by the input to the fifo in the `realign_masked` circuit, which includes `negb fifo_full` in the computation of the fifo's `data_valid` input; by construction, this circuit will never violate the fifo's precondition. Therefore, instead of passing the precondition on to any circuit using the realigner as a subcircuit, you can completely solve that requirement at this layer without any new preconditions.

The realigner precondition just says that the data has to be `< 2 ^ 32`. So that one you can duplicate.

In the end, the entire precondition would be reduced to just one line, with no references to subcircuits or even state:
```suggestion
      data < 2 ^ 32
```
This will be much easier to reason about for the next layer up!

cava2/Invariant.v
```diff
@@ -101,11 +101,19 @@ Ltac simplify_invariant c :=
   let x := constr:(_:invariant_for c _) in
   match x with
   | ?x => cbv [x] in *
+  | ?x _ => cbv [x] in *
+  | ?x _ _ => cbv [x] in *
+  | ?x _ _ _ => cbv [x] in *
   end.
```
jadephilipoom, line 107: This code looks really fragile to me. It only works for circuits with <= 3 parameters, which makes the behavior a little unpredictable. I think if you made the parameters in the `FifoSpec` section implicit, you might avoid the problem.

cava2/Invariant.v
```diff
@@ -101,11 +101,19 @@ Ltac simplify_invariant c :=
...
 Ltac simplify_spec c :=
   let x := constr:(_:specification_for c _) in
   match x with
   | ?x => cbn [x update_repr precondition postcondition] in *
+  | ?x _ => cbn [x update_repr precondition postcondition] in *
+  | ?x _ _ => cbn [x update_repr precondition postcondition] in *
+  | ?x _ _ _ => cbn [x update_repr precondition postcondition] in *
+  | ?x _ _ _ _ => cbn [x update_repr precondition postcondition] in *
+  | ?x _ _ _ _ _ => cbn [x update_repr precondition postcondition] in *
```
jadephilipoom, line 116: Same complaint here as above, and same suggested fix.

jadephilipoom (changes requested):

cava2/Components/RealignMaskedFifoProperties.v
blaxill: That makes sense thanks, let me have another go ..

cava2/Invariant.v
blaxill: Hmm, I can't get it to work- any combination of implicit via the section or by messing with `Arguments` gives the same error as the default which is 

```
Ltac variable x is bound to fifo_specification which cannot be coerced to an evaluable reference.
```

As I'm not sure of another way to address this I'll leave it for now, I agree its fragile.

cava2/Components/FifoProperties.v
blaxill: Done thanks

cava2/Components/RealignMaskedFifoProperties.v
blaxill: Done thanks, things are simpler now. `update_repr` doesn't use the specifications but does use `realign_inner_spec` simply as I can't think of a way to simplify it further.

blaxill:
@fshaked ok to merge?

cava2/Invariant.v
jadephilipoom: Ooh, I think I actually know how to solve that one -- one last thing to try could be inserting `let x := app_head x in` before the match. That error looks like `cbv` is not happy with the fact that `x` has implicit arguments already applied, and `app_head` strips all arguments from a term, so I think you can use that + implicit arguments to have a situation where typeclass inference has the arguments but `cbv` doesn't have them.

blaxill: That works! thanks

PR #961, opened 2021-11-04
Add HMAC registers to circuit needed by software:

blaxill:
- Sets the STATUS bit 1 to the value of `fifo_full`
- Uses CFG bit 2 (`endian_swap`) to switch input endianness


fshaked (approved):
LGTM

blaxill:
Shaked has identified an additional needed blocking condition, which I will shortly add to this PR.

fshaked:
A bit too late but, seems to me like the `was_fifo_full` field is not really needed, instead apply `mask_ready` to `tl_o` inside the `let/dealy`. I will do this change if you don't see anything wrong with it.

PR #965, opened 2021-11-26
Express device_implements_state_machine using only run1, without runUntilResp

fshaked:
This PR includes @samuelgruetter's #960 (which should probably be closed without merging).

Fix the problems of `runUntilResp`, discussed in #960.
Change the IncrWait end2end proof to use the invariant-spec.
Fix TL: save the address of the outstanding request, instead of using the address from the input (which might change).
Fix TL spec.
Split the IncrWait device file to cava+spec, and end2end proof.

### DIFF
--- a/firmware/IncrementWait/Incr.v
+++ b/firmware/IncrementWait/Incr.v
@@ -1,0 +1,598 @@
+Require Import Coq.Lists.List.
+Require Import Coq.micromega.Lia.
+Require Import Coq.ZArith.ZArith.
+
+Require Import Cava.Expr.
+Require Import Cava.ExprProperties.
+Require Import Cava.Invariant.
+Require Import Cava.Primitives.
+Require Import Cava.Semantics.
+Require Import Cava.TLUL.
+Require Import Cava.Types.
+Require Import Cava.Util.BitArithmetic.
+Require Import Cava.Util.List.
+Require Import Cava.Util.Tactics.
+
+Require Import coqutil.Tactics.Simp.
+Require Import coqutil.Tactics.Tactics.
+
+Import ListNotations.
+
+Section Var.
+  Import Expr.
+  Import ExprNotations.
+  Import PrimitiveNotations.
+
+  Local Open Scope N.
+
+  Context {var : tvar}.
+
+  Definition incr_state := BitVec 2.
+  Definition Idle       := Constant incr_state 0.
+  Definition Busy1      := Constant incr_state 1.
+  Definition Busy2      := Constant incr_state 2.
+  Definition Done       := Constant incr_state 3.
+
+  Definition inner
+    : Circuit _ [Bit; BitVec 32] (Bit ** BitVec 32)
+    := {{
+      fun valid data =>
+        let/delay '(istate; value) :=
+           let istate' :=
+               if istate == `Busy1` then `Busy2`
+               else if istate == `Busy2` then `Done`
+                    else if istate == `Done` then `Idle`
+                         else (* istate == `Idle` *)
+                           if valid then `Busy1`
+                           else `Idle` in
+
+           let value' :=
+               if istate == `Busy2` then value + `K 1`
+               else if istate == `Idle` then data
+                    else value in
+
+           (istate', value')
+             initially default
+           : denote_type (incr_state ** BitVec 32)
+        in
+        (istate == `Done`, value)
+       }}.
+
+  Definition incr
+    : Circuit _ [tl_h2d_t] tl_d2h_t
+    := {{
+      fun tl_h2d =>
+        (* Destruct and reassemble tl_h2d with a_address that matches the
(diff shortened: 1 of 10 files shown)


fshaked:
@jadephilipoom, notice that I use the invariant-spec in the big proof in `firmware/IncrementWait/CavaIncrementDevice.v`, showing that it is indeed very useful (no need to reason about Cava at all). But, (in that file) I never use `output_correct_pf`. In fact, the postcondition of the `incr` device is just `True`, because it is not needed. This is because the `device` class from `InternalMMIOMachine.v` includes the function `last_d2h`, from cava-device-state to cava-output, which is then used instead of the real output. We did this because, in principal, we need to record the device output in order to know if the input of the next step will be accepted (TL protocol), and we didn't want to do that, so instead the device class shifts the responsibility to the cava device, and now the cava device state must record enough information so that the output can be computed from the state alone (i.e., the cava output must be a function of the let/delay bindings, and must not include the input bindings), which happened to be the case anyway in the `incr` device (and the `hmac_top`, but that might change soon).

firmware/RiscvMachineWithCavaDevice/InternalMMIOMachine.v
```diff
@@ -54,35 +54,66 @@ Module device.
...
+    (* the d2h output the device produced when it transitioned to the state *)
+    (* TODO: probably need to add to [device_implements_state_machine] somthing like
+     [forall s h2d d2h s', run1 s h2d = (s', d2h) -> last_d2h s' = d2h] *)
+    last_d2h: state -> tl_d2h;
+
+    (* indicates an inflight operation: the device received a request on channel
+       A, but a response on channel D hasn't been exchanged yet *)
+    (* TODO: probably need to add to [device_implements_state_machine] somthing like
+     [forall s, is_ready_state s -> tl_inflight_ops s = []] *)
+    tl_inflight_ops: state -> list N;
```
samuelgruetter, line 66: Now that `device_state_related` takes a `list tl_h2d` of inflight packets, is this `tl_inflight_ops` field still needed?

firmware/RiscvMachineWithCavaDevice/InternalMMIOMachine.v
```diff
@@ -54,35 +54,66 @@ Module device.
...
-     - on the hardware level, using readResp/writeResp, which appears as
+     - on the hardware level, using runUntilResp, which appears as
        blocking I/O for the software
      - on the software level, using MMIO reads on some status register,
        where the MMIO read immediately gives a "busy" response, and the
        software keeps polling until the MMIO read returns a "done" response *)
 
-  (* returning None means out of fuel and must not happen if fuel >= device.maxRespDelay *)
+  Definition waitForResp{D: device} :=
+    fix rec(fuel: nat)(s: device.state): option tl_d2h * device.state :=
```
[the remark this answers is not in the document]
fshaked: You probably mean return `device.state`, and not `option device.state`. It's true that the `option tl_d2h` part is not strictly needed anymore, but it's probably not a good idea to remove it. If you do, the caller will have to call `last_d2h` and check the `d_valid` of the result to know if the call was successful or not.

firmware/RiscvMachineWithCavaDevice/InternalMMIOMachine.v
fshaked: Removed.

firmware/RiscvMachineWithCavaDevice/InternalMMIOMachine.v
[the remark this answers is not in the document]
samuelgruetter: I meant `option device.state`, where returning `None` means that we ran out of fuel, and returning `Some` means that `d_valid` is true, so the caller only needs to check if we got `Some`.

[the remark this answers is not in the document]
fshaked: Done

samuelgruetter (approved):

firmware/IncrementWait/Incr.v
```diff
@@ -0,0 +1,535 @@
...
+  Existing Instances inner_invariant_at_reset inner_invariant_preserved
+           inner_output_correct.
+  Global Instance inner_correctness : correctness_for inner.
+  Proof. constructor; typeclasses eauto. Defined.
+
+
+  Variant repr_state :=
+  | RSIdle
+  | RSBusy (data : N) (count: nat)
+  | RSDone (res : N).
```
samuelgruetter, line 258: Looking at the code again, I notice that `inner_state` is exactly the same as `repr_state`, so I guess you could delete one of them?
```coq
  Variant inner_state :=
  | IISIdle
  | IISBusy (data : N) (count : nat)
  | IISDone (res : N).
```

fshaked: Done

